PR #11, opened 2015-08-26
Add forgiveness on both sides of expected span durations on tests

matthewloring:
Express tests have been flaky because spans are sometimes a few MS shorter than expected duration.


ofrobots:
LGTM.

PR #21, opened 2015-08-26
Run standalone tests on travis

matthewloring:
Also moved cluster test to standalone so hooks can be run in the same mocha invocation as unit tests.


ofrobots:
LGTM.

matthewloring:
Switched over to `$(npm bin)/mocha`

Issue #2, opened 2015-08-25
config.projectId should be a string


ofrobots:
The trace API expects the project Id to be a string, but does accept the numeric project Id. Users get a confusing error message when we try to post the traces to the API if they specific a numeric project Id as a number rather than a string.

ofrobots:
Changed the requirement for the projectId to be a string instead of doing automatic coercion. The user should be provided an error message immediately if this is not the case.

ofrobots:
Landed in 1be000b.

PR #22, opened 2015-08-26
improve handling of config.projectId

ofrobots:
Fixes: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/issues/2

### DIFF
--- a/test/standalone/test-agent-metadata.js
+++ b/test/standalone/test-agent-metadata.js
@@ -18,26 +18,38 @@
 
 var assert = require('assert');
 var nock = require('nock');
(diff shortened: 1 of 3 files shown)


test/standalone/test-agent-metadata.js
```diff
@@ -18,26 +18,38 @@
 
 var assert = require('assert');
 var nock = require('nock');
+var agent = require('../../');
```
matthewloring: Nit: we have this written ../.. everywhere else.

matthewloring:
LGTM after nit

Issue #13, opened 2015-08-26
Don't use dev dependencies for instrumented frameworks


matthewloring:
This will also allow us to test against multiple versions of each framework.

matthewloring:
We can use this to track moving things like hapi and restify out of the dev dependencies. I think generating separate package.json files sounds like a good direction to take this.

matthewloring:
To expand on multiple frameworks, we should test against all published versions of a module satisfying the semver rule we specify.

PR #23, opened 2015-08-26
Removed dev dependencies badge

matthewloring:
As per offline: it doesn't make sense for us to have up to date dev dependencies since we need to trace older versions of supported frameworks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,6 @@
 [![NPM Version][npm-image]][npm-url]
 [![Build Status][travis-image]][travis-url]
 [![Dependency Status][david-image]][david-url]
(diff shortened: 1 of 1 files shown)


matthewloring:
Addresses #13.

ofrobots:
How will we test multiple different versions of the framework? Can we have multiple versions of the same framework in the package.json? Perhaps we need to come up with a different mechanism of testing these frameworks that doesn't involve devDependencies.

Thoughts?

matthewloring:
I think generating multiple package.json files is a good place to start. I'll close this pull request and start working on that.

PR #30, opened 2015-08-27
Allow for trace sampling

matthewloring:


lib/trace-writer.js
```diff
@@ -78,12 +78,16 @@ TraceWriter.prototype.queueTrace_ = function(trace) {
     if (err) { return; } // ignore as index.js takes care of this.
 
     trace.projectId = project;
-    var length = that.buffer_.push(trace);
 
-    that.logger_.info('queued trace. new size:', length);
+    if (that.buffer_.length <= that.config_.bufferSize) {
+      that.buffer_.push(trace);
+      that.logger_.info('queued trace. new size:', that.buffer_.length);
```
ofrobots: While you're in here, the log level of INFO on each trace is too aggressive. We should probably drop it down to DEBUG.

matthewloring:
As per offline, we'll close this for now to be revisited once we have a better idea of the desired sampling algorithm.

PR #31, opened 2015-08-27
Some cleanups, and fix for span warnings on own requests

### LINKED ISSUE #4: Remove warnings for own context-less outgoing http requests

ofrobots:
Fixes #4.

### DIFF
--- a/index.js
+++ b/index.js
@@ -62,7 +62,7 @@ var publicAgent = {
 
   start: function(projectConfig) {
     if (this.isActive()) { // already started.
-      return publicAgent;
+      return this;
     }
 
     var util = require('util');
@@ -72,7 +72,7 @@ var publicAgent = {
     var logger = new Logger(config.logLevel, '@google/cloud-trace');
     if (!semver.satisfies(process.versions.node, '>=0.12')) {
       logger.error('Tracing is only supported on Node versions >=0.12');
-      return publicAgent;
+      return this;
     }
 
     if (typeof config.projectId === 'undefined' &&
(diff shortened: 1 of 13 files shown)


lib/utils.js
```diff
@@ -123,23 +123,27 @@ function authorizedRequestFactory(scopes) {
...
+    headers = {};
   }
 
+  headers['Metadata-Flavor'] = 'Google';
+
   var request = require('request');
   requestWithRetry(request, {
-    url: 'http://metadata/computeMetadata/v1/project/numeric-project-id',
-    headers: { 'Metadata-Flavor': 'Google'},
+    url: 'http://metadata.google.internal/computeMetadata/v1/project/' +
```
matthewloring: Do we want to move URLs into constants as well?

ofrobots: Good idea in general, but this one I am hesitant to move because I want to extract utils into a separate module.

lib/trace-writer.js
```diff
@@ -17,10 +17,15 @@
 'use strict';
 
 var utils = require('./utils.js');
+var constants = require('./constants.js');
+var pjson = require('../package.json');
 
 /* @const {Array<string>} list of scopes needed to operate with the trace API */
 var SCOPES = ['https://www.googleapis.com/auth/trace.append'];
```
matthewloring: Another contender for constants?

index.js
```diff
@@ -67,7 +72,7 @@ var publicAgent = {
     var logger = new Logger(config.logLevel, '@google/cloud-trace');
```
matthewloring: Maybe a constant?

matthewloring:
LGTM. I pointed out a few other potential constants but they don't need to go in right now.

index.js
ofrobots: It is not used anywhere else, and is a very descriptive value. It is better left inline here.

lib/trace-writer.js
ofrobots: It is not exported or used in any other file, so there is no benefit from moving it to constants.

PR #33, opened 2015-08-27
Http spans no longer escaping on server error

matthewloring:


lib/hooks/core/hook-http.js
```diff
@@ -79,6 +79,9 @@ function requestWrap(request) {
       }
     });
     namespace.bindEmitter(returned);
+    returned.on('error', function() {
+      currContext.close();
```
ofrobots: Can we add the error as a label onto the context?

matthewloring: We have to see how other Google tracing agents handle this: added it as #36.

ofrobots:
LGTM.

PR #37, opened 2015-08-28
Added unhook test coverage

matthewloring:


test/test-span-data.js
```diff
@@ -58,6 +59,23 @@ describe('SpanData', function() {
     });
   });
 
+  it('adds context to headers', function() {
```
ofrobots: This spec doesn't read too well. If I am not mistaken this test checks that 'addContextToHeaders adds context to headers'. Current layout of the code makes it read as 'SpanData adds context to headers'.

Can you wrap this section with a `describe('addContextToHeaders', ...)`.

ofrobots:
LGTM, with a minor comment.

PR #38, opened 2015-08-28
stringify spans when they're queued

matthewloring:
This will avoid a huge stringify call holding down the cpu on publish.

### DIFF
--- a/lib/trace-writer.js
+++ b/lib/trace-writer.js
@@ -41,7 +41,7 @@ function TraceWriter(logger, config) {
   /** @private {function} authenticated request function */
   this.request_ = utils.authorizedRequestFactory(SCOPES);
 
-  /** @private {Array<Trace>} array to cache buffers */
+  /** @private {Array<string>} stringified traces to be published */
   this.buffer_ = [];
 
   // Schedule periodic flushing of the buffer, but only if we are able to get
@@ -83,7 +83,7 @@ TraceWriter.prototype.queueTrace_ = function(trace) {
     if (err) { return; } // ignore as index.js takes care of this.
 
     trace.projectId = project;
-    var length = that.buffer_.push(trace);
+    var length = that.buffer_.push(JSON.stringify(trace));
 
     that.logger_.info('queued trace. new size:', length);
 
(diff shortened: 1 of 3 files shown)


matthewloring:
Fixes #26

lib/trace-writer.js
```diff
@@ -120,14 +120,10 @@ TraceWriter.prototype.flushBuffer_ = function(projectId) {
   // Privatize and clear the buffer.
   var buffer = this.buffer_;
   this.buffer_ = [];
+  var joined = buffer.join();
+  buffer = '[' + joined.slice(0, joined.length -1) + ']';
```
ofrobots: I am not sure if I understand why the slice is necessary. Doesn't it remove the last `}`? Also, need space before `1`.

For example:
```
var a = {foo: 'bar'};
[a,a,a,a].map(JSON.stringify).join()
→ "{"foo":"bar"},{"foo":"bar"},{"foo":"bar"},{"foo":"bar"}"
```
`slice` as used above will remove the last `}`, no?

lib/trace-writer.js
```diff
@@ -120,14 +120,10 @@ TraceWriter.prototype.flushBuffer_ = function(projectId) {
...
 
-  var traces = { traces: buffer };
-  // FIXME(ofrobots): if the buffer is large, this is going to burn
-  // a lot of CPU and block the event loop. Figure out a way to do this some
-  // other way. Since node-core doesn't have a streaming JSON stringifier
-  // we'll need to look at npm modules for this.
-  // Alternatively, we should switch to protobufs
-  this.publish_(projectId, JSON.stringify(traces));
+  this.publish_(projectId, '{"traces":' + buffer + '}');
```
ofrobots: Would it be better (i.e. fewer allocations, copies) to replace this line and line 124 with a `util.format`? I am not sure how `util.format` is implemented, but worth checking.

lib/trace-writer.js
matthewloring: Good catch, this was a carry over from the fully string based buffering. It's been eliminated.

lib/trace-writer.js
matthewloring: Looks like it would be overkill in this context: https://github.com/nodejs/node/blob/master/lib/util.js#L9 it might be cleaner but I don't think it would be more performant

lib/trace-writer.js
ofrobots: Since a bug escaped our testing, we should add unit tests for this:
* Check invariant that buffer_ is an array of stringified objects.
* Check that we are able to `JSON.parse` back the buffer produced by `flushBuffer_`.

matthewloring: I am tracking the need for trace writer unit testing with #47. It's up next.

ofrobots:
LGTM.

PR #41, opened 2015-08-28
Removed common trace/debug functionality

matthewloring:
Functionality moved into https://www.npmjs.com/package/@google/cloud-diagnostics-common


index.js
```diff
@@ -17,7 +17,8 @@
 'use strict';
 
 var OpaqueSpan = require('./lib/opaque-span.js');
-var Logger = require('./lib/logger.js');
+var CloudCommon = require('@google/cloud-diagnostics-common');
```
ofrobots: Capitalization here incorrectly suggests that this is a constructor / class name. Use `cloudCommon`, or just `common` instead.

index.js
```diff
@@ -17,7 +17,8 @@
...
+var Logger = CloudCommon.logger;
```
ofrobots: ditto. This should be `var Logger = common.Logger;` as the latter is actually a class, so it should exported from that module as such.

matthewloring: Tracking logger capitalization in common here: https://github.com/matthewloring/cloud-diagnostics-common-nodejs/issues/1

ofrobots:
LGTM.

PR #42, opened 2015-08-29
Use agent api in hooks

matthewloring:


matthewloring:
Progress on #25

lib/hooks/core/hook-http.js
```diff
@@ -51,13 +51,13 @@ function requestWrap(request) {
 
     var namespace = cls.getNamespace();
     var uri = uriFromOptions(options);
-    var currContext = cls.getRootContext().createChildSpanData(uri);
-    currContext.span.setLabel(TraceLabels.HTTP_METHOD_LABEL_KEY, options.method);
-    currContext.span.setLabel(TraceLabels.HTTP_URL_LABEL_KEY, uri);
+    var span = agent.startSpan(uri);
+    span.addLabel(TraceLabels.HTTP_METHOD_LABEL_KEY, options.method);
```
ofrobots: Can you pass the labels to the `agent.startSpan`? I think the goal should be to make sure `OpaqueSpan` is true to its name, or if that's not possible, call stop calling it `OpaqueSpan`.

lib/hooks/core/hook-http.js
```diff
@@ -51,13 +51,13 @@ function requestWrap(request) {
...
+    span.addLabel(TraceLabels.HTTP_URL_LABEL_KEY, uri);
     if (options.headers) {
       // Adding context to the headers lets us trace the request
       // as it makes it through other layers of the Google infrastructure
-      currContext.addContextToHeaders(options.headers);
+      span.spanData_.span.addContextToHeaders(options.headers);
```
ofrobots: Yeah, this is a bit ugly. We don't have a mechanism to do this cleanly without reaching into the internals here. Can you open an issue for this?

lib/hooks/core/hook-http.js
```diff
@@ -68,11 +68,9 @@ function requestWrap(request) {
...
       });
       res.on('end', function() {
-        currContext.span.setLabel(
-          TraceLabels.HTTP_RESPONSE_SIZE_LABEL_KEY, numBytes);
-        currContext.span.setLabel(
-          TraceLabels.HTTP_RESPONSE_CODE_LABEL_KEY, res.statusCode);
-        currContext.close();
+        span.addLabel(TraceLabels.HTTP_RESPONSE_SIZE_LABEL_KEY, numBytes);
+        span.addLabel(TraceLabels.HTTP_RESPONSE_CODE_LABEL_KEY, res.statusCode);
+        span.end();
```
ofrobots: use `agent.endSpan`

lib/hooks/core/hook-http.js
```diff
@@ -81,7 +79,7 @@ function requestWrap(request) {
     });
     namespace.bindEmitter(returned);
     returned.on('error', function() {
-      currContext.close();
+      span.end();
```
ofrobots: likewise

lib/hooks/core/hook-http.js
matthewloring: I definitely can pass them to `agent.startSpan`. Do you want `addLabel` to be removed as well? I don't have a strong preference but if you don't want it used then it should be removed.

ofrobots: You mean remove it from the `OpaqueSpan` interface? It should be removed, or at the least, made private.

I think I might have not done a good job explaining the motivation behind why we want to do this. We want to use the public agent API instead of the private API for a few reasons:
* DRY. There should be only one way of doing something.
* If the public API is sufficient for a use-case, it should be used instead of a private API. This reduces tight coupling between internal components.
* It gets the public API tested for usefulness and finds the deficiencies. If there is use-cases that it doesn't solve for us, it is probably going to let external users down as well.

Hope this makes sense.

lib/hooks/core/hook-http.js
matthewloring: I don't think we have an `agent.endSpan` function. It is called in `index.js` but never defined.

ofrobots: That's a bug I introduced :fearful:. Do you mind fixing it as part of this PR? You can copy the implementation from phantomAgent.

lib/hooks/core/hook-http.js
matthewloring: Tracked #43 43

ofrobots:
LGTM.

PR #45, opened 2015-08-29
Don't exit process if mongo/redis isn't present

### LINKED ISSUE #44: Test failure if mongo or redis cannot connect
Right now the test exits with success and the rest of the suite isn't run.

matthewloring:
fixes #44


ofrobots:
LGTM.

PR #46, opened 2015-08-29
Publish script

matthewloring:


bin/run-bump.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+echo 'Bump to version $1 and push to $2?'
+read proceed
+git checkout master
+git fetch -v $2
+git rebase $2/master
```
ofrobots: all of these lines should be suffixed with a `|| exit 1` to  make sure we don't continue if a step along the way fails.

FYI http://www.shellcheck.net/ for linting shell scripts

matthewloring:
Addressed concerns, @ofrobots PTAL.

ofrobots:
LGTM.

PR #51, opened 2015-09-01
Fixed mongo test to hit find path with no parent span

matthewloring:
@ofrobots PTAL


ofrobots:
LGTM if the PR description and the commit log is updated to describe 'unintended'.

PR #58, opened 2015-09-02
Test coverage for adding context to headers

matthewloring:
We already have other tests guaranteeing the formatting of the headers. This just ensures that code path is exercised and the api is called correctly.

### DIFF
--- a/lib/hooks/core/hook-http.js
+++ b/lib/hooks/core/hook-http.js
@@ -101,8 +101,6 @@ function uriFromOptions(options) {
   var uri;
   if (typeof options === 'string') {
     uri = options;
-  } else if (options.uri && options.uri.href) {
-    uri = options.uri.href;
   } else {
     // In theory we should use url.format here. However, that is
     // broken. See: https://github.com/joyent/node/issues/9117 and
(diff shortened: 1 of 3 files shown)


lib/hooks/core/hook-http.js
```diff
@@ -101,8 +101,6 @@ function uriFromOptions(options) {
   var uri;
   if (typeof options === 'string') {
     uri = options;
-  } else if (options.uri && options.uri.href) {
```
matthewloring: I can't find any mention of a uri option in the http spec: 

https://nodejs.org/api/http.html#http_http_request_options_callback

matthewloring:
@ofrobots PTAL.

lib/hooks/core/hook-http.js
ofrobots: `npm.im/request` is probably the reason this existed. See also: https://github.com/request/request/blob/master/request.js#L204

Looking at _http_client though, even if `uri` made it to here, it would be the wrong thing for us to use. The following would have logged the incorrect url: `http.get({uri: "http://wrong.com/bar", host: "google.com" }...`, so thanks for fixing.

ofrobots:
LGTM.

Issue #19, opened 2015-08-26
Stack traces do not show up in the developer console ui


matthewloring:
Stack traces are not being formatted according to the associated protobuf and are not rendered by the ui.

matthewloring:
Fixed in #59

PR #59, opened 2015-09-02
Implement support for structured stack traces

ofrobots:
PTAL @matthewloring

### DIFF
--- a/lib/span-data.js
+++ b/lib/span-data.js
@@ -52,10 +52,20 @@ function SpanData(agent, trace, name, parentSpanId) {
     };
     var e = {};
     Error.captureStackTrace(e, SpanData);
-    var formatted = e.stack.reduce(function(str, frame) {
-      return str + frame.toString() + '\n';
-    }, '');
-    this.span.setLabel(TraceLabels.STACK_TRACE_DETAILS_KEY, formatted);
+
+    var stackFrames = [];
+    e.stack.forEach(function(callSite) {
+      var functionName = callSite.getFunctionName();
+      var methodName = callSite.getMethodName();
+      var name = (methodName && functionName) ?
+        functionName + ' [as ' + methodName + ']' :
+        functionName || methodName || '<anonymous function>';
+      stackFrames.push(new StackFrame(undefined, name,
+        callSite.getFileName(), callSite.getLineNumber(),
+        callSite.getColumnNumber()));
+    });
+    this.span.setLabel(TraceLabels.STACK_TRACE_DETAILS_KEY,
+      JSON.stringify({stack_frame: stackFrames}));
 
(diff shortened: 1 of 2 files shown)


lib/span-data.js
```diff
@@ -82,6 +92,23 @@ SpanData.prototype.close = function() {
...
 /**
+ * Trace API expects stack frames to be a JSON string with the following
+ * structure:
+ * STACK_TRACE := { "stack_frame" : [ FRAMES ] }
+ * FRAMES := { "class_name" : CLASS_NAME, "file_name" : FILE_NAME,
+ *             "line_number" : LINE_NUMBER, "method_name" : METHOD_NAME }*
+ * @constructor @private
+ */
+function StackFrame(className, methodName, fileName, lineNumber, columnNumber) {
+  if (className)   { this.class_name = className;  }
```
matthewloring: Why are these if's necessary? Does the visible behavior change if they are not included?

matthewloring:
Fixes #19

lib/span-data.js
ofrobots: My thinking was that I want to leave the property undefined, as opposed to being set to null, if null is passed in; this reduces the data sent on the network. For example, className will always be null in our case. Upon reflection, this does the wrong thing for lineNumber or columnNumber when they are 0. I'll try to fix this.

ofrobots:
@matthewloring ready for another look.

lib/span-data.js
```diff
@@ -82,6 +92,42 @@ SpanData.prototype.close = function() {
 };
 
 /**
+ * Trace API expects stack frames to be a JSON string with the following
+ * structure:
+ * STACK_TRACE := { "stack_frame" : [ FRAMES ] }
+ * FRAMES := { "class_name" : CLASS_NAME, "file_name" : FILE_NAME,
+ *             "line_number" : LINE_NUMBER, "method_name" : METHOD_NAME }*
```
matthewloring: Nit: could we include columnNumber in this graphic as it is a member of the class.

lib/span-data.js
```diff
@@ -82,6 +92,42 @@ SpanData.prototype.close = function() {
...
+ * @param {string|undefined} className
+ * @param {string|undefined} methodName
+ * @param {string|undefined} fileName
+ * @param {number|undefined} lineNumber
+ * @param {number|undefined} columnNumber
+ * @constructor @private
+ */
+function StackFrame(className, methodName, fileName, lineNumber, columnNumber) {
+  if (className) {
+    this.class_name = className;
```
matthewloring: Note to future me: This path will be dead code as we currently do not provide class names. It is included to be faithful to the structure expected by the server.

matthewloring:
LGTM.

PR #61, opened 2015-09-10
Patched express on public interface

matthewloring:
Unit and non-interference tests pass with these changes. I don't believe it is necessary to update unit tests with full coverage of the method interface as the code paths are fully shared. @ofrobots PTAL.


ofrobots:
LGTM.

matthewloring:
Fixes #61

PR #64, opened 2015-09-15
added mocha style run in span operation to agent

matthewloring:


matthewloring:
@ofrobots PTAL.

matthewloring:
@ofrobots Your comments seem to have disappeared from github. I've added them here:
1) Make fn the last argument to runInSpan.
Response: My thought was that in Node the pattern of taking function arguments (some potentially being optional) followed by a callback is common. However, in this case the function isn't a callback but is semantically an argument to the function so it made sense to put it before the optional argument for labels.
2) Test adding labels in endSpan.
Response: Done.

ofrobots:
Having labels last makes the interface similar to `setTimeout`, which is generally hated by everyone, myself included. From a readability perspective this would have been so much nicer:
```
setTimeout(5000, function() {
  // blargh
  // blargh-de-blargh
  // et tu, blargh?
  // blargh!
});
```

Similarly, logically it makes more sense to the labels to be next to the name of the transaction, as opposed to separated by an arbitrarily long function.

matthewloring:
@ofrobots Done!

ofrobots:
LGTM.

PR #69, opened 2015-09-17
Accept labels as objects on start and end span

matthewloring:
Previously, labels were passed to start and end span calls
with the structure [{ key: 'key', value: 'val' }]. This has been
replaced with the more concise { key: 'val' } structure.

@ofrobots PTAL.


ofrobots:
LGTM

PR #71, opened 2015-09-21
Fixes null reference in express

### LINKED ISSUE #70: Exception: cannot read property bindEmitter of Null
Run into this just now. Seems like we need a test for the path where we fail to get a project number. We need to completely disable the agent in such cases.

```shell
~/src/bookshelf2/bookshelf 2-structured-data
❯ node app.js
App listening at http://:::8080
ERROR:@google/cloud-trace: Unable to acquire the project number from metadata service. Please provide a valid project number as an env. variable, or through config.projectId passed to start().Error: Could not auto-discover project-id. Please export GCLOUD_PROJECT_NUM with your numeric project id
TypeError: Cannot read property 'bindEmitter' of null
    at middleware (/Users/ofrobots/src/bookshelf2/bookshelf/node_modules/@google/cloud-trace/lib/hooks/userspace/hook-express.js:40:12)
    at Layer.handle [as handle_request] (/Users/ofrobots/src/bookshelf2/bookshelf/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/Users/ofrobots/src/bookshelf2/bookshelf/node_modules/express/lib/router/index.js:312:13)
    at /Users/ofrobots/src/bookshelf2/bookshelf/node_modules/express/lib/router/index.js:280:7
    at Function.process_params (/Users/ofrobots/src/bookshelf2/bookshelf/node_modules/express/lib/router/index.js:330:12)
    at next (/Users/ofrobots/src/bookshelf2/bookshelf/node_modules/express/lib/router/index.js:271:10)
    at expressInit (/Users/ofrobots/src/bookshelf2/bookshelf/node_modules/express/lib/middleware/init.js:33:5)
    at Layer.handle [as handle_request] (/Users/ofrobots/src/bookshelf2/bookshelf/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/Users/ofrobots/src/bookshelf2/bookshelf/node_modules/express/lib/router/index.js:312:13)
    at /Users/ofrobots/src/bookshelf2/bookshelf/node_modules/express/lib/router/index.js:280:7
```

matthewloring:
A null reference was occuring when the agent disabled after
attaching express middleware. Middleware now verifies that
the agent is enabled before starting the trace.

Fixes #70 

@ofrobots PTAL

### DIFF
--- a/lib/hooks/userspace/hook-express.js
+++ b/lib/hooks/userspace/hook-express.js
@@ -36,11 +36,14 @@ function applicationActionWrap(method) {
 }
 
 function middleware(req, res, next) {
-  var originalEnd = res.end;
-
   var namespace = cls.getNamespace();
+  if (!namespace) {
+    return next();
+  }
   namespace.bindEmitter(req);
   namespace.bindEmitter(res);
+
+  var originalEnd = res.end;
 
   namespace.run(function() {
     var rootContext = startRootSpanForRequest(req);
--- a/lib/hooks/userspace/hook-hapi.js
+++ b/lib/hooks/userspace/hook-hapi.js
@@ -32,13 +32,17 @@ function connectionWrap(connection) {
 }
 
 function middleware(request, reply) {
+  var namespace = cls.getNamespace();
+  if (!namespace) {
+    return reply.continue();
+  }
   var req = request.raw.req;
   var res = request.raw.res;
-  var originalEnd = res.end;
 
-  var namespace = cls.getNamespace();
   namespace.bindEmitter(req);
   namespace.bindEmitter(res);
+
+  var originalEnd = res.end;
 
   namespace.run(function() {
     var rootContext = startRootSpanForRequest(req);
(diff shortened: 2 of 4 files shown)


ofrobots:
It seems that in other hooks (e.g. [hook-hapi.js](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/lib/hooks/userspace/hook-hapi.js#L40)), we are doing a similar unguarded dereference of `namespace`. Shouldn't we fix this issue in the other hooks at the same time?

ofrobots:
I picked up just the test, and not the fix, and it doesn't fail for me. This probably timing dependent. Is there a way we could make the test more reliable, or figure out another mechanism to test this race?

matthewloring:
I'm worried that it doesn't fail. It is failing 100% of the time for me.

matthewloring:
Could be environment dependent?

matthewloring:
Actually, it just failed 1/10 runs. I'll try to make it more deterministic.

matthewloring:
@ofrobots The test has been made non-flaky and now covers the other frameworks where middleware is installed.

ofrobots:
LGTM.

PR #73, opened 2015-09-23
Check env vars for enabling and log levels

### LINKED ISSUE #72: Ability to disable the trace agent using an env. var.
It would be nice to be able to disable the trace agent without having to modify the source. I am thinking an env. var. named `GCLOUD_TRACE_DISABLE`, which if set, would cause the trace agent to not start.

Similarly, it would be nice to have an environment variable that increases the log-level for the trace agent.

matthewloring:
Added two new flags:

  GCLOUD_TRACE_DISABLE: boolean, if set the agent will not enable
  GCLOUD_TRACE_LOGGING: numeric, sets the trace log level

Fixes #72

### DIFF
--- a/config.js
+++ b/config.js
@@ -18,9 +18,10 @@
 // Default configuration
 module.exports = {
   // Log levels: 0-disabled,1-error,2-warn,3-info,4-debug
-  logLevel: 1,
+  logLevel: process.env.hasOwnProperty('GCLOUD_LOG_LEVEL') ?
+    process.env.GCLOUD_LOG_LEVEL : 1,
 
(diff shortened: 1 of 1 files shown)


matthewloring:
@ofrobots PTAL

ofrobots:
Could you do this config.js, like the Debug agent? The config object is the [SSOT](https://en.wikipedia.org/wiki/Single_Source_of_Truth).

matthewloring:
@ofrobots Moved to config. PTAL.

ofrobots:
LGTM.

PR #75, opened 2015-09-23
Moved methods dep from dev to normal dependency

matthewloring:
It is required for the express hook. This wasn't caught by CI since all dev dependencies are installed there.

@ofrobots PTAL


ofrobots:
Please fix the commit title to be 50 characters at the most. Otherwise LGTM.

PR #78, opened 2015-09-25
Make log env var consistent with debug

matthewloring:


matthewloring:
@ofrobots PTAL.

ofrobots:
LGTM.

PR #79, opened 2015-09-27
Check framework versions using package.json

### LINKED ISSUE #24: Ability to find the package.json from a loaded module
We want to do version checks to make, e.g. that we are patching express v4+. The current mechanism of looking for internal properties that we know have been removed in v4 is a bit too brittle for my taste. It would be much better if we can figure out where the package.json for a loaded module is, and do a check against the version there.

matthewloring:
This patch uses module internals to load the package.json file
associated with required modules and uses them to determine module
versions instead of checking for the presence of non-public facing
api functions.

Fixes #24.

### DIFF
--- a/lib/hooks/index.js
+++ b/lib/hooks/index.js
@@ -18,7 +18,7 @@
 var Module = require('module');
 var shimmer = require('shimmer');
 var assert = require('assert');
-
+var path = require('path');
 
 //
 // All these operations need to be reversible
@@ -36,7 +36,8 @@ var toInstrument = Object.create(null, {
     value: {
       file: './userspace/hook-express.js',
       module: null,
-      hook: null
+      hook: null,
+      version: null
     }
   },
   'hapi': {
@@ -44,7 +45,8 @@ var toInstrument = Object.create(null, {
     value: {
       file: './userspace/hook-hapi.js',
       module: null,
-      hook: null
+      hook: null,
+      version: null
     }
   },
   'http': {
@@ -52,7 +54,8 @@ var toInstrument = Object.create(null, {
     value: {
       file: './core/hook-http.js',
       module: null,
-      hook: null
+      hook: null,
+      version: null
     }
   },
   'mongodb-core': {
@@ -60,7 +63,8 @@ var toInstrument = Object.create(null, {
     value: {
       file: './userspace/hook-mongodb-core.js',
       module: null,
-      hook: null
+      hook: null,
+      version: null
     }
   },
   'redis': {
@@ -68,7 +72,8 @@ var toInstrument = Object.create(null, {
     value: {
       file: './userspace/hook-redis.js',
       module: null,
-      hook: null
+      hook: null,
+      version: null
     }
   },
   'restify': {
@@ -76,10 +81,26 @@ var toInstrument = Object.create(null, {
     value: {
       file: './userspace/hook-restify.js',
       module: null,
-      hook: null
+      hook: null,
+      version: null
     }
   }
 });
+
+function findModuleVersion(request, parent, load) {
+  var mainScriptDir = path.dirname(require.resolve(request));
+  var resolvedModule = Module._resolveLookupPaths(request, parent);
+  var paths = resolvedModule[1];
+  for (var i = 0, PL = paths.length; i < PL; i++) {
+    if (mainScriptDir.indexOf(paths[i]) === 0) {
+      // This is unsafe on windows if request contains a namespace
+      // TODO: convert path separators in request
+      var pjson = load(path.join(paths[i], request, 'package.json'));
+      return pjson.version;
+    }
+  }
+  return null;
+}
 
 function activate(agent) {
 
@@ -90,7 +111,7 @@ function activate(agent) {
     // Load the hook. This file, i.e. index.js, becomes the parent module.
     var moduleHook = loadFn(instrumentation.file, module, false);
     instrumentation.hook = moduleHook;
-    moduleHook.patch(instrumentation.module, agent);
+    moduleHook.patch(instrumentation.module, agent, instrumentation.version);
   }
 
   // hook into Module._load so that we can hook into userspace frameworks
@@ -116,6 +137,8 @@ function activate(agent) {
         if (instrumentation && agent.config().excludedHooks &&
             agent.config().excludedHooks.indexOf(request) === -1 &&
             !instrumentation.hook) { // not already patched.
+          instrumentation.version = findModuleVersion(request, arguments[1],
+              originalModuleLoad);
           instrumentation.module = loaded;
           loadHookAndPatch(originalModuleLoad, instrumentation);
         }
@@ -142,7 +165,8 @@ function deactivate() {
 
 module.exports = {
   activate: activate,
-  deactivate: deactivate
+  deactivate: deactivate,
+  findModuleVersion: findModuleVersion
 };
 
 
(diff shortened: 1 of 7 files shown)


matthewloring:
@ofrobots I noticed a weird behavior with npm. If `"hapi": "^8.8.0"` is specified in the `package.json` file, npm was actually installing `8.8.1`. Does npm automatically try to get your the highest non-breaking revision of the package or do you know of another mechanism that explains this?

ofrobots:
That's the expected behaviour. This [article](https://nodesource.com/blog/semver-tilde-and-caret) might be good reading on what caret means in node-semver.

matthewloring:
I've updated to check only for major version match. PTAL.

lib/hooks/userspace/hook-express.js
```diff
@@ -101,12 +103,8 @@ function endRootSpanForRequest(rootContext, req, res) {
...
 
 module.exports = {
-  patch: function(express_, agent_) {
-    // We only support express v4. In the past express exported a version property
-    // that would have been great for this purpose, but v3.x removed it. Instead
-    // we rely on 'createServer' export that was removed in v4.
-    // See: https://github.com/strongloop/express/wiki/Migrating-from-3.x-to-4.x
-    if (typeof module.createServer !== 'undefined') {
+  patch: function(express_, agent_, version_) {
+    if (SUPPORTED_VERSIONS.indexOf(version_.split('.')[0]) === -1) {
```
ofrobots: Use the [semver](https://www.npmjs.com/package/semver) module instead of hand-rolling this.

matthewloring: Was hoping to avoid the overhead of a require. Using this module now.

lib/hooks/index.js
```diff
@@ -116,6 +122,16 @@ function activate(agent) {
         if (instrumentation && agent.config().excludedHooks &&
             agent.config().excludedHooks.indexOf(request) === -1 &&
             !instrumentation.hook) { // not already patched.
+          var mainScriptDir = path.dirname(require.resolve(request));
```
ofrobots: Can you extract this in a function of its own.

lib/hooks/index.js
```diff
@@ -116,6 +122,16 @@ function activate(agent) {
...
+          var resolvedModule = Module._resolveLookupPaths.apply(this, arguments);
+          var paths = resolvedModule[1];
+          for (var i = 0, PL = paths.length; i < PL; i++) {
+            if (mainScriptDir.indexOf(paths[i]) === 0) {
+              var pjson = originalModuleLoad(path.join(paths[i], request, 'package.json'));
```
ofrobots: This might throw. Not sure if we are 100% confidence that it can't. I am still digesting the code, so I might get to 100%.

lib/hooks/index.js
matthewloring: Extracted and tested.

lib/hooks/index.js
matthewloring: It can throw on windows (I think). I've added a todo.

matthewloring:
@ofrobots I'm not attached to the `versions.js` file and I'm happy to rely on `semver` logic. I'm just hoping to get a nice enumerable list of versions to test against. If there is a nicer way of accomplishing this I'm happy to switch to that.

test/hooks/test-hooks-index.js
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+var assert = require('assert');
+var Module = require('module');
+var findModuleVersion = require('../../lib/hooks/index.js').findModuleVersion;
+
+describe('findVersion', function() {
```
ofrobots: findModuleVersion

test/hooks/test-hooks-index.js
```diff
@@ -0,0 +1,39 @@
...
+  it('should correctly find package.json for userspace packages', function() {
+    assert.equal(findModuleVersion('express', module, Module._load), '4.13.3');
```
ofrobots: Hard-coded version number means that we need to keep this and package.json in sync. Could we do this some other way?

ofrobots:
Another problem with the versions.js file it is too rigid in how support is specified. We cannot limit support to subset of a major release. Can you elaborate on how you plan to use the enumerable versions to test against?

test/hooks/test-hooks-index.js
matthewloring: We could use the versions file and check that we loaded a compatible version. Alternatively, we could have a separate package.json that loads precise versions of certain modules. That would require a separate npm install but it would be easier to keep it up to date in parallel with the test. What we really want is the ability to require a particular version of a module or require by github url but that functionality isn't available yet.

test/hooks/test-hooks-index.js
matthewloring: Done.

matthewloring:
The structure of the file could be augmented to incorporate supported minor versions as well. The hope is to get an enumerable list as opposed to a rule so that the tests can dynamically create a project in fixtures for each supported version (as opposed to having to keep them updated in parallel). It would be easiest if the projects could be configured by iterating over versions and requiring package@version for each supported version instead of trying to determine the set of published versions satisfying a rule (likes >4).

matthewloring:
As per offline, `version.js` has been removed, we now compare against a semver expression stored per hook. We will defer the updates to the `findModuleVersion` tests until we separate modules we test from the dev dependencies.

lib/hooks/userspace/hook-express.js
```diff
@@ -19,12 +19,15 @@
...
 var TraceLabels = require('../../trace-labels.js');
 var shimmer = require('shimmer');
+var semver = require('semver');
 var patchedMethods = require('methods');
 patchedMethods.push('use', 'route', 'param', 'all');
 var constants = require('../../constants.js');
 var agent;
 var express;
 
+var SUPPORTED_VERSIONS = '4.13.x';
```
ofrobots: Why so strict? We don't support 4.12?

lib/hooks/userspace/hook-hapi.js
```diff
@@ -19,10 +19,13 @@
 var cls = require('../../cls.js');
 var TraceLabels = require('../../trace-labels.js');
 var shimmer = require('shimmer');
+var semver = require('semver');
 var constants = require('../../constants.js');
 var agent;
 var hapi;
 
+var SUPPORTED_VERSIONS = '8.8.x';
```
ofrobots: Likewise.

lib/hooks/userspace/hook-mongodb-core.js
```diff
@@ -18,9 +18,12 @@
 
 var cls = require('../../cls.js');
 var shimmer = require('shimmer');
+var semver = require('semver');
 var agent;
 var mongo;
 
+var SUPPORTED_VERSIONS = '1.2.x';
```
ofrobots: Likewise.

lib/hooks/userspace/hook-redis.js
```diff
@@ -18,9 +18,12 @@
 
 var cls = require('../../cls.js');
 var shimmer = require('shimmer');
+var semver = require('semver');
 var agent;
 var redis;
 
+var SUPPORTED_VERSIONS = '0.12.x';
```
ofrobots: Likewise.

matthewloring:
I can relax all of these major versions (except redis) but we have not yet ever tested against another other than the minors included above. I was hoping to relax them once we had tests in place to validate that we do in fact support them. Redis is different because with it still being in version `0.x` there are api breaking changes in previous minor versions if my understanding of `semver` is correct.

ofrobots:
LGTM if we can address #13 next and relax the versions listed in this CL.

PR #80, opened 2015-09-29
Tracing support for mysql

### LINKED ISSUE #53: Add tracing support for mysql

matthewloring:
Fixes #53.

### DIFF
--- a/lib/hooks/userspace/hook-mysql.js
+++ b/lib/hooks/userspace/hook-mysql.js
@@ -1,0 +1,78 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+var cls = require('../../../lib/cls.js');
+var shimmer = require('shimmer');
+var semver = require('semver');
+var SpanData = require('../../../lib/span-data.js');
+var agent;
+
+var SUPPORTED_VERSIONS = '^2.9.x';
+
+function createQueryWrap(createQuery) {
+  return function createQuery_trace(sql, values, cb) {
+    var root = cls.getRootContext();
+    if (!root) {
+      agent.logger.warn('Cannot trace mysql outside of a supported framework.');
+      return createQuery.apply(this, arguments);
+    } else if (root === SpanData.nullSpan) {
+      return createQuery.apply(this, arguments);
+    }
+    var span = agent.startSpan('mysql-query');
+    var query = createQuery.apply(this, arguments);
+    span.addLabel('sql', query.sql);
+    cls.getNamespace().bindEmitter(query);
+    if (query._callback) {
+      query._callback = wrapCallback(span, query._callback);
+    } else {
+      query.on('end', function() {
+        agent.endSpan(span);
+      });
+    }
+    return query;
+  };
+}
+
+function wrapCallback(span, done) {
+  var fn = function(err, res) {
+    agent.endSpan(span);
+    if (done) {
+      done(err, res);
+    }
+  };
+  return cls.getNamespace().bind(fn);
+}
+
+module.exports = function(version_, agent_) {
+  if (!semver.satisfies(version_, SUPPORTED_VERSIONS)) {
+    agent_.logger.info('Mysql: unsupported version ' + version_ + ' loaded');
+    return {};
+  }
+  return {
+    'lib/Connection.js': {
+      patch: function(Connection) {
+        agent = agent_;
+        shimmer.wrap(Connection, 'createQuery', createQueryWrap);
+      },
+      unpatch: function(Connection) {
+        shimmer.unwrap(Connection, 'createQuery');
+        agent = null;
+      }
+    }
+  };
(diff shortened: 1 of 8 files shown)


matthewloring:
@ofrobots PTAL. The loss of context was a lot more apparent in the unit tests for mysql than it was for mongo. The primary culprit were the events being emitted on the query object so the patch moved from from the protocol level to the connection level where the query is constructed.

lib/hooks/userspace/hook-mysql.js
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+'use strict';
+
+var cls = require('../../../lib/cls.js');
+var shimmer = require('shimmer');
+var semver = require('semver');
+var agent;
+
+var SUPPORTED_VERSIONS = '2.9.x';
```
ofrobots: Can this be relaxed to be `'>= 2.9.x'`?

matthewloring: I believe it can be relaxed as far as `>= 2.x`. It is more a philosophical question of how/when we want to catch hook breaking errors. The current form makes it more likely that they're caught in our unit tests and less likely they would be reported as user facing errors. I'm happy to change it if we're ok with less test coverage of supported versions.

lib/hooks/userspace/hook-mysql.js
```diff
@@ -0,0 +1,68 @@
...
+
+function createQueryWrap(createQuery) {
+  return function createQuery_trace(sql, values, cb) {
+    var root = cls.getRootContext();
+    if (!root) {
+      agent.logger.warn('Cannot trace mysql outside of a supported framework.');
+      return createQuery.apply(this, arguments);
+    }
+    var query = createQuery.apply(this, arguments);
+    var span = agent.startSpan('mysql-query', { sql: query.sql });
```
ofrobots: This doesn't count the time spent in createQuery. This is probably negligible, but it would be still nice to count it if it is possible.

Secondly, should the sql label be guarded with a `config.enhancedDatabaseReporting` check?

lib/hooks/userspace/hook-mysql.js
```diff
@@ -0,0 +1,68 @@
...
+    cls.getNamespace().bindEmitter(query);
+    query._callback = wrapCallback(span, query._callback);
```
ofrobots: Patching a private property looks too intrusive. Can you instead wrap `cb` upfront before passing it to `Connection.createQuery`?

ofrobots:
Need a few more tests exercising the following:
* Use events instead of callback for the query.
* Use events and callback.
* Neither.
* Run some queries inside database transactions and exercise commits and aborts.

lib/hooks/userspace/hook-mysql.js
matthewloring: We should definitely capture time spent in `createQuery`. Fixed.

I didn't think that a `config.enhancedDatabaseReporting` check was necessary here because the sql query is just a template that substitutes in user provided values later. Looking back, I am not as confident that this is _always_ the case. I was going to wait to add the guard check until the PR introducing `config.enhancedDatabaseReporting` landed but I can add it now since there isn't a type checker to complain about using an undefined property *grumble grumble*...

lib/hooks/userspace/hook-mysql.js
matthewloring: `cb` isn't guaranteed to be defined at this point. If we don't want to access the private property we will have to duplicate the callback resolution logic inside `createQuery`. I thought that accessing the private property was safer than maintaining the resolution logic in parallel here. If you would prefer the other way, it can be changed.

matthewloring:
I will expand testing to cover these cases except for events and callback. Quoted from their readme: "You MUST NOT provide a callback to the query() method when streaming rows."

lib/hooks/userspace/hook-mysql.js
matthewloring: Actually, I think I'm going to hold off on guarding this until we have enhanced database reporting landed so that I don't have parallel test harness fixes in both PRs.

matthewloring:

lib/hooks/userspace/hook-mysql.js
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+'use strict';
+
+var cls = require('../../../lib/cls.js');
+var shimmer = require('shimmer');
+var semver = require('semver');
+var agent;
+
+var SUPPORTED_VERSIONS = '~2.9.x';
```
ofrobots: I was wrong. This should be '^2.9.x'

lib/hooks/userspace/hook-mysql.js
```diff
@@ -0,0 +1,75 @@
...
+  return function createQuery_trace(sql, values, cb) {
+    var root = cls.getRootContext();
+    if (!root) {
+      agent.logger.warn('Cannot trace mysql outside of a supported framework.');
+      return createQuery.apply(this, arguments);
+    }
+    var span = agent.startSpan('mysql-query');
+    var query = createQuery.apply(this, arguments);
+    span.addLabel('sql', query.sql);
+    cls.getNamespace().bindEmitter(query);
```
ofrobots: Should this be done if they provided a callback?

matthewloring: We need to do this to maintain context. Even if we don't provide a callback, the query object has internal only event emitters attached to it to manage connection related lifecycle. See https://github.com/felixge/node-mysql/blob/master/lib/protocol/Protocol.js#L142

lib/hooks/userspace/hook-mysql.js
matthewloring: Done.

ofrobots:
LGTM. We need e2e tests for this. That is a separate issue. Before landing can you run https://cloud.google.com/nodejs/getting-started/using-cloud-sql manually at least to make sure this looks sensible?

matthewloring:
The instructions here produced some good mysql spans. Landing this.

matthewloring:
I also added a span leak test to `test-hooks-sample-warning.js`

PR #81, opened 2015-09-29
Run integration tests on travis

matthewloring:


matthewloring:
@ofrobots PTAL. I've excluded node 4 from the express and restify tests as they do not pass even with our instrumentation. In a follow on PR, i can check failing tests with and without instrumentation and ensure that they are equal allowing us to run on all versions.

ofrobots:
>  I've excluded node 4 from the express and restify tests as they do not pass even with our instrumentation.

Did you mean 'without our instrumentation'?

matthewloring:
Yes. Fixed above.

ofrobots:
LGTM.

PR #83, opened 2015-09-30
Resolve all module scripts correctly

### LINKED ISSUE #82: require.resolve doesn't find all modules
We patch modules that may not be on the load path for the app running the agent (mongodb-core may not be on the load path if the app is using mongoose). We need to load these uniformly.

matthewloring:
The module file finder is able to search with respect to a
provided parent module allowing us to find modules that are
not in the load path of the primary application module.

Fixes #82

### DIFF
--- a/lib/hooks/index.js
+++ b/lib/hooks/index.js
@@ -88,7 +88,7 @@ var toInstrument = Object.create(null, {
 });
 
 function findModuleVersion(request, parent, load) {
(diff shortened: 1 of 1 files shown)


matthewloring:
@ofrobots PTAL

ofrobots:
Is there an issue open to add a regression test for this? Otherwise LGTM.

matthewloring:
Yes, #84

PR #86, opened 2015-09-30
Remove frameworks from dev dependencies

### LINKED ISSUE #13: Don't use dev dependencies for instrumented frameworks

matthewloring:
All instrumented frameworks are now installed inside
test/hooks/fixtures and are no longer included in the
dev dependencies of the primary module.

Fixes #13


matthewloring:
@ofrobots Currently we will install dev fixtures even when we are required by another module. Do you know of an npm hook that will only be triggered if dev dependencies should be installed?

.npmignore
```diff
@@ -0,0 +1,10 @@
+/bin
+/node_modules
+/test
+.editorconfig
+.gitignore
+.jshintignore
+.jshintrc
+.npmignore
+.travis.yml
+CONTRIBUTING.md
```
ofrobots: Do not remove `CONTRIBUTING.md`. `node_modules` is [handled by npm automatically](https://docs.npmjs.com/misc/developers). Infact, it would be better to keep it minimal. `test` is the main thing we want to avoid packaging in.

matthewloring:
@ofrobots .npmignore has been cleaned out to only exclude test.

bin/run-test.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 
 jshint . || exit 1
+for dir in test/hooks/fixtures/*/ ;
+do
+  cd ${dir} && npm install && cd -
```
ofrobots: Add `echo "npm install in ${dir}"`.
For bonus marks, add `echo -en "travis_fold:start:npm install in ${dir}\\r"` along with an matching end.

ofrobots:
LGTM w/ suggestion for travis readability.

PR #90, opened 2015-10-02
Relaxed required version checks to major version only

matthewloring:
Since Redis is still in 0.x, I'm leaving minor version requirement there as there could be breaking changes on minor differences.

### DIFF
--- a/lib/hooks/userspace/hook-express.js
+++ b/lib/hooks/userspace/hook-express.js
@@ -26,7 +26,7 @@ var constants = require('../../constants.js');
 var agent;
 var express;
 
-var SUPPORTED_VERSIONS = '4.13.x';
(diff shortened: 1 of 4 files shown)


matthewloring:
@ofrobots PTAL

ofrobots:
Me so confused. [This redis](https://www.npmjs.com/package/redis) seems to be at 2.x? Looking at [our source](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/lib/hooks/index.js#L70), we are indeed looking for that redis.

I might be missing something :confused:.

matthewloring:
Redis went from 0.12 to 2.1 in the last month. We can update to redis 2 in a later PR.

ofrobots:
LGTM.

PR #91, opened 2015-10-02
Enable trace throttling

matthewloring:
This also fixes a bug where we were ignoring `bufferSize` in config due to a typo. Arrggg.

### DIFF
--- a/config.js
+++ b/config.js
@@ -34,6 +34,12 @@ module.exports = {
   // See `bufferSize`.
   flushDelaySeconds: 30,
 
+  // If false, traces will be published as soon as `bufferSize` traces have
+  // been collected if this occurs before `flushDelaySeconds`.
+  // If true, traces will flush every `flushDelaySeconds` and additional
+  // traces beyond `bufferSize` will be dropped.
+  samplingEnabled: false,
+
   // The number of transactions we buffer before we publish to the Cloud Trace
   // API, unless we hit `flushDelaySeconds` first.
   bufferSize: 1000
--- a/lib/trace-writer.js
+++ b/lib/trace-writer.js
@@ -86,12 +86,16 @@ TraceWriter.prototype.queueTrace_ = function(trace) {
     }
 
     trace.projectId = project;
-    var length = that.buffer_.push(JSON.stringify(trace));
 
-    that.logger_.info('queued trace. new size:', length);
+    if (that.buffer_.length <= that.config_.bufferSize) {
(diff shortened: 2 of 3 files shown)


matthewloring:
@ofrobots PTAL

config.js
```diff
@@ -34,6 +34,12 @@ module.exports = {
   // See `bufferSize`.
   flushDelaySeconds: 30,
 
+  // If true, traces will be published as soon as `bufferSize` traces have
+  // been collected if this occurs before `flushDelaySeconds`.
+  // If false, traces will flush every `flushDelaySeconds` and additional
+  // traces beyond `bufferSize` will be dropped.
+  publishImmediately: true,
+
```
ofrobots: I don't like the property name too much as it is not describe what it actually does. `publishOnBufferFull` would be more better, but even that doesn't describe the behaviour that the subsequent traces will be dropped.

matthewloring: how about `samplingEnabled`? That captures the idea that traces will be lost and the description above it gives precise semantics.

lib/trace-writer.js
```diff
@@ -83,12 +83,16 @@ TraceWriter.prototype.queueTrace_ = function(trace) {
     if (err) { return; } // ignore as index.js takes care of this.
 
     trace.projectId = project;
-    var length = that.buffer_.push(JSON.stringify(trace));
 
-    that.logger_.info('queued trace. new size:', length);
+    if (that.buffer_.length <= that.config_.bufferSize) {
+      that.buffer_.push(JSON.stringify(trace));
+      that.logger_.info('queued trace. new size:', that.buffer_.length);
```
ofrobots: Change this to be logger_.debug. This is going to be too high volume to be enabled at info level.

matthewloring: Done.

config.js
```diff
@@ -34,6 +34,12 @@ module.exports = {
   // See `bufferSize`.
   flushDelaySeconds: 30,
 
+  // If true, traces will be published as soon as `bufferSize` traces have
+  // been collected if this occurs before `flushDelaySeconds`.
+  // If false, traces will flush every `flushDelaySeconds` and additional
+  // traces beyond `bufferSize` will be dropped.
+  samplingEnabled: true,
```
ofrobots: When `samplingEnabled` we trace everything? That sounds wrong.

ofrobots:
LGTM if you can reverse what samplingEnabled being true means.

PR #93, opened 2015-10-03
Use protobufs to communicate with trace api

matthewloring:
Protos were imported from:
  https://github.com/google/googleapis
  https://github.com/google/protobuf

### DIFF
--- a/lib/trace-span.js
+++ b/lib/trace-span.js
@@ -19,17 +19,28 @@
 // TODO(ofrobots): replace this file with the protobuf
 
 /**
+ * Converts the provided time in millis into a seconds/nanos
+ * style timestamp.
+ */
+var getTimestamp = function(time) {
+  return {
+    seconds: Math.floor(time / 1000),
+    nanos: Math.floor((time % 1000) * 1000000)
+  };
+};
+
+/**
  * Creates a trace span object.
  * @constructor
  */
 function TraceSpan(name, spanId, parentSpanId, time) {
   this.name = name;
-  this.parentSpanId = parentSpanId;
-  this.spanId = spanId;
-  this.kind = 'RPC_CLIENT';
+  this.parent_span_id = parentSpanId;
+  this.span_id = spanId;
+  this.kind = 2; // The enum representation of RPC_CLIENT, see trace.proto
   this.labels = {};
-  this.startTime = (time || new Date()).toISOString();
-  this.endTime = '';
+  this.start_time = getTimestamp(time || Date.now());
+  this.end_time = null;
 }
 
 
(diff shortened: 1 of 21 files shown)


matthewloring:
@ofrobots This is passing!!!

third_party/proto/google/api/annotations.proto
```diff
@@ -0,0 +1,29 @@
...
+import "../../google/protobuf/descriptor.proto";
+
+option java_multiple_files = true;
+option java_outer_classname = "AnnotationsProto";
+option java_package = "com.google.api";
+
+extend google.protobuf.MethodOptions {
+  // See `HttpRule`.
+  HttpRule http = 72295728;
+}
```
ofrobots: missing new line at end

ofrobots: But this is a copied file. Nevermind if the original doesn't have it.

lib/trace.js
```diff
@@ -23,8 +23,8 @@
  * @constructor
  */
 function Trace(projectId, traceId) {
-  this.projectId = projectId;
-  this.traceId = traceId;
+  this.project_id = projectId;
+  this.trace_id = traceId;
```
ofrobots: Do we even need this file anymore? Can't we construct the protobuf class directly?

lib/trace-span.js
```diff
@@ -19,17 +19,28 @@
...
+    seconds: Math.floor(time / 1000),
+    nanos: Math.floor((time % 1000) * 1000000)
+  };
+};
+
+/**
  * Creates a trace span object.
  * @constructor
  */
 function TraceSpan(name, spanId, parentSpanId, time) {
```
ofrobots: Similarly, I don't think we need this class anymore.

third_party/proto/google/api/annotations.proto
matthewloring: The original does not have it.

lib/trace.js
matthewloring: We can, I'd like to land a functionality CL as this is already getting really big and then independently land a refactor that removes our structural classes.

lib/trace-span.js
matthewloring: I agree. I can get rid of our structural classes now if we don't mind making this CL larger. Or I can do it in the next one.

ofrobots: That works for me.

ofrobots: To clarify; doing it in the next one is fine.

ofrobots:
The changes here look good, but this should not be landed yet. The pre-requisites would be:
1. grpc release w/ the necessary support.
2. e2e testing infrastructure should be functional enough to test this.
3. performance evaluation.

matthewloring:
After the performance review done by @cristiancavalli, it seems that grpc is not a feasible solution at this time.

PR #94, opened 2015-10-05
A more complete solution for sampling

### LINKED ISSUE #95: Span objects should not be created when they will not be reported
When sampling is enabled, we should only create span objects for the traces we record.

matthewloring:
Fixes #95.

### DIFF
--- a/lib/hooks/userspace/hook-express.js
+++ b/lib/hooks/userspace/hook-express.js
@@ -84,8 +84,8 @@ function startRootSpanForRequest(req) {
   // url as a label
   var rootContext = agent.createRootSpanData(req.originalUrl, traceId,
     parentSpanId);
-  rootContext.span.setLabel(TraceLabels.HTTP_METHOD_LABEL_KEY, req.method);
-  rootContext.span.setLabel(TraceLabels.HTTP_URL_LABEL_KEY, url);
+  rootContext.addLabel(TraceLabels.HTTP_METHOD_LABEL_KEY, req.method);
+  rootContext.addLabel(TraceLabels.HTTP_URL_LABEL_KEY, url);
   return rootContext;
 }
 
@@ -98,12 +98,12 @@ function startRootSpanForRequest(req) {
  */
 function endRootSpanForRequest(rootContext, req, res) {
   if (req.route && req.route.path) {
-    rootContext.span.setLabel(
+    rootContext.addLabel(
       'express/request.route.path', req.route.path);
   }
-  rootContext.span.setLabel(
+  rootContext.addLabel(
       TraceLabels.HTTP_RESPONSE_CODE_LABEL_KEY, res.statusCode);
-  rootContext.close();
+  rootContext.spanData_.close();
 }
 
 module.exports = {
--- a/lib/trace-agent.js
+++ b/lib/trace-agent.js
@@ -24,6 +24,7 @@ var SpanData = require('./span-data.js');
 var TraceWriter = require('./trace-writer.js');
 var uuid = require('uuid');
 var constants = require('./constants.js');
+var tracingPolicy = require('./tracing-policy.js');
 
 /** @type {TraceAgent} */
 var traceAgent;
@@ -39,6 +40,8 @@ function TraceAgent(config, logger) {
 
   this.namespace = cls.createNamespace();
   this.traceWriter = new TraceWriter(logger, config);
+
+  this.policy = tracingPolicy.createTracePolicy(config);
 
   logger.info('trace agent activated');
 }
@@ -156,12 +159,16 @@ TraceAgent.prototype.addTransactionLabel = function(key, value) {
  * Call this from inside a namespace.run().
  */
 TraceAgent.prototype.createRootSpanData = function(name, traceId, parentId) {
+  var spanStart = new Date();
+  if (!this.policy.shouldTrace(spanStart.getTime())) {
+    return OpaqueSpan.nullSpan();
+  }
   traceId = traceId || (uuid.v4().split('-').join(''));
   parentId = parentId || 0;
   var trace = new Trace(0, traceId); // project number added later
-  var spanData = new SpanData(this, trace, name, parentId);
+  var spanData = new SpanData(this, trace, name, parentId, spanStart);
   cls.setRootContext(spanData);
-  return spanData;
+  return new OpaqueSpan(spanData);
 };
 
 /**
--- a/lib/trace-writer.js
+++ b/lib/trace-writer.js
@@ -86,17 +86,12 @@ TraceWriter.prototype.queueTrace_ = function(trace) {
     }
 
     trace.projectId = project;
+    that.buffer_.push(JSON.stringify(trace));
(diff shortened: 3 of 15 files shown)


matthewloring:
@ofrobots PTAL

ofrobots:
What are your thoughts on using the sampling decision earlier? Right now this change traces all the operations, just omits writing the collected traces to the buffer. An alternative implementation would skip tracing the root span.

matthewloring:
It will be a refactor of the agent and hooks that I'd prefer to do in a separate PR. Tracking in #95.

lib/trace-writer.js
```diff
@@ -44,6 +45,12 @@ function TraceWriter(logger, config) {
   /** @private {Array<string>} stringified traces to be published */
   this.buffer_ = [];
 
+  if (config.samplingRate === -1) {
```
ofrobots: This code belongs in tracing-policy.js. Trace-writer should not be in the business of looking at config.samplingRate.

lib/trace-writer.js
```diff
@@ -108,7 +113,7 @@ TraceWriter.prototype.queueTrace_ = function(trace) {
  * TraceWriter's config.
  */
 TraceWriter.prototype.scheduleFlush_ = function(project) {
-  this.logger_.info('Flushing: timeout');
+  this.logger_.info('Flushing: flushDelay seconds passed');
```
ofrobots: something like 'Performing periodic flush'

lib/tracing-policy.js
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+
+'use strict';
+
+function RateLimiterPolicy(samplesPerSecond) {
+  var that = this;
+  var timeout = 1000 / samplesPerSecond;
+  this.enabled = false;
+  timeout = timeout < 1 ? 1 : timeout;
```
ofrobots: We should make it more explicit in config.js that we definitely do not support samplingRates faster than 1000/sec.

lib/tracing-policy.js
```diff
@@ -0,0 +1,47 @@
...
+  var runner = function() {
```
ofrobots: Downside: This drives events through the event loop even if there is no real IO that the user's app is doing.

lib/trace-writer.js
matthewloring: Done

lib/trace-writer.js
matthewloring: Done.

lib/tracing-policy.js
matthewloring: Done.

lib/tracing-policy.js
matthewloring: In the process of rewriting this, we're moving away from that system.

matthewloring:
`test-trace-policy` is not yet able to check the higher sampling rates due to the slowdown of date creation. I'll look into mocking that out tomorrow with `timekeeper`.

matthewloring:
@ofrobots PTAL with the removal of opaque spans in mind.

lib/hooks/userspace/hook-express.js
```diff
@@ -98,12 +98,12 @@ function startRootSpanForRequest(req) {
...
   if (req.route && req.route.path) {
-    rootContext.span.setLabel(
+    rootContext.addLabel(
       'express/request.route.path', req.route.path);
   }
-  rootContext.span.setLabel(
+  rootContext.addLabel(
       TraceLabels.HTTP_RESPONSE_CODE_LABEL_KEY, res.statusCode);
-  rootContext.close();
+  rootContext.spanData_.close();
```
ofrobots: Ugh. Accessing a private property here.

lib/trace-agent.js
```diff
@@ -156,12 +159,16 @@ TraceAgent.prototype.addTransactionLabel = function(key, value) {
  * Call this from inside a namespace.run().
  */
 TraceAgent.prototype.createRootSpanData = function(name, traceId, parentId) {
+  var curr = new Date();
```
ofrobots: Use a more descriptive variable name.

lib/tracing-policy.js
```diff
@@ -0,0 +1,45 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+function RateLimiterPolicy(samplesPerSecond) {
+  var sampleDelay = 1000 / samplesPerSecond;
```
ofrobots: Guard against samplesPerSeconds being an invalid number here, specially 0.

lib/tracing-policy.js
```diff
@@ -0,0 +1,45 @@
...
+  this.traceWindow = sampleDelay < 1 ? 1 : sampleDelay;
+  this.lastTraceStart = Date.now() - sampleDelay;
+}
+
+RateLimiterPolicy.prototype.shouldTrace = function(dateMillis) {
+  var enable = dateMillis - this.lastTraceStart >= this.traceWindow;
```
ofrobots: Keep `this.nextTraceStart` instead of `this.lastTraceStart` in the state. Reduces the number of subtracts from the hot path.

lib/hooks/userspace/hook-express.js
matthewloring: Yea, there are private accesses all over due to the extra opaque span wrapping. I was hoping to clean this out when I delete `OpaueSpan` next but if it needs to go now I can do that as part of this PR.

ofrobots: I'm okay with it being done in the next PR.

lib/trace-agent.js

lib/tracing-policy.js
matthewloring: Done.

lib/tracing-policy.js
matthewloring: Done.

matthewloring:

lib/tracing-policy.js
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+function RateLimiterPolicy(samplesPerSecond) {
+  if (samplesPerSecond > 1000) {
+    samplesPerSecond = 1000;
+  } else if (samplesPerSecond === 0) {
```
ofrobots: If the user says 0 samples per second, I think it is incorrect for us to take silently samples. It would be better to throw.

ofrobots: Hmm.. not very sure about the throw. But I do think it would be not very nice to sample when they say 0.

matthewloring: Done.

ofrobots:
LGTM, w/ expectation that OpaqueSpan will be cleaned up in a follow up PR.

PR #99, opened 2015-10-08
OpaqueSpan: eradicated

matthewloring:
The OpaqueSpan class was originally created to wrap SpanData objects when exposed through the external facing custom span API. Once we began sampling, we needed to have nullSpan functionality inside the hooks. To achieve this, OpaqueSpans were used everywhere SpanData objects were needed. Since OpaqueSpan was needed everywhere, it no longer makes sense to wrap SpanData objects.


matthewloring:
@ofrobots PTAL

lib/trace-agent.js
```diff
@@ -134,6 +133,9 @@ TraceAgent.prototype.runInSpan = function(name, labels, fn) {
  */
 TraceAgent.prototype.setTransactionName = function(name) {
   var rootSpan = cls.getRootContext();
+  if (rootSpan === SpanData.nullSpan) {
```
ofrobots: It would be nice if we can remove the need to do this test. One way of doing this would be:
1. Make `SpanData#span` private.
2. Create a method on `SpanData` that transparently handles this.

matthewloring: We can push the name access down into `SpanData` but to remove this check from `addContextToHeaders`, we would need to push the context logic down as well. Since the accesses to `SpanData.span` are isolated to `TraceAgent` I think we're ok for now.

ofrobots: sgtm.

ofrobots:
This is great! LGTM.

PR #102, opened 2015-10-13
Patch the mongo connection pool

### LINKED ISSUE #96: Escaping mongo spans should not report an error in all cases
Some spans "escape" when mongo does periodic health checks and these should not be reported as errors. We should also further investigate mongo spans escaping in onconnection.

matthewloring:
ismaster checks occuring on behalf of incoming requests were escaping. By binding listeners on the connection pool event emitter to the current cls context, we can correctly attribute the work that they do.

Fixes #96

### DIFF
--- a/lib/hooks/index.js
+++ b/lib/hooks/index.js
@@ -17,7 +17,6 @@
 
 var Module = require('module');
 var shimmer = require('shimmer');
-var assert = require('assert');
 var path = require('path');
 
 //
@@ -29,67 +28,38 @@ var path = require('path');
 //   These are done on-demand via our Module._load wrapper
 // Install a tracer in the global scope.
 
-
+// Patches is a map from module load path to { filename1: { patchFunction: ...,
+// unpatchFunction: ..., module: ..., active: ... }, filename2: ... }
+// Each filename key represents a relative path which will be appended to the
+// absolute path to the root of the module. Leaving this relative path blank
+// will result in the module itself being patched.
+// Note: the order in which filenames are defined in the hooks determines the
+// order in which they are loaded.
 var toInstrument = Object.create(null, {
-  'express': {
-    enumerable: true,
-    value: {
-      file: './userspace/hook-express.js',
-      module: null,
-      hook: null,
-      version: null
-    }
-  },
-  'hapi': {
-    enumerable: true,
-    value: {
-      file: './userspace/hook-hapi.js',
-      module: null,
-      hook: null,
-      version: null
-    }
-  },
-  'http': {
-    enumerable: true,
-    value: {
-      file: './core/hook-http.js',
-      module: null,
-      hook: null,
-      version: null
-    }
-  },
-  'mongodb-core': {
-    enumerable: true,
-    value: {
-      file: './userspace/hook-mongodb-core.js',
-      module: null,
-      hook: null,
-      version: null
-    }
-  },
-  'redis': {
-    enumerable: true,
-    value: {
-      file: './userspace/hook-redis.js',
-      module: null,
-      hook: null,
-      version: null
-    }
-  },
-  'restify': {
-    enumerable: true,
-    value: {
-      file: './userspace/hook-restify.js',
-      module: null,
-      hook: null,
-      version: null
-    }
-  }
+  'express': { enumerable: true, value: { file: './userspace/hook-express.js',
+      patches: {} } },
+  'hapi': { enumerable: true, value: { file: './userspace/hook-hapi.js',
+      patches: {} } },
+  'http': { enumerable: true, value: { file: './core/hook-http.js',
+      patches: {} } },
+  'mongodb-core': { enumerable: true, value: { file: './userspace/hook-mongodb-core.js',
+      patches: {} } },
+  'redis': { enumerable: true, value: { file: './userspace/hook-redis.js',
+      patches: {} } },
+  'restify': { enumerable: true, value: { file: './userspace/hook-restify.js',
+      patches: {} } }
 });
 
 var logger;
 
-function findModuleVersion(request, parent, load) {
+/**
+ * Determines the path at which the requested module will be loaded given
+ * the provided parent module.
+ *
+ * @param {string} request The name of the module to be loaded.
+ * @param {object} parent The module into which the requested module will be loaded.
+ */
+function findModulePath(request, parent) {
   var mainScriptDir = path.dirname(Module._resolveFilename(request, parent));
   var resolvedModule = Module._resolveLookupPaths(request, parent);
   var paths = resolvedModule[1];
@@ -97,25 +67,51 @@ function findModuleVersion(request, parent, load) {
     if (mainScriptDir.indexOf(paths[i]) === 0) {
       // This is unsafe on windows if request contains a namespace
       // TODO: convert path separators in request
-      var pjson = load(path.join(paths[i], request, 'package.json'));
-      return pjson.version;
+      return path.join(paths[i], request);
     }
   }
   return null;
+}
+
+/**
+ * Determines the version of the module located at `modulePath`.
+ *
+ * @param {?string} modulePath The absolute path to the root directory of the
+ *    module being loaded. This may be null if we are loading an internal module
+ *    such as http.
+ */
+function findModuleVersion(modulePath, load) {
+  return modulePath ? load(path.join(modulePath, 'package.json')).version :
+    process.version;
 }
 
 function activate(agent) {
 
   logger = agent.logger;
 
-  function loadHookAndPatch(loadFn, instrumentation) {
-    assert(instrumentation.module);
-    assert(!instrumentation.hook);
+  function loadHookAndPatch(loadFn, instrumentation, moduleRoot, version) {
+    var modulePatch = instrumentation.patches[moduleRoot];
+    if (!modulePatch) {
+      // Load the hook. This file, i.e. index.js, becomes the parent module.
+      var moduleHook = loadFn(instrumentation.file, module, false);
+      modulePatch = moduleHook(version, agent);
+    }
+    Object.keys(modulePatch).forEach(function(file) {
+      if (!modulePatch[file].module) {
+        var loadPath = moduleRoot ? path.join(moduleRoot, file) : file;
+        modulePatch[file].module = loadFn(loadPath, module, false);
+      }
+      modulePatch[file].patch(modulePatch[file].module);
+      modulePatch[file].active = true;
+    });
+    instrumentation.patches[moduleRoot] = modulePatch;
+  }
 
-    // Load the hook. This file, i.e. index.js, becomes the parent module.
-    var moduleHook = loadFn(instrumentation.file, module, false);
-    instrumentation.hook = moduleHook;
-    moduleHook.patch(instrumentation.module, agent, instrumentation.version);
+  function moduleAlreadyPatched(instrumentation, moduleRoot) {
+    var modulePatch = instrumentation.patches[moduleRoot];
+    return modulePatch && Object.keys(modulePatch).every(function(curr) {
+      return modulePatch[curr].active;
+    }, true);
   }
 
   for (var moduleName in toInstrument) {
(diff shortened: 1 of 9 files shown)


matthewloring:
After looking here (https://github.com/christkv/mongodb-core/blob/master/lib/topologies/server.js#L364) it seems this is in fact being done on behalf of an incoming request. This needs to be reworked.

matthewloring:
@ofrobots PTAL.

lib/hooks/userspace/hook-express.js
```diff
@@ -106,28 +105,26 @@ function endRootSpanForRequest(rootContext, req, res) {
...
-      express = null;
-      agent = null;
-    }
+module.exports = function(version_, agent_) {
+  if (!semver.satisfies(version_, SUPPORTED_VERSIONS)) {
+    agent_.logger.info('Express: unsupported version ' + version_ + ' loaded');
+    return {};
   }
+  return {
+    'index.js': {
```
ofrobots: This makes me uncomfortable. We are dependent on the internal file layout for the express module. If the `index.js` filename changes in a minor version for whatever reason, we will stop working.

I can understand the need for depending, somewhat, on the internal layout for mongodb because they don't expose the Pool object, but we shouldn't over-constrain all the other modules we are patching.

matthewloring: I have changed these relative file paths to be the empty string in the case where we want to patch the module itself. I've added documentation describing the semantics of the empty key here.

matthewloring:

lib/hooks/userspace/hook-restify.js
```diff
@@ -103,24 +102,23 @@ function endRootSpanForRequest(rootContext, req, res) {
...
-      restify = null;
-      agent = null;
-    }
+module.exports = function(version_, agent_) {
+  if (!semver.satisfies(version_, SUPPORTED_VERSIONS)) {
+    agent_.logger.info('Restify: unsupported version ' + version_ + ' loaded');
+    return {};
   }
+  return {
+    // An empty relative path here results in the root module being loaded.
```
ofrobots: s/results in/matches

matthewloring: Fixed throughout.

lib/hooks/index.js
```diff
@@ -97,25 +58,45 @@ function findModuleVersion(request, parent, load) {
...
-    instrumentation.hook = moduleHook;
-    moduleHook.patch(instrumentation.module, agent, instrumentation.version);
+    if (!versionHooks) {
+      var moduleHook = loadFn(instrumentation.file, module, false);
+      versionHooks = moduleHook(version, agent);
+    }
+    Object.keys(versionHooks).forEach(function(file) {
+      if (!versionHooks[file].module) {
+        var loadPath = moduleRoot ? path.join(moduleRoot, file) : file;
+        var loaded = loadFn(loadPath, module, false);
```
ofrobots: Remove this `var loaded` and assign directly to `.module`

lib/hooks/index.js
```diff
@@ -97,25 +58,45 @@ function findModuleVersion(request, parent, load) {
...
+    var versionHooks = instrumentation.versions[version];
     // Load the hook. This file, i.e. index.js, becomes the parent module.
-    var moduleHook = loadFn(instrumentation.file, module, false);
-    instrumentation.hook = moduleHook;
-    moduleHook.patch(instrumentation.module, agent, instrumentation.version);
+    if (!versionHooks) {
+      var moduleHook = loadFn(instrumentation.file, module, false);
+      versionHooks = moduleHook(version, agent);
+    }
+    Object.keys(versionHooks).forEach(function(file) {
```
ofrobots: I have a disconnect on the naming. The keys of the `versionHooks` are `file`s? Why is it called versionHooks then?

lib/hooks/index.js
```diff
@@ -97,25 +58,45 @@ function findModuleVersion(request, parent, load) {
...
+
 function activate(agent) {
 
   logger = agent.logger;
 
-  function loadHookAndPatch(loadFn, instrumentation) {
-    assert(instrumentation.module);
-    assert(!instrumentation.hook);
-
+  function loadHookAndPatch(loadFn, instrumentation, moduleRoot, version) {
```
ofrobots: Can a given module at a given version be loaded at multiple module root directories? Does this code deal with that case?

E.g. module a depends on modules b and c. Module b and c both depend on express@4.0.1. I think we will have two copies of express. Right?

lib/hooks/index.js
matthewloring: Done.

lib/hooks/index.js
matthewloring: It is the set of hooks that should be applied for the given version where a hook is a file -> patch/unpatch map. Is there a name that would clarify this structure better?

lib/hooks/index.js
matthewloring: My understanding of the module loader is that the second time express@4.0.1 is loaded, the cached version will be used and this cached version will correspond to the module that was originally loaded at a different path. I will put together a small example to confirm this is actually the case.

matthewloring: This caching does not appear to happen the way I expected. I would be inclined to open this as a bug and fix it independently as this functionality wasn't present before this change and it doesn't represent a regression introduced by this change. I expect this to be a relatively rare case since web servers and databases are rarely buried in dependencies the way utilities are.

lib/hooks/index.js
matthewloring: I've changed this to `modulePatch` in the most recent version.

lib/hooks/index.js
matthewloring: Actually, I thought of a minimally intrusive way of supporting multiple copies of the same version. We now key on load path as opposed to version number.

matthewloring:

lib/hooks/index.js
```diff
@@ -97,25 +58,44 @@ function findModuleVersion(request, parent, load) {
...
       // TODO: convert path separators in request
-      var pjson = load(path.join(paths[i], request, 'package.json'));
-      return pjson.version;
+      return path.join(paths[i], request);
     }
   }
   return null;
 }
 
+function findModuleVersion(modulePath, load) {
```
ofrobots: Add document on function usage. Empty module path implies native module?

matthewloring: This modulePath cannot be empty. This is the absolute path to the module. I have added documentation to this effect.

matthewloring:

lib/hooks/index.js
ofrobots:
```
+function findModuleVersion(modulePath, load) {
+  return modulePath ? load(path.join(modulePath, 'package.json')).version :
+    process.version;
+}
```
If the modulePath cannot be empty, why is there a test in there?

matthewloring: Sorry, you are correct. It may be null in the case where we are patching an internal module like http. I've updated the comment to reflect this.

lib/hooks/index.js
```diff
@@ -97,25 +67,51 @@ function findModuleVersion(request, parent, load) {
...
 
+/**
+ * Determines the version of the module located at `modulePath`.
+ *
+ * @param {?string} modulePath The absolute path to the root directory of the
+ *    module being loaded. This may be null if we are loading an internal module
+ *    such as http.
+ */
+function findModuleVersion(modulePath, load) {
+  return modulePath ? load(path.join(modulePath, 'package.json')).version :
```
ofrobots: This predates your changes. What happens when we try to load a module from a module directory without a `package.json`?

lib/hooks/index.js
```diff
@@ -136,33 +132,33 @@ function activate(agent) {
     // time that we need to go and patch pro-actively.
```
ofrobots: The indentation is wrong here. If you're in this code, please cleanup the indentation too.

lib/hooks/index.js
```diff
@@ -136,33 +132,33 @@ function activate(agent) {
...
-              originalModuleLoad);
-          logger.info('Patching ' + request + ' at version ' +
-              instrumentation.version);
-          instrumentation.module = loaded;
-          loadHookAndPatch(originalModuleLoad, instrumentation);
+      var instrumentation = toInstrument[request];
+
+      if (instrumentation &&
+          agent.config().excludedHooks.indexOf(request) === -1) {
+        var moduleRoot = findModulePath(request, arguments[1]);
```
ofrobots: Let's start using named arguments. The module API is locked, so it ought to be safe.

lib/hooks/index.js
```diff
@@ -136,33 +132,33 @@ function activate(agent) {
...
+        if (moduleAlreadyPatched(instrumentation, moduleRoot)) {
+          return originalModuleLoad.apply(this, arguments);
         }
+        var moduleVersion = findModuleVersion(moduleRoot, originalModuleLoad);
+        logger.info('Patching ' + request + ' at version ' + moduleVersion);
+        loadHookAndPatch(originalModuleLoad, instrumentation, moduleRoot,
+          moduleVersion);
       }
 
       // no hooks installed. return the original module as-is.
```
ofrobots: This comment is no longer accurate. We reach here from line 157 where we have installed hooks.

lib/hooks/index.js
```diff
@@ -97,25 +67,51 @@ function findModuleVersion(request, parent, load) {
...
+    instrumentation.patches[moduleRoot] = modulePatch;
+  }
 
-    // Load the hook. This file, i.e. index.js, becomes the parent module.
-    var moduleHook = loadFn(instrumentation.file, module, false);
-    instrumentation.hook = moduleHook;
-    moduleHook.patch(instrumentation.module, agent, instrumentation.version);
+  function moduleAlreadyPatched(instrumentation, moduleRoot) {
+    var modulePatch = instrumentation.patches[moduleRoot];
+    return modulePatch && Object.keys(modulePatch).every(function(curr) {
```
ofrobots: Is it ever possible for a subset of the entries to be active and others to be not active? If no, does it make sense to have .active on everything? Should it be moved up to modulePatch?

lib/hooks/index.js
```diff
@@ -97,25 +67,51 @@ function findModuleVersion(request, parent, load) {
...
+}
+
 function activate(agent) {
 
   logger = agent.logger;
 
-  function loadHookAndPatch(loadFn, instrumentation) {
-    assert(instrumentation.module);
-    assert(!instrumentation.hook);
+  function loadHookAndPatch(loadFn, instrumentation, moduleRoot, version) {
```
ofrobots: 1. Lets move this function and moduleAlreadyPatched inside the shimmer.wrap below. Get rid of the loadFn argument and reuse the originalModuleLoad from the closure.
2. It might be better to rename this to `loadAndPatch`.

lib/hooks/index.js
matthewloring: My understanding is that you cannot publish a module on npm w/o a package.json in your root directory and earlier today we agreed that we already do not patch non-npm installed modules at all.

lib/hooks/index.js
matthewloring: I think it is correct in this case.

lib/hooks/index.js
matthewloring: Fixed.

lib/hooks/index.js
matthewloring: Fixed.

lib/hooks/index.js
matthewloring: This is not currently possible but there isn't a nice way to store this info a level up without adding anther layer of wrapping around things. Additionally, there isn't anything fundamental to say that they couldn't be patched individually in the future.

lib/hooks/index.js
matthewloring: Done.

matthewloring:

lib/hooks/index.js
ofrobots: My question was not about published modules. Convince me that the access to `.version` will never crash.

lib/hooks/index.js
ofrobots: Ack. Not sure why I got confused about the indentation.

lib/hooks/index.js
matthewloring: The contract of `findModulePath` is that it will either return a valid path to a module or null based on the use of `Module#_resolveFilename` and `Module#_resolveLookupPaths`. Since the argument passed into `findModuleVersion` is the return value of a call to `findModulePath`, the load should succeed. We can add additional checks or fold this functionality into the `findModulePath` logic but in the current form I don't think this can fail.

ofrobots: Sorry, I'll be less obtuse this time. 

> earlier today we agreed that we already do not patch non-npm installed modules at all

That's not what we agreed. My claim is that you can't load a module specified by name (as opposed to a relative or absolute path) unless it exists inside some `node_modules/${package_name}`. That `node_modules` needs to be in the module search path for the parent module.

I think it is possible for a user to self-install a directory inside `node_modules`. If that directory doesn't contain a `package.json`, the code above would crash.

It should never crash.

matthewloring: Ah sorry, that makes sense! I've added a check to make sure the file exists before trying to load it.

ofrobots: Alternatively, you can make sure that `Module._load` behaves predictably in the case the file doesn't exist (e.g. doesn't throw or crash), and check the return value. I would expect `Module._load` already needs to do a file existence check; it would good not to repeat it ourselves if at all possible.

matthewloring: I just gave this a try and it threw an error, we could catch the error if that isn't too much overhead and inspect it to see if it has the message "Cannot find module ..."

test/hooks/test-hooks-interop-mongo-express.js
```diff
@@ -0,0 +1,58 @@
...
+};
+
+var server;
+
+describe('mongodb + express', function() {
+  it('should not lose context on startup', function(done) {
+    var app = express();
+    app.get('/', function (req, res) {
+      mongoose.connect('mongodb://localhost:27017/testdb', function(err) {
+        assert(!err, 'Skipping: no mongo server found at localhost:27017.');
```
ofrobots: The assertion message here seems incorrect.

matthewloring: I agree, I've fixed this for mongo and redis.

lib/hooks/userspace/hook-mongodb-core.js
```diff
@@ -66,32 +65,46 @@ function wrapCallback(span, done) {
...
+
+module.exports = function(version_, agent_) {
+  if (!semver.satisfies(version_, SUPPORTED_VERSIONS)) {
+    agent_.logger.info('Mongo: unsupported version ' + version_ + ' loaded');
+    return {};
   }
+  return {
+    'lib/connection/pool.js': {
+      patch: function(pool) {
+        shimmer.wrap(pool.prototype, 'once', onceWrap);
```
ofrobots: Is there a reason we don't need to wrap `.on`, or alternatively use `bindEmitter` on pool instances?

matthewloring: I believe `once` is just abusing the naming on an event emitter like object. When I try binding it with cls, I get `AssertionError: can only bind real EEs`.

matthewloring: We could wrap `on` as long as it is guaranteed to be called by `once` since `once` is the function invoked directly by the client code.

ofrobots: I am confused with the response as the Pool is actually an EventEmitter subclass as per: https://github.com/christkv/mongodb-core/blob/6c9c21150163e4e7ac97809b48a04430adf36873/lib/connection/pool.js#L68

Were you perchance trying to bindEmitter on pool.prototype?

matthewloring: I was trying to bind the Pool object itself. Maybe bind can only be called on instances of type Pool? that would involve patching the constructor.

ofrobots:
Retroactive LGTM.

PR #105, opened 2015-10-19
Report GCE instance information as labels on spans

### LINKED ISSUE #100: Add GCE VM information as a label to the traces
For people running on Compute Engine, it would be useful to be able to do filtering of the traces by VM. Right now the trace backend has no knowledge of which VM a particular trace was captured on. We should add this as a label on the traces. Once that is available, the UI can offer a way to filter by it. This issue only tracks the former.

matthewloring:
This depends on the changes in
GoogleCloudPlatform/cloud-diagnostics-common-nodejs#10

Fixes #100

### DIFF
--- a/index.js
+++ b/index.js
@@ -102,6 +102,22 @@ var publicAgent = {
       return this;
     }
 
+    common.utils.getHostname(function(err, hostname) {
+      if (err) {
+        logger.info('Unable to retrieve GCE hostname.', err);
+      } else {
+        config.hostname = hostname;
+      }
+    });
+
+    common.utils.getInstanceId(function(err, instanceId) {
+      if (err) {
+        logger.info('Unable to retrieve GCE instance id.', err);
+      } else {
+        config.instanceId = instanceId;
+      }
+    });
+
     if (typeof config.projectId === 'undefined') {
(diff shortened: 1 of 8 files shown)


index.js
```diff
@@ -102,6 +102,22 @@ var publicAgent = {
...
+    });
+
+    common.utils.getInstanceId(function(err, instanceId) {
+      if (err) {
+        logger.info('Unable to retrieve GCE instance id.', err);
+      } else {
+        config.instanceId = instanceId;
+      }
+    });
+
```
ofrobots: 1. This is not an actionable item, but rather an observation. Any traces captured before these async functions complete will not have the instance and hostname fields attached to them. I do think it is worth capturing traces even if these operations haven't completed.

2. Upgrade the logger message to `.warn` instead of `.info`. If we are running on GCP and can't get this data, this is a problem; albeit not a critical one.

3. When not running on GCP, neither a `warn` or `info` is warranted.

matthewloring: Fixed 2 and 3. Change is reflected in #109

matthewloring:
@ofrobots Ready for another  look.

ofrobots:
Retroactive LGTM.

PR #106, opened 2015-10-19
Possible to omit all traces at specified urls

### LINKED ISSUE #103: Selectively omit traces for user specified urls/paths

matthewloring:
Introduced a new tracing policy that ignores all traces occuring at
designated paths before sampling occurs to prevent ignored urls from
being sampled.

Fixes #103.

### DIFF
--- a/config.js
+++ b/config.js
@@ -34,6 +34,14 @@ module.exports = {
   // See `bufferSize`.
   flushDelaySeconds: 30,
 
+  // If paths are present in this array, then these paths will be ignored before
+  // `samplingRate` based decisions are made. Paths must include a leading
+  // forward slash and be of the form:
+  //   /componentOne/componentTwo/...
+  // Paths can additionally be classified by regex in which case any path matching
+  // any provided regex will be ignored.
+  ignoreUrls: [],
+
   // An upper bound on the number of traces to gather each second. If set to
   // -1, all traces will be reported. Sampling rates greater than 1000 are not
   // supported and will result in at most 1000 samples per second.
--- a/lib/tracing-policy.js
+++ b/lib/tracing-policy.js
@@ -35,16 +35,47 @@ RateLimiterPolicy.prototype.shouldTrace = function(dateMillis) {
   return true;
 };
 
+function FilterPolicy(basePolicy, filterUrls) {
+  this.basePolicy = basePolicy;
+  this.filterUrls = filterUrls;
+}
+
+FilterPolicy.prototype.matches = function(url) {
+  for (var i = 0; i < this.filterUrls.length; i++) {
+    var candidate = this.filterUrls[i];
+    if (typeof candidate === 'string') {
+      if (candidate === url) {
+        return true;
+      }
+    } else {
+      if (url.match(candidate)) {
+        return true;
+      }
+    }
+  }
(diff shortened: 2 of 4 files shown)


matthewloring:
@ofrobots PTAL.

matthewloring:
This patch assumes that we use the request url as the name for each root span. This is currently the case for all web frameworks we patch. We would have to refactor the hooks api to expose this information independently of the name. I'm happy to do this if preferable.

config.js
```diff
@@ -34,6 +34,12 @@ module.exports = {
   // See `bufferSize`.
   flushDelaySeconds: 30,
 
+  // If paths are present in this array, then these paths will be ignored before
+  // `samplingRate` based decisions are made. Paths must include a leading
+  // forward slash and be of the form:
+  //   /componentOne/componentTwo/...
+  ingoreUrls: [],
```
ofrobots: typo

matthewloring: Dang, thanks for catching this. Fixed.

ofrobots: How about generalizing this to be an array of regular expressions?

matthewloring: Done. We now accept both regex and direct string matches. I don't want people to have to escape '/' as direct matches will likely be the common case and paths have lots of them.

config.js
```diff
@@ -34,6 +34,14 @@ module.exports = {
   // See `bufferSize`.
   flushDelaySeconds: 30,
 
+  // If paths are present in this array, then these paths will be ignored before
+  // `samplingRate` based decisions are made. Paths must include a leading
+  // forward slash and be of the form:
```
ofrobots: Looking at `FilterPolicy#matches` implementation, it doesn't seem like we actually depend on the leading slash as a requirement. Am I reading the code right?

lib/tracing-policy.js
```diff
@@ -35,16 +35,47 @@ RateLimiterPolicy.prototype.shouldTrace = function(dateMillis) {
   return true;
 };
 
+function FilterPolicy(basePolicy, filterUrls) {
+  this.basePolicy = basePolicy;
+  this.filterUrls = filterUrls;
+}
+
+FilterPolicy.prototype.matches = function(url) {
+  for (var i = 0; i < this.filterUrls.length; i++) {
```
ofrobots: Use `Array.prototype.some`.

config.js
matthewloring: If you are not providing a regex, then your path must include a leading slash to satisfy the the condition on lib/tracing-policy.js#47. If you are using a regex there is no need to provide the leading slash. This is required so that if they specify an ignoreUrl of 'test/get' then we only want to ignore 'www.example.com/test/get' and not 'www.example.com/ns/test/get'. We could autofill the slash if not present but this would be ambiguous as it is not clear which of the urls the programmer intended. If they want both they should use a regex.

lib/tracing-policy.js
matthewloring: Tracking this fix in #109

matthewloring:

config.js
ofrobots: sgtm.

ofrobots:
Retroactive LGTM.

PR #107, opened 2015-10-19
Report query and result information for mongodb

### LINKED ISSUE #97: Save more operation data on mongodb operations
* Query/Update parameter
* Result Size? Results?

matthewloring:
This information is only reported if the `enhancedReporting` flag is
set.

Fixes #97.

### DIFF
--- a/config.js
+++ b/config.js
@@ -22,6 +22,14 @@ module.exports = {
 
   enabled: true,
 
+  // If true, information about query parameters and results will be
+  // attached to spans representating database operations.
+  enhancedDatabaseReporting: false,
+
+  // The maximum result size in characters to report on database spans if
+  // `enhancedDatabaseReporting` is enabled.
+  databaseResultReportingSize: 127,
+
   // Valid entries are:
   // 'express', 'hapi', 'http', 'mongodb-core', restify'
   excludedHooks: [],
--- a/lib/hooks/userspace/hook-mongodb-core.js
+++ b/lib/hooks/userspace/hook-mongodb-core.js
@@ -16,6 +16,7 @@
 
 'use strict';
 
+var traceUtil = require('../../util.js');
 var cls = require('../../cls.js');
 var shimmer = require('shimmer');
 var semver = require('semver');
@@ -34,7 +35,11 @@ function nextWrap(next) {
     } else if (root === SpanData.nullSpan) {
       return next.apply(this, arguments);
     }
-    var span = agent.startSpan('mongo-cursor', { db: this.ns });
+    var labels = { db: this.ns };
+    if (agent.config_.enhancedDatabaseReporting) {
+      labels.cmd = JSON.stringify(this.cmd);
+    }
+    var span = agent.startSpan('mongo-cursor', labels);
     return next.call(this, wrapCallback(span, cb));
   };
 }
@@ -50,7 +55,11 @@ function wrapWithLabel(label) {
       } else if (root === SpanData.nullSpan) {
         return original.apply(this, arguments);
       }
-      var span = agent.startSpan(label, { db: ns });
+      var labels = { db: ns };
+      if (agent.config_.enhancedDatabaseReporting) {
+        labels.operations = JSON.stringify(ops);
+      }
+      var span = agent.startSpan(label, labels);
       if (typeof options === 'function') {
         return original.call(this, ns, ops,
           wrapCallback(span, options));
@@ -62,9 +71,29 @@ function wrapWithLabel(label) {
   };
 }
 
+/**
+ * Wraps the provided callback so that the provided span will
+ * be closed immediately after the callback is invoked.
+ *
+ * @param {Span} span The span to be closed.
+ * @param {Function} done The callback to be wrapped.
+ * @return {Function} The wrapped function.
+ */
 function wrapCallback(span, done) {
   var fn = function(err, res) {
-    agent.endSpan(span);
+    var labels = {};
+    if (agent.config_.enhancedDatabaseReporting) {
+      if (err) {
+        // Errors may contain sensitive query parameters.
+        labels.mongoError = err;
+      }
+      if (res) {
+        var result = res.result ? res.result : res;
+        labels.results = traceUtil.stringifyPrefix(result,
+          agent.config_.databaseResultReportingSize);
+      }
+    }
+    agent.endSpan(span, labels);
     if (done) {
       done(err, res);
     }
--- a/lib/util.js
+++ b/lib/util.js
@@ -1,0 +1,68 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+/**
+ * Produces an object summarization of limited size. This summarization
+ * does not adhere to the JSON spec so it cannot be reparsed even if
+ * the entire object fits inside the desired size.
+ *
+ * @param {Object} o The object to be summarized.
+ * @param {number} n Max length of the summary.
+ */
+function stringifyPrefix(o, n) {
+  var buf = new Buffer(n);
+  var pos = 0;
+  var worklist = [];
+  function pushObject(o) {
+    var keys = Object.keys(o);
+    for (var i = Math.min(keys.length - 1, Math.floor(n/4)); i >= 0; i--) {
+      worklist.push((i === keys.length - 1) ? '}' : ',');
+      worklist.push(o[keys[i]]);
+      worklist.push(':');
+      worklist.push(keys[i]);
+    }
+    worklist.push('{');
+  }
+  worklist.push(o);
+  while (worklist.length > 0) {
+    var elem = worklist.pop();
+    if (elem && typeof elem === 'object') {
+      pushObject(elem);
+    } else {
+      var val;
+      if (typeof elem === 'function') {
+        val = '[Function]';
+      } else if (typeof elem === 'string') {
+        val = elem;
+      } else {
+        // Undefined, Null, Boolean, Number, Symbol
+        val = String(elem);
+      }
+      pos += buf.write(val, pos);
+      if (buf.length === pos) {
+        buf.write('...', pos - 3);
+        break;
+      }
+    }
+  }
+  return buf.toString('utf8', 0, pos);
+}
+
+module.exports = {
+  stringifyPrefix: stringifyPrefix
+};
--- a/test/hooks/common.js
+++ b/test/hooks/common.js
@@ -20,7 +20,8 @@ if (!process.env.GCLOUD_PROJECT_NUM) {
   process.exit(1);
 }
 
-var agent = require('../..').start({ samplingRate: -1 }).private_();
+var config = { enhancedReporting: true, samplingRate: -1 };
+var agent = require('../..').start(config).private_();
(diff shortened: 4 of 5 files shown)


matthewloring:
@ofrobots PTAL. We can incorporate some way of not reporting large results such as "if more than n results, display count instead of full result" but this doesn't account for large single documents. I'm not sure there is a nice way to handle the latter without asking them to filter by documents fields which we would have to take as a config option. I wanted to make sure there was a real concern about response size before implementing a heuristic like the one above.

config.js
```diff
@@ -22,6 +22,10 @@ module.exports = {
 
   enabled: true,
 
+  // If true, information about query parameters and results will be
+  // attached to spans representating database operations.
+  enhancedReporting: false,
```
ofrobots: Could we use a more descriptive name? Enhanced could mean anything. It also closes the door on us adding a really enhanced tracing mode later on :smile:.

ofrobots:
I agree with both concerns; we shouldn't trace too many documents, and neither should we trace large single documents. Both should be controlled using a size limit of some sort. i.e. never trace more than, say, 256 bytes on a single capture. So you return the document count (always) and the first `config_.documentResultsLimitBytes` bytes of the resulting documents(s).

Ideally such limiting would be done before any stringification is done.

matthewloring:
Is there a natural way to gather the "first" n bytes of a javascript object? I can add logging of counts but I'm not sure how document limiting could be done. I'm also not sure how easily a partial javascript object could be stringified.

config.js

ofrobots:
> Is there a natural way to gather the "first" n bytes of a javascript object? I can add logging of counts but I'm not sure how document limiting could be done. I'm also not sure how easily a partial javascript object could be stringified.

Can you do some research?

matthewloring:
On it!

matthewloring:
I didn't find anything too useful for truncating an object to a particular size. The problem is that there are lots of possible heuristics and it takes information about the structure of the objects being stored in mongo to choose the best one. It is unlikely that there is a single heuristic that provides a meaningful result most of the time so I would be inclined to leave this for now (until large single documents become problematic). I've limited the result to at most 3 objects for now which should handle the much more common case of querying many documents.

matthewloring:

ofrobots:
> until large single documents become problematic

This should be fixed from the start. We know large documents exist. I think you will have to do much of the work JSON.stringify does, manually.

matthewloring:
Hmm, ok. I could see objects growing large for a few reasons:

1. Many top level keys
1. Deeply nested values
1. Very large non-json values
1. Very large keys

I'm going to hope that 4. won't be a problem for us. I could apply a heuristic that addresses each of the others if you foresee one being substantially more important. Otherwise we could analyze the object and apply some combination to address each of these concerns as they are discovered. 

I would prefer if the UI allowed some nice display of large JSON (collapsing nested objects, etc) since it doesn't feel good to omit arbitrary data without having an understanding of what is important for the user. Unfortunately, I'm not sure if/when that would be a priority for the UI team.

Do you have a preference for addressing the cases above or do you want a fully general solution that addresses all of them at the cost of complexity.

ofrobots:
I'm not very sure about 3.. what would be a large non-json value. 

If the results documents are to be included at all in the captured traces, we need to address these issues. I would be okay with this PR only capturing the result count as opposed to the documents, and you can address the documents later on as a follow-on PR when you have time to implement truncation heuristics above.

matthewloring:
3 would be the result of storing arbitrarily sized strings or binary data (this happens frequently with compression to my understanding). I can implement something like "show the first 20 bytes of each key and the first 30 bytes of each value for the first 5 key value pairs". My only concern is that the more dimensions we truncate on with arbitrary cutoffs, the more reasonable documents we truncate. Alternatively, would could gather relative sizes of objects stored at each key value location and eliminate values in reverse size order until we are under a total byte limit. I am happy to include a truncation mechanism in this PR, I'm just trying to get a gauge of how much time complexity you want to invest in preserving readability as much as possible.

matthewloring:
I put together an example heuristic as food for discussion:

```javascript
function getReplacer(numObjects) {
  return function replacer(k, v) {
    var isArray = Array.isArray(v);
    if (isArray || typeof v === 'object') {
      if (numObjects === 0) {
        return '...';
      }
      numObjects--
      if (isArray) {
        if (v.length > 5) {
          v = v.slice(0, 5);
          v.push('...');
        }
      }
      return v;
    } else if (typeof v === 'string') {
      if (v.length > 15) {
        v = v.substr(0, 15) + '...';
      }
      return v;
    }
    return v;
  };
}
```

If we define the object:

```javascript
var o = {"a": [1, 2, 3, 4, 6, 7, 9, 10], "b": "mysuperlongcompressedbyinarydataorhashorsaltorsomething", "c": {"a": {"b": {"c": {"d": {"e": "hi"}}}}}};
```

Then `JSON.parse(JSON.stringify(o, getReplacer(3)))` produces:

```javascript
{"a": [1, 2, 3, 4, 6, "..."], "b": "mysuperlongcomp...", "c": {"a": "..."}}
```

A solution like this would address 2 and 3 above but does not currently limit the number of object keys displayed. If we want to pursue something sufficiently more complicated, it may make sense to pull the logic out into a separate module.

ofrobots:
For food for thought also consider what the debug agent does: https://github.com/ofrobots/cloud-debug-nodejs/blob/master/lib/state.js#L120

Basically you do a breadth-first-walk of the object graph and abandon the queue once you have reached a particular size heuristic.

matthewloring:
That would be a good alternative. My hope was to display at least something for each key in the top level object if possible. The worry would be that the very first field of the document takes up the entirety of the size quota preventing other keys from being displayed.

ofrobots:
Even the first field should be truncated if it is too big.

matthewloring:
Definitely, I was just hoping to come up with something that would truncate a little more aggressively so that other parts of the documents make it to the UI as well. I'm happy to proceed with the dfs approach if that is sufficient!

ofrobots:
I think a combination of the two approaches might be nice as long as there are configuration knobs that can be used to adjust (sounds like than two knobs might be sufficient). BTW, you want bfs rather than dfs.

matthewloring:
@ofrobots Thoughts on an approach like this? It requires generators which means transpilation for 0.12. The primary reason to use generators over a more traditional approach is to avoid unfolding all of an objects key, value pairs before we know that they will be required.

```javascript
function stringifyPrefix(n, o) {
  var buf = new Buffer(n);
  var pos = 0;
  function* yielder(o) {
    if (typeof o === 'object') {
      yield '{';
      var keys = Object.keys(o);
      for (var i = 0; i < keys.length; i++) {
        yield* [keys[i], ':'];
        yield* yielder(o[keys[i]]);
        yield (i < keys.length - 1) ? ',' : '}';
      }
    } else {
      yield String(o);
    }
  }
  for (var val of yielder(o)) {
    pos += buf.write(val, pos);
    if (buf.length === pos) {
      buf.write('...', pos - 3);
      break;
    }
  }
  return buf.toString('utf8', 0, pos);
}
```

Alternatively, we can do a slightly more greedy but non generator dependent version:

```javascript
function stringifyPrefix(n, o) {
  var buf = new Buffer(n);
  var pos = 0;
  var worklist = [];
  function pushObject(o) {
    var keys = Object.keys(o);
    for (var i = keys.length - 1; i >= 0; i--) {
      worklist.unshift((i === keys.length - 1) ? '}' : ',');
      worklist.unshift(o[keys[i]]);
      worklist.unshift(':');
      worklist.unshift(keys[i]);
    }
    worklist.unshift('{');
  }
  worklist.push(o);
  while (worklist.length > 0) {
    var elem = worklist.shift();
    if (typeof elem === 'object') {
      pushObject(elem);
    } else {
      pos += buf.write(String(elem), pos);
      if (buf.length === pos) {
        buf.write('...', pos - 3);
        break;
      }
    }
  }
  return buf.toString('utf8', 0, pos);
}
```

ofrobots:
The problem with using ES6 + transpilation is that I will no longer be qualified to do code reviews as I still have to pick up ES6 in detail and would want to know the performance implications of whatever transpiler we use (i.e. know what it does). This would be an ugly hack, but, could you statically transpile that generator code above and check it in? I don't expect we'll be changing the source logic a lot going forward?

matthewloring:
I think the statically compiled code will be too unpleasant to have in the codebase. Lets use the lower version for now and we can revisit generators if there are still performance issues.

ofrobots:
Added comments. LMK when this is ready for review again.

matthewloring:

matthewloring:
@ofrobots Added comment on error reporting.

lib/hooks/userspace/hook-mongodb-core.js
```diff
@@ -43,21 +48,45 @@ function wrapWithLabel(label) {
...
-      var span = agent.startSpan(label, { db: ns });
+      var labels = { db: ns };
+      if (agent.config_.enhancedDatabaseReporting) {
+        labels.operations = JSON.stringify(ops);
+      }
+      var span = agent.startSpan(label, labels);
       if (typeof options === 'function') {
         return original.call(this, ns, ops,
-          wrapCallback(span, options));
+          wrapCallback(span, options, 127));
```
ofrobots: Mumble about hard-coded constant. I would prefer this to be exposed in the config so that the user can customize this if necessary.

lib/util.js
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Produces an object summarization of limited size. This summarization
+ * does not adhere to the JSON spec so it cannot be reparsed even if
+ * the entire object fits inside the desired size.
+ *
+ * @param {number} n Max length of the summary.
+ * @param {Object} o The object to be summarized.
+ */
+function stringifyPrefix(n, o) {
```
ofrobots: Minor nit: it would preferable to have the order of the arguments reversed – that would sound more 'standard' to me, but I won't necessarily be able to dig a up 'standard' documenting that.

lib/hooks/userspace/hook-mongodb-core.js
matthewloring: Done.

lib/util.js
matthewloring: Fixed.

matthewloring:

lib/util.js
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * @param {Object} o The object to be summarized.
+ * @param {number} n Max length of the summary.
+ */
+function stringifyPrefix(o, n) {
+  var buf = new Buffer(n);
+  var pos = 0;
+  var worklist = [];
+  function pushObject(o) {
+    var keys = Object.keys(o);
```
ofrobots: will throw if o is null.

matthewloring: Fixed.

matthewloring:
@ofrobots There are lots of possible optimizations that involve tweaking array allocations but I think what we have now is more performant than fully stringifying and addresses the correctness concerns.

matthewloring:
@ofrobots Removed environmental variable for controlling enhanced logging in anticipation of #115

lib/util.js
```diff
@@ -0,0 +1,68 @@
...
+    for (var i = keys.length - 1; i >= 0; i--) {
```
ofrobots: You may want to consider something similar to https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/blob/master/lib/state.js#L351 to prevent pathologies with really large numbers of keys. This might be as simple as initializing i as `var i = min(keys.length - 1, config.maxProperties)`. WDYT?

matthewloring: Hmm, protecting against many keys is good but I hesitate to expose this as a user facing knob as it feels too much like an algorithmic detail. They've already given us a max buffer size and shouldn't have to reason about the details of key processing. We could limit it at `n/4` since we expect to put at least 4 characters in the buffer per key in the object. It is a rougher approximation but shields the user. On another note, I like the tonicdev approach to summarization and I was thinking about how to implement it efficiently but it only really works if you're able to drill down into the objects replaced with `...`.

ofrobots: n/4 sgtm.

ofrobots:
LGTM w/ following comments
1. once n/4 heuristics on the keys is implemented.
2. Please add a few unit tests for stringifyPrefix.

Do not this land yet however.

matthewloring:
All fixed, I will land this when possible.

PR #108, opened 2015-10-21
Run redis v2 against unit tests

### LINKED ISSUE #92: Update to monkeypatch Redis 2

matthewloring:
This also tests the patching of multiple versions of a module
simultaneously. Also, removed a test that was not testing
agent functionality (just a property of redis).

Fixes #92.

### DIFF
--- a/lib/hooks/userspace/hook-redis.js
+++ b/lib/hooks/userspace/hook-redis.js
@@ -21,7 +21,7 @@ var shimmer = require('shimmer');
 var semver = require('semver');
 var agent;
 
-var SUPPORTED_VERSIONS = '0.12.x';
+var SUPPORTED_VERSIONS = '<=2.2.x';
 
 function createClientWrap(createClient) {
   return function createClientTrace() {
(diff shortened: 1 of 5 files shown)


matthewloring:
@ofrobots PTAL.

lib/hooks/userspace/hook-redis.js
```diff
@@ -21,7 +21,7 @@ var shimmer = require('shimmer');
 var semver = require('semver');
 var agent;
 
-var SUPPORTED_VERSIONS = '0.12.x';
+var SUPPORTED_VERSIONS = '<=2.2.5';
```
ofrobots: What's the rationale for not supporting bug fix releases on the 2.2.x line?

ofrobots:
LGTM with nit about the supported versions. Either relax the requirement, or add a comment explaining why the requirements are as written.

lib/hooks/userspace/hook-redis.js
matthewloring: Fixed.

PR #109, opened 2015-10-26
Cleanup for submitted PRs

matthewloring:
Some PRs were submitted without LGTM while reviewers were away last
week. This PR fixes retroactive review comments for #102, #105, and #106.

### DIFF
--- a/index.js
+++ b/index.js
@@ -104,7 +104,11 @@ var publicAgent = {
 
     common.utils.getHostname(function(err, hostname) {
       if (err) {
-        logger.info('Unable to retrieve GCE hostname.', err);
+        if (err.code !== 'ENOTFOUND') {
+          // We are running on GCP.
+          logger.warn('Unable to retrieve GCE hostname.', err);
+        }
+        config.hostname = require('os').hostname();
       } else {
         config.hostname = hostname;
       }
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -104,7 +104,10 @@ var publicAgent = {
 
     common.utils.getHostname(function(err, hostname) {
       if (err) {
-        logger.info('Unable to retrieve GCE hostname.', err);
+        if (err.code !== 'ENOTFOUND') {
+          logger.warn('Unable to retrieve GCE hostname.', err);
+        }
+        config.hostname = require('os').hostname();
```
ofrobots: This logic belongs in getHostname.

index.js
```diff
@@ -112,7 +115,9 @@ var publicAgent = {
 
     common.utils.getInstanceId(function(err, instanceId) {
       if (err) {
-        logger.info('Unable to retrieve GCE instance id.', err);
+        if (err.code !== 'ENOTFOUND') {
+          logger.warn('Unable to retrieve GCE instance id.', err);
+        }
```
ofrobots: I can be fine with checking the return code from getInstanceId, but we should document this in the API. Also, either you need a comment here explaining what `ENOTFOUND` implies, or make the API return a more descriptive error.

index.js
matthewloring: I think it would be best if we made the functions inside util just connect us to different pieces of metadata from the metadata service and resolve local options inside the trace agent the same way we do for project number currently.

ofrobots: Okay, that reasoning works for me.

index.js
matthewloring: Added a comment.

ofrobots:
LGTM.

PR #111, opened 2015-10-26
Reports agent version as a label on every span

### LINKED ISSUE #110: Report agent information for tracking metrics
Add the label "trace.cloud.google.com/agent" with format "\<name\> \<version\>" for tracking.

matthewloring:
Fixes #110.

### DIFF
--- a/lib/trace-writer.js
+++ b/lib/trace-writer.js
@@ -17,15 +17,11 @@
 'use strict';
 
 var utils = require('@google/cloud-diagnostics-common').utils;
-var constants = require('./constants.js');
 var traceLabels = require('./trace-labels.js');
 var pjson = require('../package.json');
 
 /* @const {Array<string>} list of scopes needed to operate with the trace API */
 var SCOPES = ['https://www.googleapis.com/auth/trace.append'];
-
-var defaultHeaders = {};
-defaultHeaders[constants.TRACE_API_HEADER_NAME] = 'node agent ' + pjson.version;
 
 /**
  * Creates a basic trace writer.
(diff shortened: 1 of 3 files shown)


matthewloring:
@ofrobots PTAL.

lib/trace-writer.js
```diff
@@ -74,6 +74,7 @@ TraceWriter.prototype.writeSpan = function(spanData) {
   if (this.config_.instanceId) {
     spanData.addLabel(traceLabels.GCE_INSTANCE_ID, this.config_.instanceId);
   }
+  spanData.addLabel(traceLabels.AGENT_DATA, 'node ' + pjson.version);
```
ofrobots: Can we use the package name instead (or in addition to) of just `node`?

ofrobots:
1. See the duplication here: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/lib/trace-writer.js#L28. Please reconcile.

2. It makes sense to put this (and while we are at it, the hostname and instance id) in the default headers since these don't vary for the lifetime of the app.

lib/trace-writer.js
matthewloring: The name currently is "@google/cloud-trace" which doesn't carry any language specific information (and I don't think it necessarily should). The fact that this is a trace agent can be inferred from the fact that we're reading the information from a trace span.

matthewloring:
1. I am not sure what format the consumer of the header is expecting. I can modify the new label we attach, my hope was to keep the string as small as possible since this label will be appearing on every single trace but I can use the longer form included in the header if preferable.

2. What is the benefit of adding information to the headers without any intended consumers of that information?

matthewloring:
@ofrobots Default headers have been removed as discussed yesterday. Ready for another look.

lib/trace-writer.js
ofrobots: We agreed that keeping the name shorter is better for now.

ofrobots:
LGTM.

PR #122, opened 2015-10-29
Report source IP on spans

### LINKED ISSUE #121: Report source IP on trace spans

matthewloring:
Fixes #121

### DIFF
--- a/test/hooks/test-trace-express.js
+++ b/test/hooks/test-trace-express.js
@@ -15,6 +15,9 @@
  */
 'use strict';
 
+var traceLabels = require('../../lib/trace-labels.js');
+var http = require('http');
+var assert = require('assert');
 var common = require('./common.js');
 var express = require('./fixtures/express4');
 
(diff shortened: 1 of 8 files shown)


matthewloring:
@ofrobots PTAL.

ofrobots:
1. Can you add a test?
2. How confident are we that the `req.connection` will be a valid object in all three frameworks?

matthewloring:
Added tests to ensure that `req.connection` is valid across the board. @ofrobots ready for another look

test/hooks/test-trace-express.js
```diff
@@ -123,6 +126,23 @@ describe('test-trace-express', function() {
...
+    app.get('/', function (req, res) {
+      res.send(common.serverRes);
+    });
+    server = app.listen(common.serverPort, function() {
+      http.get({port: common.serverPort}, function(res) {
+        var labels = common.getMatchingSpan(expressPredicate).labels;
+        assert.equal(labels[traceLabels.HTTP_RESPONSE_CODE_LABEL_KEY], '200');
+        assert.equal(labels[traceLabels.HTTP_METHOD_LABEL_KEY], 'GET');
+        assert.equal(labels[traceLabels.HTTP_URL_LABEL_KEY], 'http://localhost/');
+        assert.equal(labels[traceLabels.HTTP_SOURCE_IP], '::ffff:127.0.0.1');
```
ofrobots: This is dependent on the machine network stack. Maybe check for presence of the property rather than the value.

ofrobots:
LGTM w/ comment.

matthewloring:
@ofrobots I'm going to lock our redis2 fixture version to 2.2.5. It seems like 2.3.0 was just released and it's completely broken. We can take this out once they are stable (or the problem on our end is fixed but we haven't touched that code recently).

PR #126, opened 2015-10-30
Enable sampling by default

### LINKED ISSUE #125: Enable sampling by default

matthewloring:
Chose a rate of 10 traces per second.

Fixes #125

### DIFF
--- a/config.js
+++ b/config.js
@@ -47,7 +47,7 @@ module.exports = {
   // supported and will result in at most 1000 samples per second.
   // Note: `samplingRate` should never be set to 0. To disable tracing, use
   // `enabled` above.
-  samplingRate: -1,
+  samplingRate: 10,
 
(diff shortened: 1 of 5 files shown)


matthewloring:
@ofrobots Ready for a look.

config.js
```diff
@@ -47,7 +47,7 @@ module.exports = {
   // supported and will result in at most 1000 samples per second.
   // Note: `samplingRate` should never be set to 0. To disable tracing, use
```
ofrobots: Not part of this CL, but, what happens when you set it to 0? Users will set it to zero despite our warning. At the least the documentation here should explain what would happen. At best, we should detect samplingRate of 0 and avoid whatever the problem is, or throw.

ofrobots:
LGTM w/ a comment that should be dealt with it another PR.

PR #136, opened 2015-11-02
Run tests once on Travis

### LINKED ISSUE #135: Tests execute twice on travis (once for test once for coverage)
This should happen once.

matthewloring:
Also use the new Travis container infrastructure.

Fixes #135.


matthewloring:
@ofrobots Do we have a strong need for separate coverage and test scripts or could they be merged?

ofrobots:
I'm fine with merging the scripts.

matthewloring:
@ofrobots Coverage now includes all the test logic. Ready for a look.

ofrobots:
Can we have a single script that takes an option command line argument whether or not to run coverage?

matthewloring:
@ofrobots It would be nice to reduce code duplication between the if and else branches in the new script but I'm not sure if it's possible to pass conditional arguments or make arguments go away in some cases for a bash command. Any thoughts?

ofrobots:
You can do something like this:
```bash
C=
if [ coverage ]; then
  C="instanbul cover"
fi
...
$C $(npm bin)/_mocha --dir ./coverage/0 -- test test/hooks --timeout 4000 --R spec
```

matthewloring:
Right, but we also only want the --dir parameter if $C is istanbul cover.

ofrobots:
```bash
M=$(npm bin)/_mocha
if [ need coverage ]; then
  M="instanbul cover $(npm bin)/_mocha --dir ./coverage/0"
fi
...
$M -- test test/hooks
```

matthewloring:
I think since the two coverage calls require different prefixes and there is conditional work to be done per stand alone test and there is conditional cleanup work, trying to merge these branches is actually going to lead to more code. I would be inclined to leave as is to avoid littering if statements everywhere.

matthewloring:
@ofrobots PTAL.

bin/run-test.sh
```diff
@@ -1,18 +1,56 @@
...
+# Lint
 jshint . || exit 1
+
+# Install framework deps
 for dir in test/hooks/fixtures/*/ ;
 do
   echo -en "travis_fold:start:npm_install_${dir}\\r" | tr / _
   echo "npm install in ${dir}"
-  cd ${dir} && npm install && cd -
+  cd "${dir}" && npm install && cd - || exit
```
ofrobots: It would be better to do it this way:
```bash
(cd ${dir}; npm install) || exit 1
```
Parentheses give you a sub-shell, so you don't have to cd back.

matthewloring: Fixed

bin/run-test.sh
```diff
@@ -1,18 +1,56 @@
...
   echo -en "travis_fold:end:npm_install_${dir}\\r" | tr / _
 done
-istanbul test $(npm bin)/_mocha -- test test/hooks --timeout 4000 --R spec || exit 1
+
+# Get test/coverage command
+function command {
+  C="istanbul test"
+  if [ "$cover" ]; then
+    C="istanbul cover --dir ./coverage/$RANDOM"
```
ofrobots: It would be better to avoid random number collisions. Use this instead:

Initialize `counter` globally to 0.
```bash
C="istanbul cover --dir ./coverage/$((counter++))"
```

bin/run-test.sh
```diff
@@ -1,18 +1,56 @@
...
   echo -en "travis_fold:start:npm_install_${dir}\\r" | tr / _
   echo "npm install in ${dir}"
-  cd ${dir} && npm install && cd -
+  cd "${dir}" && npm install && cd - || exit
   echo -en "travis_fold:end:npm_install_${dir}\\r" | tr / _
 done
-istanbul test $(npm bin)/_mocha -- test test/hooks --timeout 4000 --R spec || exit 1
+
+# Get test/coverage command
+function command {
```
ofrobots: it would be better to call this function `prefix`.

bin/run-test.sh
matthewloring: Fixed.

bin/run-test.sh
```diff
@@ -1,18 +1,56 @@
...
+  C="istanbul test"
+  if [ "$cover" ]; then
+    C="istanbul cover --dir ./coverage/$RANDOM"
+  fi
+  echo "$C"
+}
+
+# Run test/coverage
+$(command) "$(npm bin)/_mocha" -- test test/hooks --timeout 4000 --R spec || exit 1
```
ofrobots: Either use `$(npm bin)` for everything (including `istanbul`) or nothing.

bin/run-test.sh
matthewloring: Fixed.

ofrobots:
LGTM w/ comments.

bin/run-test.sh
matthewloring: Using it everywhere for safety.

PR #139, opened 2015-11-04
ignore /_ah/health by default

ofrobots:
A few other minor issues also fixed at the same time
* BSD `sed` on a mac requires an argument for `-i`.
* Don't run jshint on test/hooks/fixtures

### DIFF
--- a/test/standalone/test-default-ignore-ah-health.js
+++ b/test/standalone/test-default-ignore-ah-health.js
@@ -15,24 +15,24 @@
  */
 'use strict';
 
-var agent = require('../..').start({ignoreUrls: ['/test'], samplingRate: -1});
+var agent = require('../..').start({samplingRate: -1});
 
 var assert = require('assert');
 var http = require('http');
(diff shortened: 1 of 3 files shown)


ofrobots:
Dropped the `sed` change as the empty string was getting lost in translation somewhere between node.js and actual execution on travis.

ofrobots:
@matthewloring PTAL.

matthewloring:
LGTM.

test/standalone/test-default-ignore-ah-health.js
```diff
@@ -0,0 +1,43 @@
...
+
+var assert = require('assert');
+var http = require('http');
+var express = require('../hooks/fixtures/express4');
+
+describe('test-default-ignore-ah-health', function() {
+  it('should ignore /_ah/health traces by default', function(done) {
+    var app = express();
+    app.get('/_ah/health', function (req, res) {
+      res.send('🏥');
```
matthewloring: nice touch!

PR #153, opened 2015-11-06
Identify root spans with inherited context

### LINKED ISSUE #152: Trace headers are not added to all outgoing http requests

matthewloring:
Root spans that inherit their trace context from incoming headers are
now properly identified as root spans and logged. Independently this
fixes and tests a bug where outgoing contexts were not attached to all
requests.

Fixes #152.

### DIFF
--- a/lib/hooks/core/hook-http.js
+++ b/lib/hooks/core/hook-http.js
@@ -21,6 +21,7 @@ var httpAgent = require('_http_agent');
 var cls = require('../../cls.js');
 var TraceLabels = require('../../trace-labels.js');
 var SpanData = require('../../span-data.js');
+var util = require('util');
 var url = require('url');
 var shimmer = require('shimmer');
 /** @type {TraceAgent} */
@@ -33,14 +34,12 @@ function requestWrap(request) {
       return request.apply(this, arguments);
     }
 
-    var localOptions = (typeof options === 'string') ?
-      url.parse(options) : options;
-
     // Don't trace ourselves lest we get into infinite loops
     // Note: this would not be a problem if we guarantee buffering
     // of trace api calls. If there is no buffering then each trace is
     // an http call which will get a trace which will be an http call
-    if (agent.isTraceAgentRequest(localOptions)) {
+    if (typeof options === 'object' &&
+        agent.isTraceAgentRequest(options)) {
       return request.apply(this, arguments);
     }
 
@@ -54,17 +53,19 @@ function requestWrap(request) {
       return request.apply(this, arguments);
     }
 
+    options = (typeof options === 'string') ?
+      url.parse(options) : clone(options);
+    options.headers = options.headers || {};
+
     var namespace = cls.getNamespace();
     var uri = uriFromOptions(options);
     var labels = {};
     labels[TraceLabels.HTTP_METHOD_LABEL_KEY] = options.method;
     labels[TraceLabels.HTTP_URL_LABEL_KEY] = uri;
     var span = agent.startSpan(uri, labels);
-    if (options.headers) {
-      // Adding context to the headers lets us trace the request
-      // as it makes it through other layers of the Google infrastructure
-      agent.addContextToHeaders(span, options.headers);
-    }
+    // Adding context to the headers lets us trace the request
+    // as it makes it through other layers of the Google infrastructure
+    agent.addContextToHeaders(span, options.headers);
 
     var returned = request.call(this, options, function(res) {
       namespace.bindEmitter(res);
@@ -99,6 +100,12 @@ function requestWrap(request) {
   };
 }
 
+function clone(o) {
+  var r = {};
+  util._extend(r, o);
+  return r;
+}
+
 function uriFromOptions(options) {
   var uri;
   if (typeof options === 'string') {
@@ -112,8 +119,9 @@ function uriFromOptions(options) {
     var defaultAgent = options._defaultAgent || httpAgent.globalAgent;
     var protocol = options.protocol || defaultAgent.protocol;
     var host = options.hostname || options.host || 'localhost';
+    var port = options.port ? (':' + options.port) : '';
     var path = options.path || options.pathname || '/';
-    uri = protocol + '//' + host + path;
+    uri = protocol + '//' + host + port + path;
   }
   return uri;
 }
--- a/test/standalone/test-trace-header-context.js
+++ b/test/standalone/test-trace-header-context.js
@@ -1,0 +1,76 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


matthewloring:
@ofrobots PTAL.

lib/hooks/core/hook-http.js
```diff
@@ -35,6 +35,7 @@ function requestWrap(request) {
 
     var localOptions = (typeof options === 'string') ?
       url.parse(options) : options;
+    localOptions.headers = localOptions.headers || {};
```
ofrobots: An alternative here would be to do:
```javascript
options = (..) ? url.parse(options) : options
```
We are trying to be careful to avoid changing the caller's options object, but if that is really our intent; the code at present is not rigorous enough. This statement adds a header property on the original options object if the original was not a string.

So: either make it more rigorous (and add a comment) or relent.

ofrobots: For reference; here's how request deals with this: https://github.com/request/request/blob/master/request.js#L126.

ofrobots: Lastly, if we are to do any object copying, this needs to happen *after* the root span checks below. i.e. only traces that are actually being sampled.

matthewloring: I've made things more rigorous! We only copy when we know we need to and we don't modify their options object at all.

matthewloring:

lib/hooks/core/hook-http.js
```diff
@@ -54,19 +53,21 @@ function requestWrap(request) {
       return request.apply(this, arguments);
     }
 
+    var localOptions = (typeof options === 'string') ?
```
ofrobots: Use the same check as above, i.e. `typeof options === 'object'`. That way the optimizer (value numbering) can realize that it is the same check as before, and void the second check completely.

ofrobots: Since `options` are not referenced below this line; just assign to `options`. That way we cannot have an errant use of the original `options` object below this line.

matthewloring: Fixed.

matthewloring:

lib/hooks/core/hook-http.js
```diff
@@ -54,19 +53,21 @@ function requestWrap(request) {
       return request.apply(this, arguments);
     }
 
+    var localOptions = (typeof options === 'object') ?
```
ofrobots: Did you miss my other comment? Get rid of the `localOptions` variable and assign to `options`.

matthewloring: Sorry for the misunderstanding, `localOptions` has been eliminated.

test/standalone/test-trace-header-context.js
```diff
@@ -0,0 +1,76 @@
...
+'use strict';
+
+var common = require('../hooks/common.js');
+var http = require('http');
+var assert = require('assert');
+var express = require('../hooks/fixtures/express4');
+var constants = require('../../lib/constants.js');
+
+describe('test-trace-header-context', function() {
+  it('should work string url', function(done) {
```
ofrobots: *work with

test/standalone/test-trace-header-context.js
```diff
@@ -0,0 +1,76 @@
...
+      common.cleanTraces();
+      server.close();
+      done();
+    });
+    server = app.listen(common.serverPort, function() {
+      http.get({ port: common.serverPort });
+    });
+  });
+
+  it('should work options object url', function(done) {
```
ofrobots: *work with

test/standalone/test-trace-header-context.js
```diff
@@ -0,0 +1,76 @@
...
+    app.get('/self', function(req, res) {
+      assert(req.headers[constants.TRACE_CONTEXT_HEADER_NAME.toLowerCase()]);
+      res.send(common.serverRes);
+      var traces = common.getTraces();
+      assert.equal(traces.length, 2);
+      assert.equal(traces[0].spans.length, 2);
+      assert.equal(traces[1].spans.length, 1);
+      assert.equal(traces[0].spans[0].name, '/');
+      assert.equal(traces[0].spans[1].name, 'http://localhost/self');
+      assert.equal(traces[1].spans[0].name, '/self');
```
ofrobots: How would these show up in the trace UI? Are the incoming and outgoing spans are distinguishable? I know that full url shows up when it is an outgoing request, but hopefully that's not the only thing that distinguishes those.

test/standalone/test-trace-header-context.js
```diff
@@ -0,0 +1,76 @@
...
+    });
+    app.get('/self', function(req, res) {
+      assert(req.headers[constants.TRACE_CONTEXT_HEADER_NAME.toLowerCase()]);
+      res.send(common.serverRes);
+      var traces = common.getTraces();
+      assert.equal(traces.length, 2);
+      assert.equal(traces[0].spans.length, 2);
+      assert.equal(traces[1].spans.length, 1);
+      assert.equal(traces[0].spans[0].name, '/');
+      assert.equal(traces[0].spans[1].name, 'http://localhost/self');
```
ofrobots: Observation: we don't include the port in the formatted url !

ofrobots:
LGTM, but can you verify that `url.parse` works correctly even for non-object, non-string argument.

test/standalone/test-trace-header-context.js
matthewloring: Fixed.

test/standalone/test-trace-header-context.js
matthewloring: Fixed.

test/standalone/test-trace-header-context.js
matthewloring: The labels on these spans will distinguish those generated by express from those generated by http. All spans generated by http represent outgoing requests.

matthewloring:
From the node documentation, "options can be an object or a string. If options is a string, it is automatically parsed with url.parse().". Url parse will throw an exception for non-object, non-string arguments but that is exactly what the documentation says will happen.

ofrobots:
Thanks for checking. I think the behavior we want here is that we should allow the exception to be thrown, but we shouldn't be the throwers. If the fix looks ugly, feel free to ignore my previous recommendation for optimization.

test/standalone/test-trace-header-context.js
ofrobots: sgtm.

matthewloring:
The fix will just involve doing an additional set of string and object type checks. I will undo the optimization for now since it would result in more total checks (even if some get optimized away).

test/standalone/test-trace-header-context.js
matthewloring: Good addition, added.

lib/hooks/core/hook-http.js
```diff
@@ -33,14 +34,12 @@ function requestWrap(request) {
...
 
-    var localOptions = (typeof options === 'string') ?
-      url.parse(options) : options;
-
     // Don't trace ourselves lest we get into infinite loops
     // Note: this would not be a problem if we guarantee buffering
     // of trace api calls. If there is no buffering then each trace is
     // an http call which will get a trace which will be an http call
-    if (agent.isTraceAgentRequest(localOptions)) {
+    if (typeof options === 'object' &&
```
ofrobots: This check seems unnecessary. `isTraceAgentRequest` checks that options is truthy, and has a headers property.

PR #156, opened 2015-11-06
Fix mongodb crash if project num isn't found

### LINKED ISSUE #149: Agent crashes when unable to get project number
App listening at http://:::8080
WARN :@google/cloud-trace: Cannot trace outbound http requests outside of a supported http framework.
WARN :@google/cloud-trace: Cannot trace outbound http requests outside of a supported http framework.
WARN :@google/cloud-trace: Cannot trace outbound http requests outside of a supported http framework.
WARN :@google/cloud-trace: Cannot trace outbound http requests outside of a supported http framework.
ERROR:@google/cloud-trace: Unable to acquire the project number from metadata service. Please provide a valid project number as an env. variable, or through config.projectId passed to start().Error: Could not auto-discover project-id. Please export GCLOUD_PROJECT_NUM with your numeric project id
/Users/ofrobots/src/cloud-trace-performance/bookshelf/bookshelf/node_modules/mongodb/lib/mongo_client.js:458
          throw err
          ^

TypeError: Cannot read property 'logger' of null
    at next_trace [as _next] (/Users/ofrobots/src/cloud-trace-performance/bookshelf/bookshelf/node_modules/@google/cloud-trace/lib/hooks/userspace/hook-mongodb-core.js:31:12)
    at fetchDocs (/Users/ofrobots/src/cloud-trace-performance/bookshelf/bookshelf/node_modules/mongodb/lib/cursor.js:838:10)
    at toArray (/Users/ofrobots/src/cloud-trace-performance/bookshelf/bookshelf/node_modules/mongodb/lib/cursor.js:865:3)
    at Cursor.toArray (/Users/ofrobots/src/cloud-trace-performance/bookshelf/bookshelf/node_modules/mongodb/lib/cursor.js:818:44)
    at /Users/ofrobots/src/cloud-trace-performance/bookshelf/bookshelf/books/model-mongodb.js:67:10
    at /Users/ofrobots/src/cloud-trace-performance/bookshelf/bookshelf/books/model-mongodb.js:54:7
    at /Users/ofrobots/src/cloud-trace-performance/bookshelf/bookshelf/node_modules/mongodb/lib/mongo_client.js:455:11
    at /Users/ofrobots/src/cloud-trace-performance/bookshelf/bookshelf/node_modules/@google/cloud-trace/node_modules/continuation-local-storage/node_modules/async-listener/glue.js:188:31
    at doNTCallback0 (node.js:417:9)
    at process._tickCallback (node.js:346:13)

matthewloring:
The crash is not reproducible for the other frameworks as those patches
do not persist on long lived objects and thus execute after hooks have
been disabled. This PR also fixes a race condition crash that could
occur if hooks were disabled between the start and end of a single span
(resulting in a null dereference of the agent).

Fixes #149

### DIFF
--- a/lib/hooks/core/hook-http.js
+++ b/lib/hooks/core/hook-http.js
@@ -24,7 +24,6 @@ var SpanData = require('../../span-data.js');
 var util = require('util');
 var url = require('url');
 var shimmer = require('shimmer');
-/** @type {TraceAgent} */
 var agent;
 
 // http.request
@@ -135,7 +134,7 @@ module.exports = function(version_, agent_) {
       },
       unpatch: function(http) {
         shimmer.unwrap(http, 'request');
-        agent = null;
+        agent_.logger.info('http: unpatched');
       }
     }
   };
--- a/lib/hooks/userspace/hook-express.js
+++ b/lib/hooks/userspace/hook-express.js
@@ -40,9 +40,7 @@ function applicationActionWrap(method) {
 function middleware(req, res, next) {
   var namespace = cls.getNamespace();
   if (!namespace) {
-    if (agent) {
-      agent.logger.info('Express: no namespace found, ignoring request');
-    }
+    agent.logger.info('Express: no namespace found, ignoring request');
     return next();
   }
   namespace.bindEmitter(req);
@@ -124,8 +122,7 @@ module.exports = function(version_, agent_) {
         patchedMethods.forEach(function(method) {
           shimmer.unwrap(express.application, method);
         });
-        agent.logger.info('Express: unpatched');
-        agent = null;
+        agent_.logger.info('Express: unpatched');
       }
     }
   };
--- a/lib/hooks/userspace/hook-hapi.js
+++ b/lib/hooks/userspace/hook-hapi.js
@@ -36,9 +36,7 @@ function connectionWrap(connection) {
 function middleware(request, reply) {
   var namespace = cls.getNamespace();
   if (!namespace) {
-    if (agent) {
-      agent.logger.info('Hapi: no namespace found, ignoring request');
-    }
+    agent.logger.info('Hapi: no namespace found, ignoring request');
     return reply.continue();
   }
   var req = request.raw.req;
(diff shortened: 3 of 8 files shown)


matthewloring:
@ofrobots PTAL.

ofrobots:
I would prefer if we could make the hook completely stateless w.r.t. `agent`. Taking `hook-http.js` as an example, you can get rid of the global variable, and instead use `agent_` from the closure.

Even though you have fixed the null/non-null race condition, we still have a race between two different (non-null) trace agents being accessible in the file.

WDYT?

matthewloring:
I'm not sure I understand the second race condition you mentioned. My understanding is that we only ever have one 'real' agent and one 'phantom' agent. The value of `agent` may switch between these mid-hook execution but the semantics aren't broken if they do. Is that correct?

ofrobots:
Semantics ought not to be broken if they switch mid hook execution. But why
take that risk when we can avoid it? Closure solves this more robustly.

matthewloring:
Sorry, I have misspoken. There are never two different agents from the standpoint of the hooks. The hooks are initialized and only ever reassigned to the 'active' agent. The hooks will never see the non-active agent. I can move the hook logic into the closure of module.exports but the semantics are identical to those before since the value of `agent` is constant from the standpoint of the hooks.

ofrobots:
Acknowledged. Looking at `require('./lib/trace-agent.js').get()`, it seems that upon repeated start/stop, a new trace agent object is created.

Here's an invariant I want. Once a hook has been applied, it should never be possible it to access an agent allocated in the future. The race condition is when you stop and start the agent. We patch the hook and the value of agent global variable gets modified. A currently in-flight request may now refer to the newly allocated agent.

It would be good defensive programming here if we weren't dependent on the agents to be idempotent and instead made the hook stateless w.r.t the agent.

matthewloring:
@ofrobots It sounds like making this closure change makes more sense in a future PR. PTAL.

ofrobots:
lgtm.

PR #162, opened 2015-11-09
Parsed spanids should have type number

### LINKED ISSUE #161: Span ids should be of number type when parsed from headers.

matthewloring:
Fixes #161

### DIFF
--- a/lib/trace-agent.js
+++ b/lib/trace-agent.js
@@ -197,12 +197,13 @@ TraceAgent.prototype.parseContextFromHeader = function(str) {
     return null;
   }
   var matches = str.match(/^([0-9a-fA-F]+)(?:\/([0-9a-fA-F]+))?(?:;o=(.*))?/);
-  if (!matches || matches.length !== 4 || matches[0] !== str) {
+  if (!matches || matches.length !== 4 || matches[0] !== str ||
+      (matches[2] && isNaN(matches[2]))) {
     return null;
   }
   return {
     traceId: matches[1],
-    spanId: matches[2],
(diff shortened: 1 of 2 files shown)


matthewloring:
@ofrobots PTAL.

lib/trace-agent.js
```diff
@@ -202,7 +202,7 @@ TraceAgent.prototype.parseContextFromHeader = function(str) {
   }
   return {
     traceId: matches[1],
-    spanId: matches[2],
+    spanId: Number(matches[2]),
```
ofrobots: When the argument cannot be converted to a number, `Number(argument)` returns `NaN`. I don't think the Trace API would like seeing `NaN`s here.

test/test-trace-agent.js
```diff
@@ -96,7 +97,7 @@ describe('Trace Agent', function() {
           '123456/667');
         assert(result);
         assert.equal(result.traceId, '123456');
-        assert.equal(result.spanId, '667');
+        assert.equal(result.spanId, 667);
         assert(!result.options);
       });
```
ofrobots: Add test(s) to deal with a spanId that cannot be parsed as a number.

lib/trace-agent.js
matthewloring: Fixed.

test/test-trace-agent.js
matthewloring: I've added the tests to the invalid inputs section below. I think it should be a parse error if we get a non-numeric spanId.

ofrobots:
lgtm

PR #164, opened 2015-11-10
Propagate context through sequelize correctly

matthewloring:


matthewloring:
Note: This patch prevents users of sequelize from specifying their own custom cls namespaces for use with sequelize. They are still able to achieve the same outcome by storing their desired values inside of our namespace. If we don't want this caveat, we will have to patch bluebird directly.

matthewloring:
@ofrobots PTAL.

matthewloring:
We have determined that bluebird is not the culprit for context loss in sequelize and rather, it is the connection pool module.

Issue #169, opened 2015-11-12
MVM: Add module labels to the captured traces


ofrobots:
From an email bug report:
> [...] individual traces are not correctly associated with their module, ie. default, worker, etc. They only appear in Pantheon if I select "All modules"

jonparrott:
To add to this, should GCE VMs in general use their hostname as the module?

ofrobots:
Client side fix landed, but there might be an issue on the service side. Re-opening until the service side issue is resolved.

ofrobots:
Fixes landed in f1fa3819138e616f946476b7b362eebb020afa1d and 4471696cac16ad75ab5f896009699141908a9788. We'll try to get these into a release later in the week.

Also included is the change for GCE/GKE VMs to report their hostname as the module. Thanks for the suggestion!

PR #172, opened 2015-11-13
attach GAE labels when available

ofrobots:
Module and version drop down boxes in the Trace UI use these labels to allow
filtering by module and version. For GCE, GKE we use the hostname as the module
to allow the same functionality in the UI.

Fixes: #169

### DIFF
--- a/index.js
+++ b/index.js
@@ -159,6 +159,13 @@ var publicAgent = {
       return this;
     }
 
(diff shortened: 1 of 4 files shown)


ofrobots:
@matthewloring PTAL.

matthewloring:
LGTM.

PR #176, opened 2015-11-20
host,version,app labels computed only once

ofrobots:


lib/trace-writer.js
```diff
@@ -54,8 +57,63 @@ function TraceWriter(logger, config) {
...
+      Object.freeze(labels);
+      that.defaultLabels_ = labels;
+    });
+  });
 }
 
+TraceWriter.prototype.getHostname = function(cb) {
+  var that = this;
+  utils.getHostname(headers, function(err, hostname) {
+    if (err) {
```
matthewloring: collapse these conditionals?

lib/trace-writer.js
```diff
@@ -54,8 +57,63 @@ function TraceWriter(logger, config) {
...
+      }
+    }
+    cb(hostname);
+  });
+};
+
+TraceWriter.prototype.getInstanceId = function(cb) {
+  var that = this;
+  utils.getInstanceId(headers, function(err, instanceId) {
+    if (err) {
```
matthewloring: same here

lib/trace-writer.js
```diff
@@ -68,32 +126,10 @@ TraceWriter.prototype.writeSpan = function(spanData) {
...
-      if (moduleName !== 'default') {
-        versionLabel = moduleName + ':';
-      }
-      versionLabel += this.config_.gae_module_version + '.' +
-          this.config_.gae_minor_version;
-      spanData.addLabel(traceLabels.GAE_VERSION, versionLabel);
-    }
+  // Copy properties from the default labels.
+  for (var k in this.defaultLabels_) {
+    spanData.addLabel(k, this.defaultLabels_[k]);
```
matthewloring: This is minor but it might be nice to initialize span data with default labels as opposed to copying them in one by one for every span.

ofrobots:
@matthewloring PTAL.

lib/trace-writer.js
```diff
@@ -54,8 +57,59 @@ function TraceWriter(logger, config) {
...
+      }
+      var moduleName = process.env.GAE_MODULE_NAME || hostname;
+      labels[traceLabels.GAE_MODULE_NAME] = moduleName;
+
+      var moduleVersion = process.env.GAE_MODULE_VERSION;
+      if (moduleVersion) {
+        labels[traceLabels.GAE_MODULE_VERSION] = moduleVersion;
+        var minorVersion = process.env.GAE_MINOR_VERSION;
+        if (minorVersion) {
+          var versionLabel = '';
```
matthewloring: It would be nice to move this craziness into a helper function for easier reasoning and so it can be easily eliminated once it is no longer necessary (which hopefully will be soon).

matthewloring:
LGTM w/ optional suggestion.

lib/trace-writer.js
ofrobots: I am not sure this will become unnecessary any time soon.

matthewloring: :cry: This comment was just a minor suggestion. LGTM

PR #177, opened 2015-11-21
WIP: Avoid unnecessary emitter binding

ofrobots:
@matthewloring this is not complete yet, but WDYT?

### DIFF
--- a/lib/hooks/userspace/hook-express.js
+++ b/lib/hooks/userspace/hook-express.js
@@ -43,6 +43,10 @@ function middleware(req, res, next) {
     agent.logger.info('Express: no namespace found, ignoring request');
     return next();
   }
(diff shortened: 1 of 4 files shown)


matthewloring:
We need to make sure that the context is still set to be null so that db hooks don't break but otherwise it seems solid. I would be curious which bound emitters are accounting for the majority of the slow down. My guess would be that the req/res emitters emit relatively less frequently than some of the patches in mongo (specifically the pool event emitter). This patch should definitely be a performance improvement but I worry that mongo may account for a larger overall hit.

ofrobots:
I think it is the express emitters that are causing most of the overhead.

matthewloring:
Awesome, then this should have a pretty big impact. The idea looks solid as long as the context remains set.

matthewloring:
Continuing in #178

PR #180, opened 2015-11-25
Do not modify methods in the module cache

ofrobots:
@matthewloring PTAL.


matthewloring:
LGTM. Any idea what changed in 5 to cause breakages to begin there?

ofrobots:
No idea at this point. This should be broken on all versions of node.

matthewloring:
That's scary, I can look back through the node PRs and see if any culprits stand out but I think this is good to go in. It would be good to add a test but this is a pretty specific thing to test for and the more general property (not modifying cached modules) is tough to isolate in unit tests for all modules.

ofrobots:
I don't think there is much value in a post facto test for this specific scenario. I think e2e tests are the right answer here.

PR #183, opened 2015-11-27
fix GAE_MODULE fall-back to local hostname

ofrobots:
We were sending undefined as the GAE_MODULE label.

@matthewloring ptal.


matthewloring:
The change looks good, unfortunately, it looks like redis 2.3 -> 2.4 broke our hooks and is causing travis to go red. This may be the first instance where we need separate patching logic at the minor version level.

matthewloring:
CI failure should be a flake. LGTM when it goes green.

PR #184, opened 2015-11-28
restrict redis support to <= 2.3.x

ofrobots:
This is a work-arounds a change is redis 2.4 that breaks our tracing.


ofrobots:
@matthewloring PTAL.

matthewloring:
LGTM. I experimented with a fix for 2.4 is it should be isomorphic to what exists now with a few changed function names.

PR #186, opened 2015-11-28
increase test-no-self-tracing timeout

ofrobots:
Travis intermittently fails on these.

@matthewloring PTAL.


matthewloring:
LGTM.

PR #188, opened 2015-11-29
Support for redis 2.4

### LINKED ISSUE #185: Support redis 2.4

matthewloring:
Fixes #185


matthewloring:
@ofrobots PTAL.

lib/hooks/userspace/hook-redis.js
```diff
@@ -32,6 +32,23 @@ function createClientWrap(createClient) {
   };
 }
 
+// Used for redis version > 2.3
+function createStreamWrap(create_stream) {
+  return function create_stream_trace() {
+    if (!this.stream) {
+      Object.defineProperty(this, 'stream', {
```
ofrobots: Why does the define property need to happen at `create_stream` invocation? Could you define the proxy by wrapping the constructor?

lib/hooks/userspace/hook-redis.js
```diff
@@ -32,6 +32,23 @@ function createClientWrap(createClient) {
...
+        get: function () { return this._trace_stream; },
```
ofrobots: The monkey-patched property needs to be a longer string, like how we do this in express, to reduce collision probability.

lib/hooks/userspace/hook-redis.js
matthewloring: The issue with wrapping the constructor is that before constructor invocation, the `this` object isn't accessible to my knowledge and after constructor invocation, it is already too late and create_streams has been called (https://github.com/NodeRedis/node_redis/blob/master/index.js#L109). This was the only way that I could think of to get a handle on `this` before the constructor finished executing.

lib/hooks/userspace/hook-redis.js
matthewloring: Done.

matthewloring:
@ofrobots PTAL.

ofrobots:
LGTM.

PR #190, opened 2015-12-01
load cls before all else

matthewloring:
This ensures that the code async APIs get patched befored anything in
user-space can get loaded and interfere with tracing. This is a
problem with request in particular as it caches setImmediate.

Fixes: #182
Lands: #187

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,6 +16,10 @@
 
 'use strict';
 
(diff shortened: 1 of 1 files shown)


matthewloring:
Transitive LGTM from #187

PR #192, opened 2015-12-09
Experimental tracing hooks for koa

### LINKED ISSUE #191: Add automatic tracing of HTTP requests for koa

matthewloring:
Fixes #191

### DIFF
--- a/bin/run-test.sh
+++ b/bin/run-test.sh
@@ -42,7 +42,10 @@ function run {
 run test test/hooks
 for test in test/standalone/test-*.js ;
 do
-  run "${test}"
+  if [[ ! $(node --version) =~ v0\.12\..* || ! "${test}" =~ .*trace\-koa\.js ]]
+  then
+    run "${test}"
+  fi
 done
 
 # Conditionally publish coverage
--- a/lib/hooks/userspace/hook-koa.js
+++ b/lib/hooks/userspace/hook-koa.js
@@ -1,0 +1,131 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const cls = require('../../cls.js');
+const TraceLabels = require('../../trace-labels.js');
+const shimmer = require('shimmer');
+const semver = require('semver');
+const constants = require('../../constants.js');
+var agent;
+
+const SUPPORTED_VERSIONS = '1.x';
+
+function useWrap(use) {
+  return function useTrace() {
+    if (!this._google_trace_patched) {
+      this._google_trace_patched = true;
+      this.use(middleware);
+    }
+    return use.apply(this, arguments);
+  };
+}
+
+function* middleware(next) {
+  /* jshint validthis:true */
+  const namespace = cls.getNamespace();
+  if (!namespace) {
+    agent.logger.info('Koa: no namespace found, ignoring request');
+    return;
+  }
+  if (!agent.shouldTrace(this.req.url)) {
+    return;
+  }
+  const req = this.req;
+  const res = this.res;
+
+  namespace.bindEmitter(req);
+  namespace.bindEmitter(res);
+
+  const originalEnd = res.end;
+
+  namespace.run(function() {
+    const rootContext = startRootSpanForRequest(req);
+
+    // wrap end
+    res.end = function(chunk, encoding) {
+      res.end = originalEnd;
+      const returned = res.end(chunk, encoding);
+
+      endRootSpanForRequest(rootContext, req, res);
+      return returned;
+    };
+    namespace.bind(next);
+  });
+  yield next;
+}
+
+/**
+ * Creates and sets up a new root span for the given request.
+ * @param {Object} req The request being processed.
+ * @returns {!SpanData} The new initialized trace span data instance.
+ */
+function startRootSpanForRequest(req) {
+  const result = agent.parseContextFromHeader(
+    req.headers[constants.TRACE_CONTEXT_HEADER_NAME]) || {};
+
+  const traceId = result.traceId;
+  const parentSpanId = result.spanId;
+  const url = (req.headers['X-Forwarded-Proto'] || 'http') +
+    '://' + req.headers.host + req.url;
+
+  // we use the path part of the url as the span name and add the full
+  // url as a label
+  const rootContext = agent.createRootSpanData(req.url, traceId,
+    parentSpanId);
+  rootContext.addLabel(TraceLabels.HTTP_METHOD_LABEL_KEY, req.method);
+  rootContext.addLabel(TraceLabels.HTTP_URL_LABEL_KEY, url);
+  rootContext.addLabel(TraceLabels.HTTP_SOURCE_IP, req.connection.remoteAddress);
+  return rootContext;
+}
+
+
+/**
+ * Ends the root span for the given request.
+ * @param {!SpanData} rootContext The span to close out.
+ * @param {Object} req The request being processed.
+ * @param {Object} res The response being processed.
+ */
+function endRootSpanForRequest(rootContext, req, res) {
+  if (req.route && req.route.path) {
+    rootContext.addLabel(
+      'koa/request.route.path', req.route.path);
+  }
+  rootContext.addLabel(
+      TraceLabels.HTTP_RESPONSE_CODE_LABEL_KEY, res.statusCode);
+  rootContext.close();
+}
+
+module.exports = function(version_, agent_) {
+  if (!semver.satisfies(version_, SUPPORTED_VERSIONS)) {
+    agent_.logger.info('Koa: unsupported version ' + version_ + ' loaded');
+    return {};
+  }
+  return {
+    // An empty relative path here matches the root module being loaded.
+    '': {
+      patch: function(koa) {
+        agent = agent_;
+        shimmer.wrap(koa.prototype, 'use', useWrap);
+      },
+      unpatch: function(koa) {
+        shimmer.unwrap(koa.prototype, 'use');
+        agent_.logger.info('Koa: unpatched');
+      }
+    }
+  };
+};
--- a/lib/hooks/index.js
+++ b/lib/hooks/index.js
@@ -43,6 +43,8 @@ var toInstrument = Object.create(null, {
       patches: {} } },
   'http': { enumerable: true, value: { file: './core/hook-http.js',
       patches: {} } },
+  'koa': { enumerable: true, value: { file: './userspace/hook-koa.js',
+      patches: {} } },
   'mongodb-core': { enumerable: true, value: { file: './userspace/hook-mongodb-core.js',
       patches: {} } },
   'mysql': { enumerable: true, value: { file: './userspace/hook-mysql.js',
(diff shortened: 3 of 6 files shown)


matthewloring:
@ofrobots I need to experiment more to make sure we're not losing context for DB operations made inside koa but the current state of the hooks is capable of tracing incoming requests. PTAL.

lib/hooks/userspace/hook-koa.js
```diff
@@ -0,0 +1,130 @@
...
+  return function useTrace() {
+    if (!this._google_trace_patched) {
+      this._google_trace_patched = true;
+      this.use(middleware);
+    }
+    return use.apply(this, arguments);
+  };
+}
+
+function *middleware(next) {
```
ofrobots: `function*` as per the guide w/ reasoning: http://www.2ality.com/2014/08/formatting-generator-asterisk.html

lib/hooks/userspace/hook-koa.js
```diff
@@ -0,0 +1,130 @@
+/**
```
ofrobots: Meta: since this file is using generators already, should it use more ES6 features?

lib/hooks/userspace/hook-koa.js
```diff
@@ -0,0 +1,130 @@
...
+ * @returns {!SpanData} The new initialized trace span data instance.
+ */
+function startRootSpanForRequest(req) {
+  var result = agent.parseContextFromHeader(
+    req.headers[constants.TRACE_CONTEXT_HEADER_NAME]) || {};
+
+  var traceId = result.traceId;
+  var parentSpanId = result.spanId;
+  var url = (req.headers['X-Forwarded-Proto'] || 'http') +
+    '://' + req.headers.host + req.url;
```
ofrobots: Looks complicated. Why is `req.href` not adequate?

lib/hooks/userspace/hook-koa.js
```diff
@@ -0,0 +1,130 @@
...
+
+  // we use the path part of the url as the span name and add the full
+  // url as a label
+  var rootContext = agent.createRootSpanData(req.url, traceId,
+    parentSpanId);
+  rootContext.addLabel(TraceLabels.HTTP_METHOD_LABEL_KEY, req.method);
+  rootContext.addLabel(TraceLabels.HTTP_URL_LABEL_KEY, url);
+  rootContext.addLabel(TraceLabels.HTTP_SOURCE_IP, req.connection.remoteAddress);
```
ofrobots: `req.ip` might work too, and better, based on my reading of the koa docs. ManagedVMs are a NAT setup, so connection.remoteAddress is most likely not really what we want, and `X-Forwarded-For` is what we need (separate bug that probably needs fixing for express, hapi, ...).

lib/hooks/userspace/hook-koa.js
```diff
@@ -0,0 +1,130 @@
...
+
+
+/**
+ * Ends the root span for the given request.
+ * @param {!SpanData} rootContext The span to close out.
+ * @param {Object} req The request being processed.
+ * @param {Object} res The response being processed.
+ */
+function endRootSpanForRequest(rootContext, req, res) {
+  if (req.route && req.route.path) {
```
ofrobots: koa docs don't describe these properties. Are these express-specific, or are they defined by Node?

lib/hooks/userspace/hook-koa.js
```diff
@@ -0,0 +1,130 @@
...
+    rootContext.addLabel(
+      'koa/request.route.path', req.route.path);
+  }
+  rootContext.addLabel(
+      TraceLabels.HTTP_RESPONSE_CODE_LABEL_KEY, res.statusCode);
```
ofrobots: I'm confused by the koa docs on what the difference between `res.statusCode` and `res.status` is.

test/hooks/test-trace-koa.js
```diff
@@ -0,0 +1,46 @@
...
+          cb(null, common.serverRes);
+        }, common.serverWait);
+      };
+    });
+    server = app.listen(common.serverPort, function() {
+      common.doRequest('GET', done, koaPredicate);
+    });
+  });
+});
+
```
ofrobots: Needs tests to make sure the labels are arriving with sensible values.

lib/hooks/userspace/hook-koa.js
```diff
@@ -0,0 +1,130 @@
...
+var shimmer = require('shimmer');
+var semver = require('semver');
+var constants = require('../../constants.js');
+var agent;
+
+var SUPPORTED_VERSIONS = '1.x';
+
+function useWrap(use) {
+  return function useTrace() {
+    if (!this._google_trace_patched) {
```
ofrobots: Is this necessary in the koa case?

lib/hooks/userspace/hook-koa.js
ofrobots: Found this:

> Bypassing Koa's response handling is not supported. Avoid using the following node properties:
> res.statusCode

lib/hooks/userspace/hook-koa.js
ofrobots: Ignorez, s'il vous plaît.

lib/hooks/userspace/hook-koa.js
```diff
@@ -0,0 +1,130 @@
...
+  if (!semver.satisfies(version_, SUPPORTED_VERSIONS)) {
+    agent_.logger.info('Koa: unsupported version ' + version_ + ' loaded');
+    return {};
+  }
+  return {
+    // An empty relative path here matches the root module being loaded.
+    '': {
+      patch: function(koa) {
+        agent = agent_;
+        shimmer.wrap(koa.prototype, 'use', useWrap);
```
ofrobots: Do we have to patch use? Could we patch the top-level function returned by `require('koa')`? That may avoid the monkey-patching of the `_google_trace_patched` property.

test/hooks/test-trace-koa.js
```diff
@@ -0,0 +1,46 @@
...
+describe('test-trace-koa', function() {
+  afterEach(function() {
+    common.cleanTraces();
+    server.close();
+  });
+
+  it('should accurately measure get time, get', function(done) {
+    var app = koa();
+    app.use(function *() {
```
ofrobots: `function*`

lib/hooks/userspace/hook-koa.js
matthewloring: Fixed.

lib/hooks/userspace/hook-koa.js
matthewloring: I've added some `consts` throughout. If there are particular features you have in mind I can definitely incorporate them.

lib/hooks/userspace/hook-koa.js
matthewloring: That property is `undefined` in all of my testing.

lib/hooks/userspace/hook-koa.js
matthewloring: `req.ip` is `undefined` in my testing as well. I believe I'm accessing it as described in the docs. Not sure why these are `undefined`.

lib/hooks/userspace/hook-koa.js
matthewloring: They happen to be present in express (and in koa apparently). We can steer away from using undocumented things but the koa documentation doesn't appear to be fully accurate (or at least doesn't describe when request properties will be undefined.

lib/hooks/userspace/hook-koa.js
matthewloring: Pas de problème.

test/hooks/test-trace-koa.js
matthewloring: Done.

lib/hooks/userspace/hook-koa.js
matthewloring: To my knowledge, `shimmer` does not provide a way to patch a module itself but only properties of it. This could be done custom w/o shimmer if it is really required but it involves messing with the node module cache (or at least that is how I have investigated it in the past).

test/hooks/test-trace-koa.js
matthewloring: We currently do not have this type of testing for any of our frameworks and discussed incorporating it into the e2e tests. The method, url, ip, status_code, agent, hostname, and app/module labels have been verified manually. I can open a separate bug for this across all the frameworks if you'd like.

matthewloring:
@ofrobots Ready for another look (tests passing now)

bin/run-test.sh
```diff
@@ -35,7 +35,7 @@ function run {
     C="$(npm bin)/istanbul cover --dir ./coverage/${counter}"
     ((counter++))
   fi
-  ($C "$(npm bin)/_mocha" -- $* --timeout 4000 --R spec) || exit 1
+  (node --harmony $C "$(npm bin)/_mocha" -- $* --timeout 4000 --R spec) || exit 1
```
ofrobots: -1 on the use of --harmony. Can we skip koa testing on Node 0.12?

lib/hooks/userspace/hook-koa.js
ofrobots: Has this been addressed?

lib/hooks/userspace/hook-koa.js
matthewloring: I've added tests to verify well formed labels on captured spans.

lib/hooks/userspace/hook-koa.js
matthewloring: This is now covered by testing.

bin/run-test.sh
matthewloring: Fixed.

lib/hooks/userspace/hook-koa.js
matthewloring: We now have a test asserting that this value is always defined as a label on captured spans.

matthewloring:
@ofrobots This should be working without the harmony flag. PTAL.

bin/run-test.sh
```diff
@@ -42,7 +42,10 @@ function run {
 run test test/hooks
 for test in test/standalone/test-*.js ;
 do
-  run "${test}"
+  if [[ ! $(node --version) =~ v0\.12\..* || ! "${test}" =~ .*trace\-koa\.js ]]
+  then
+    run "${test}"
+  fi
```
ofrobots: More succinctly (untested) : `[[ $(node --version) =~ v0\.12\..* ]] || run "${test}"`

ofrobots: Ugh.. forget the second half of the comparison, but you get the idea

ofrobots: Given that the second half makes it much longer, I am not sure this is a good idea anymore.

matthewloring: Yea, it is a possibility but pushes the line over 90 characters forcing it to be broken up anyways.

lib/hooks/userspace/hook-koa.js
ofrobots: This is not koa-specific, but we need a issue for the `X-Forwarded-For` problem.

bin/run-test.sh
ofrobots: Feel free to ignore this suggestion.

lib/hooks/userspace/hook-koa.js
```diff
@@ -0,0 +1,131 @@
...
+      res.end = originalEnd;
+      const returned = res.end(chunk, encoding);
+
+      endRootSpanForRequest(rootContext, req, res);
+      return returned;
+    };
+    namespace.bind(next);
+  });
+  yield next;
+}
```
ofrobots: Would a middleware like the response-time middleware that koa docs show work for our use-case?
```
app.use(function *(next){
  startRootSpan...
  yield next;
  endRootSpan...
});
```

matthewloring: The issue is that we would need to wrap the body of that middleware in a `namespace.run` however we would be unable to yield out of the middle of the run function. What we would need is something like:

```javascript
app.use(function *(next){
  const namespace = cls.getNamespace();
  const that = this;
  const rootContext = startRootSpanForRequest(that.req);
  namespace.run(function() {
    yield_from_outer_function next;
    endRootSpanForRequest(rootContext, that.req, that.res);
  });
});
```

However I don't know of an operation providing the `yield_from_outer_function` functionality.

ofrobots:
LGTM as experimental support.

PR #195, opened 2015-12-28
Updated logging in a few places

### LINKED ISSUE #194: Warning that trace-agent is the first module loaded doesn't show up at the default log level
```
WARN :@google/cloud-trace: express tracing might not work as /Users/ofrobots/src/scavenger-hunt-prod/default/Server/node_modules/express/index.js was loaded before the trace agent was initialized.
```

I only got this when I increased the logLevel. This should show up at the default logLevel.

matthewloring:
Detecting modules loaded before the trace agent now causes an error to
be logged (up from warning). Message logged for full trace buffer now
states this as the cause.

Fixes #194


matthewloring:
@ofrobots PTAL.

ofrobots:
LGTM.

PR #197, opened 2015-12-29
Warn about preloaded modules loaded before trace

### LINKED ISSUE #196: Warn about preloaded modules that could interfere with trace
1. Update readme to load trace agent before --require'ing any other modules
2. Warn if any modules were preloaded before agent

matthewloring:
Fixes #196

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,14 @@ This module provides Cloud Trace support for Node.js applications. [Google Cloud
 
 Your project ID is visible in the [Google Cloud Console Console](https://console.cloud.google.com/project?_ga=1.258049870.576536942.1443543237), it may be something like `particular-future-12345`. If your application is [running on Google Cloud Platform](running-on-google-cloud-platform), you don't need to specify the project ID.
 
+## Configuration
+
+See [the default configuration](config.js) for a list of possible configuration options. These options can be passed to the agent through the object argument to the start command shown above:
+
+         require('@google/cloud-trace').start({projectId: 'your-project-id', samplingRate: 500});
+
+Alternatively, you can provide configuration through a config file. This can be useful if you want to load our module using `--require` on the command line instead of editing your main script. You can start by copying the default config file and modifying it to suit your needs. The `GCLOUD_DIAGNOSTICS_CONFIG` environment variable should point to your configuration file.
+
 ## Running on Google Cloud Platform
 
 There are three different services that can host Node.js application to Google Cloud Platform.
--- a/lib/hooks/index.js
+++ b/lib/hooks/index.js
@@ -95,10 +95,7 @@ function findModuleVersion(modulePath, load) {
   return process.version;
 }
 
-function activate(agent) {
-
-  logger = agent.logger;
-
+function checkLoadedModules(logger) {
   for (var moduleName in toInstrument) {
     var regex = new RegExp('node_modules' + path.sep + moduleName + path.sep);
     for (var file in require.cache) {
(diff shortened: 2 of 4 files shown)


matthewloring:
@ofrobots PTAL.

README.md
```diff
@@ -24,10 +24,16 @@ This module provides Cloud Trace support for Node.js applications. [Google Cloud
 
         npm install --save @google/cloud-trace
 
-2. Include the library at the *top of the main script of your application*. It's important that Cloud Trace is the first thing executed so that it can accurately gather data:
+2. Include the library by requiring it as part of your start up command (your project id must be specified using a config file or environment variable, see below):
+
+        node --require @google/cloud-trace app.js
+
+or at the *top of the main script of your application*:
```
ofrobots: I am afraid this is a bit too confusing. Specially the 'see below' part. If the `--require` is our preferred mechanism we should give complete instructions here. If that is too verbose, then `--require` should *not* be our preferred mechanism.

README.md
```diff
@@ -24,10 +24,16 @@ This module provides Cloud Trace support for Node.js applications. [Google Cloud
...
 
         require('@google/cloud-trace').start({projectId: 'your-project-id'});
 
+It's important that Cloud Trace is the first thing executed so that it can accurately gather data. If you require multiple modules as part of your start up command, cloud trace must be the first since they are loaded in order.
+
```
ofrobots: I think it would be better to state this *before* the mechanism to load is described.

README.md
matthewloring: I've taken another pass at this.

README.md
matthewloring: How does this read after the rewrite?

matthewloring:
@ofrobots PTAL.

pmuellr:
As someone familiar with using this type of tooling, wording looks good to me.  I assume the path to the config file can be relative, and if so would be relative to the current working directory (expected behavior). If true, doesn't require any further clarification, but if not, then clarify further.

matthewloring:
@pmuellr Thank you for the suggestion! I agree that config files should be loaded relative to the current working directory. Tracked in #198.

ofrobots:
I would prefer if we added a new section in README.md on the different ways a user can pass configuration to the agent and left the default getting started instructions as they are. That keeps things simpler for new users.

I am not too worried anymore about the problem of the user using having some other code loaded through `--require` before the agent is started because you just added a check to detect that.

matthewloring:
@ofrobots I've reverted the quick start section of the README and added a new section on configuration. I've omitted the specialized environment variables (log level, etc) so that we don't get locked into those names. If you'd like these to be documented as well I can add them.

matthewloring:
@ofrobots Updated. Ready for a look.

README.md
```diff
@@ -30,6 +30,14 @@ This module provides Cloud Trace support for Node.js applications. [Google Cloud
 
 Your project ID is visible in the [Google Cloud Console Console](https://console.cloud.google.com/project?_ga=1.258049870.576536942.1443543237), it may be something like `particular-future-12345`. If your application is [running on Google Cloud Platform](running-on-google-cloud-platform), you don't need to specify the project ID.
 
+## Configuration
+
+See [the default configuration](config.js) for a list of possible configuration options. These options can be passed to the agent through the object argument to the start command shown above:
+
+         require('@google/cloud-trace').start({projectId: 'your-project-id', samplingRate: 500});
+
+Alternatively, you can provide configuration through a config file. This can be useful if you want to load our module using --require on the command line instead of editing your main script. You can start by copying the default config file and modifying it to suit your needs. The `GCLOUD_DIAGNOSTICS_CONFIG` environment variable should point to your configuration file.
```
ofrobots: backticks around `--require`.

lib/hooks/index.js
```diff
@@ -109,6 +106,20 @@ function activate(agent) {
       }
     }
   }
+  if (process._preload_modules && process._preload_modules.length > 0) {
+    var fst = process._preload_modules[0];
```
ofrobots: `fst`: unnecessary abbreviation.

ofrobots:
LGTM w/ comments.

README.md
matthewloring: Done.

lib/hooks/index.js
matthewloring: [Habit](http://caml.inria.fr/pub/docs/manual-ocaml/libref/Pervasives.html).

ofrobots: You mean hbt?

PR #199, opened 2016-01-06
Support relative config file paths from cwd

### LINKED ISSUE #198: Config files are loaded relative to agent installation directory
Config files should be loaded relative to the current working directory as opposed to the agent installation directory

matthewloring:
Fixes #198

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,6 +24,7 @@ var SpanData = require('./lib/span-data.js');
 var common = require('@google/cloud-diagnostics-common');
 var semver = require('semver');
 var constants = require('./lib/constants.js');
+var path = require('path');
 
 /**
  * Phantom implementation of the trace agent. This allows API users to decouple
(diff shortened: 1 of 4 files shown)


matthewloring:
@ofrobots PTAL.

index.js
```diff
@@ -53,7 +54,11 @@ var initConfig = function(projectConfig) {
   util._extend(config, require('./config.js'));
   util._extend(config, projectConfig);
   if (process.env.hasOwnProperty('GCLOUD_TRACE_CONFIG')) {
-    util._extend(config, require(process.env.GCLOUD_TRACE_CONFIG));
+    var config_path = process.env.GCLOUD_TRACE_CONFIG;
+    if (config_path[0] !== path.sep) {
```
pmuellr: Not a biggie, this should work (except for windows drive letter prefixes?), but I prefer to put all the responsibility on someone else, with something like [`path.resolve(process.env.GCLOUD_TRACE_CONFIG)`](https://nodejs.org/dist/latest-v4.x/docs/api/path.html#path_path_resolve_from_to)

ofrobots:
LGTM w/ the suggestion from @pmuellr.

Aside (follow-on issues): 
* We should support `.json` config files in addition the `.js`. I think the existing code ought to work, but it would be good to add a test.
* The config file should have a top level `trace` property. i.e. `util._extend(config, require(config_path))` should probably look like `util._extend(config, require(config_path).trace)`. This would allow users to use a single configuration file for all the agents (trace, debug, profile).

index.js
matthewloring: This is definitely more clean. I've updated to use `path.resolve`.

matthewloring:
I've added the test case for json here since it is already handled correctly. I agree with your second suggestion as well and we can address that in a future PR.

PR #200, opened 2016-01-06
Change config format to support multiple agents

matthewloring:
The environment variable name for the config file has been updated to
the more general GCLOUD_CONFIG. The config file is also now structured
to support separate configs for each of the profile, debug, and trace
agents.

### DIFF
--- a/index.js
+++ b/index.js
@@ -53,8 +53,11 @@ var initConfig = function(projectConfig) {
   var config = {};
   util._extend(config, require('./config.js'));
   util._extend(config, projectConfig);
-  if (process.env.hasOwnProperty('GCLOUD_TRACE_CONFIG')) {
-    util._extend(config, require(path.resolve(process.env.GCLOUD_TRACE_CONFIG)));
+  if (process.env.hasOwnProperty('GCLOUD_DIAGNOSTICS_CONFIG')) {
+    var c = require(path.resolve(process.env.GCLOUD_DIAGNOSTICS_CONFIG));
+    if (c) {
+      util._extend(config, c.trace);
+    }
   }
   if (process.env.hasOwnProperty('GCLOUD_TRACE_LOGLEVEL')) {
     config.logLevel = process.env.GCLOUD_TRACE_LOGLEVEL;
--- a/test/standalone/test-config-json.js
+++ b/test/standalone/test-config-json.js
@@ -23,7 +23,8 @@ var assert = require('assert');
 
 // Fixtures configuration:
 // { logLevel: 4, stackTraceLimit: 1 };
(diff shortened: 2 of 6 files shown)


matthewloring:
@ofrobots PTAL.

index.js
```diff
@@ -53,8 +53,9 @@ var initConfig = function(projectConfig) {
   var config = {};
   util._extend(config, require('./config.js'));
   util._extend(config, projectConfig);
-  if (process.env.hasOwnProperty('GCLOUD_TRACE_CONFIG')) {
-    util._extend(config, require(path.resolve(process.env.GCLOUD_TRACE_CONFIG)));
+  if (process.env.hasOwnProperty('GCLOUD_CONFIG')) {
+    util._extend(config,
+      require(path.resolve(process.env.GCLOUD_CONFIG)).trace);
```
ofrobots: Need to check that `require(..)` returns an object that can be dereferenced.

test/standalone/test-config-json.js
```diff
@@ -23,7 +23,7 @@ var assert = require('assert');
 
 // Fixtures configuration:
 // { logLevel: 4, stackTraceLimit: 1 };
-process.env.GCLOUD_TRACE_CONFIG = path.join('test', 'fixtures', 'test-config.json');
+process.env.GCLOUD_CONFIG = path.join('test', 'fixtures', 'test-config.json');
```
ofrobots: `GCLOUD_CONFIG` is too generic a name. Could be confused with the gcloud cli, or with the `gcloud` npm module.

index.js
matthewloring: What exactly do you mean by this? Do you want the contents of the config to be validated to only contain expected properties? If the result of the `require` is a `number` or `function` or `undefined`, this shouldn't break.

test/standalone/test-config-json.js
matthewloring: How about `GCLOUD_DIAGNOSTICS_CONFIG`?

index.js
ofrobots: This breaks (i.e. throws) when `require` returns a value that is not deference-able, e.g. `null` or `undefined`. You should check that `require` result is truthy.

test/standalone/test-config-json.js
ofrobots: sgtm.

index.js
matthewloring: Fixed.

test/standalone/test-config-json.js
matthewloring: Updated.

matthewloring:
@ofrobots PTAL

ofrobots:
LGTM.

PR #201, opened 2016-01-11
Warn about modules loaded before trace agent

matthewloring:
Add additional logging to report any modules loaded before the trace
agent even if they are not modules that we patch directly.

Fixes #187.

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,6 +16,8 @@
 
 'use strict';
 
+var filesLoadedBeforeTrace = Object.keys(require.cache);
+
 // Load continuation-local-storage first to ensure the core async APIs get
 // patched before any user-land modules get loaded.
 require('continuation-local-storage');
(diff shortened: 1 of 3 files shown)


matthewloring:
@ofrobots PTAL.

index.js
```diff
@@ -121,6 +132,12 @@ var publicAgent = {
     var headers = {};
     headers[constants.TRACE_AGENT_REQUEST_HEADER] = 1;
 
+    if (modulesLoadedBeforeTrace.length > 0) {
```
ofrobots: Does this check need to be done during `start`? Could it be done earlier?

matthewloring: We don't have the ability to log before this point. We precompute as much of the information as possible before `start` but need to wait until a logger is initialized to warn.

ofrobots:
Needs a test.

index.js
```diff
@@ -16,6 +16,17 @@
 
 'use strict';
 
+var filesLoadedBeforeTrace = Object.keys(require.cache);
+var modulesLoadedBeforeTrace = [];
+for (var i = 0; i < filesLoadedBeforeTrace.length; i++) {
+  var path = filesLoadedBeforeTrace[i];
+  var matches = /.*?node_modules\/([^\/]*).*/.exec(path);
```
ofrobots: The regular expression uses platform specific slashes

matthewloring: Replaced with `path.sep`.

matthewloring:
@ofrobots Test is added.

ofrobots:
LGTM.

PR #204, opened 2016-01-14
Handle remaining todos in code

matthewloring:


matthewloring:
@ofrobots PTAL.

lib/trace-agent.js
```diff
@@ -234,7 +236,11 @@ TraceAgent.prototype.addContextToHeaders = function(spanData, headers) {
 module.exports = {
   get: function(config, logger) {
     if (traceAgent) {
-      // TODO: log error if config object is different from traceAgent.config
+      if (!_.isEqual(config, traceAgent.config_)) {
+        traceAgent.logger.warn('New configuration does not match configuration' +
+          'of existing agent:\nNew: ' + JSON.stringify(config) + '\nExisting: ' +
+          JSON.stringify(traceAgent.config_));
```
ofrobots: The message needs to inform the user that their old config is still in effect.

package.json
```diff
@@ -50,6 +50,7 @@
   "dependencies": {
     "@google/cloud-diagnostics-common": "^0.2.1",
     "continuation-local-storage": "^3.1.4",
+    "lodash": "^4.0.0",
```
ofrobots: Do you want to use `lodash.isequal` instead?

ofrobots:
lgtm w/ comments.

lib/trace-agent.js

package.json
matthewloring: Didn't realize this was out there. Definitely a better fit.

PR #213, opened 2016-03-15
Update docs to configure credentials with gcloud

matthewloring:


matthewloring:

JustinBeckwith:
LGTM

ofrobots:
What about the scenario when the user is running on GCE, but their VMs don't have the `cloud-platform` scope. Our section (above) tells them to following the instructions here – which won't work unless they also install gcloud on their VMs.

PR #214, opened 2016-03-15
Logging improvements for first error logging

matthewloring:


matthewloring:
@ofrobots PTAL.

lib/hooks/core/hook-http.js
```diff
@@ -43,19 +43,19 @@ function requestWrap(request) {
 
     var root = cls.getRootContext();
 
+    options = (typeof options === 'string') ?
+      url.parse(options) : clone(options);
+    options.headers = options.headers || {};
+    var uri = uriFromOptions(options);
```
ofrobots: We are increasing the path length for the case where `root` is the `nullSpan` (common case) for the benefit of logging. With some refactoring you should be able to avoid this. You might want to outline this into a separate function that you can call from within the `if (!root) {` test.

matthewloring: The logic required to get the Uri is quite small. I've added it inside the `if (!root) {` case.

lib/hooks/core/hook-http.js
```diff
@@ -44,7 +44,10 @@ function requestWrap(request) {
     var root = cls.getRootContext();
 
     if (!root) {
-      agent.logger.debug('Untraced http options:', options);
+      if (typeof options === 'string') {
+        options = url.parse(options);
```
ofrobots: Very very minor nit: get rid of this assignment and conditional, and inline the whole expression into the call to `uriFromOptions`.

This would allow the optimizer to avoid having to merge type-information about `option` from a rarely (or never taken) path with the mainline path.

ofrobots:
LGTM w/ nit.

lib/hooks/core/hook-http.js
ofrobots: nevermind. There is a return following this statement, so there is no merge with the mainline path. This code is good as it is now.

PR #216, opened 2016-03-16
Treat spanIds as strings to avoid integer overflow

matthewloring:


matthewloring:
@ofrobots PTAL.

ofrobots:
LGTM if e2e tests pass (i.e. we have verified that the trace API accepts these).

PR #217, opened 2016-03-16
Respect incoming trace enabled options

matthewloring:
Previously, if the app server decided to ignore a request that our sampling process decided to trace, the span would not be rendered in the UI because the parent span to the root reported by the trace agent was not present. Inversely, if the app server decided to trace a request that our sampling process decided to ignore, an empty root span would show up in the UI without any additional Node.js information.

To remedy these issues, we have adopted the following semantics:

If there is no trace context header; we use our sampling
If there is a trace context header, but no options or non-numeric options; we use our sampling
If there is a trace context header with options, and options is a number; we use that to decide.

### DIFF
--- a/lib/hooks/userspace/hook-express.js
+++ b/lib/hooks/userspace/hook-express.js
@@ -42,7 +42,9 @@ function middleware(req, res, next) {
     agent.logger.info('Express: no namespace found, ignoring request');
     return next();
   }
-  if (!agent.shouldTrace(req.originalUrl)) {
+  var traceHeader = agent.parseContextFromHeader(
+    req.get(constants.TRACE_CONTEXT_HEADER_NAME)) || {};
+  if (!agent.shouldTrace(req.originalUrl, traceHeader.options)) {
     return next();
   }
 
@@ -52,7 +54,7 @@ function middleware(req, res, next) {
   var originalEnd = res.end;
 
   namespace.run(function() {
-    var rootContext = startRootSpanForRequest(req);
+    var rootContext = startRootSpanForRequest(req, traceHeader);
 
     // wrap end
     res.end = function(chunk, encoding) {
@@ -70,14 +72,12 @@ function middleware(req, res, next) {
 /**
  * Creates and sets up a new root span for the given request.
  * @param {Object} req The request being processed.
+ * @param {Object} traceHeader The incoming trace header.
  * @returns {!SpanData} The new initialized trace span data instance.
  */
-function startRootSpanForRequest(req) {
-  var result = agent.parseContextFromHeader(
-    req.get(constants.TRACE_CONTEXT_HEADER_NAME)) || {};
-
-  var traceId = result.traceId;
-  var parentSpanId = result.spanId;
+function startRootSpanForRequest(req, traceHeader) {
+  var traceId = traceHeader.traceId;
+  var parentSpanId = traceHeader.spanId;
   var url = req.protocol + '://' + req.hostname + req.originalUrl;
 
   // we use the path part of the url as the span name and add the full
--- a/lib/hooks/userspace/hook-hapi.js
+++ b/lib/hooks/userspace/hook-hapi.js
@@ -41,7 +41,9 @@ function middleware(request, reply) {
   }
   var req = request.raw.req;
   var res = request.raw.res;
-  if (!agent.shouldTrace(req.url)) {
+  var traceHeader = agent.parseContextFromHeader(
+    req.headers[constants.TRACE_CONTEXT_HEADER_NAME]) || {};
+  if (!agent.shouldTrace(req.url, traceHeader.options)) {
     return reply.continue();
   }
 
@@ -51,7 +53,7 @@ function middleware(request, reply) {
   var originalEnd = res.end;
 
   namespace.run(function() {
-    var rootContext = startRootSpanForRequest(req);
+    var rootContext = startRootSpanForRequest(req, traceHeader);
 
     // wrap end
     res.end = function(chunk, encoding) {
@@ -69,14 +71,12 @@ function middleware(request, reply) {
 /**
  * Creates and sets up a new root span for the given request.
  * @param {Object} req The request being processed.
+ * @param {Object} traceHeader The incoming trace header.
  * @returns {!SpanData} The new initialized trace span data instance.
  */
-function startRootSpanForRequest(req) {
-  var result = agent.parseContextFromHeader(
-    req.headers[constants.TRACE_CONTEXT_HEADER_NAME]) || {};
-
-  var traceId = result.traceId;
-  var parentSpanId = result.spanId;
+function startRootSpanForRequest(req, traceHeader) {
+  var traceId = traceHeader.traceId;
+  var parentSpanId = traceHeader.spanId;
   var url = (req.headers['X-Forwarded-Proto'] || 'http') +
     '://' + req.headers.host + req.url;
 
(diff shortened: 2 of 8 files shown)


matthewloring:
@ofrobots PTAL.

lib/trace-agent.js
```diff
@@ -209,7 +212,7 @@ TraceAgent.prototype.parseContextFromHeader = function(str) {
   return {
     traceId: matches[1],
     spanId: Number(matches[2]),
-    options: matches[3]
+    options: Number(matches[3])
```
ofrobots: Is there any public documentation (e.g. protobuf) that you can link to that describes the schema?

lib/trace-agent.js
```diff
@@ -163,8 +163,11 @@ TraceAgent.prototype.addTransactionLabel = function(key, value) {
  * Determines whether a trace of the given name should be recorded based
  * on the current tracing policy.
  */
-TraceAgent.prototype.shouldTrace = function(name) {
-  return this.policy.shouldTrace(Date.now(), name);
+TraceAgent.prototype.shouldTrace = function(name, options) {
```
ofrobots: Add jsdoc for the types.

lib/trace-agent.js
```diff
@@ -189,8 +192,8 @@ TraceAgent.prototype.isTraceAgentRequest = function(options) {
 
 /**
  * Parse a cookie-style header string to extract traceId, spandId and options
- * ex: '123456/667;o=something'
- * -> {traceId: '123456', spanId: '667', options: 'something'}
+ * ex: '123456/667;o=3'
+ * -> {traceId: '123456', spanId: '667', options: '3'}
```
ofrobots: Add jsdoc for the types of the properties of the returned object.

test/standalone/test-trace-options.js
```diff
@@ -0,0 +1,62 @@
...
+    });
+    var server = app.listen(common.serverPort, function() {
+      var doneCount = 0;
+      [1,3,5,7].forEach(function(option) {
+        var headers = {};
+        headers['x-cloud-trace-context'] = '42/1729;o=' + option;
+        http.get({port: common.serverPort, headers: headers}, function(res) {
+          res.on('data', function() {});
+          res.on('end', function() {
+            if (++doneCount === 4) {
```
ofrobots: Avoid using hard-coded 4. Save the array in a variable and use length instead.

test/standalone/test-trace-options.js
```diff
@@ -0,0 +1,62 @@
...
+        assert(!err, 'Skipping: no mongo server found at localhost:27017.');
+        mongoose.connection.close(function(err) {
+          assert(!err);
+          res.sendStatus(200);
+        });
+      });
+    });
+    var server = app.listen(common.serverPort, function() {
+      var doneCount = 0;
+      [1,3,5,7].forEach(function(option) {
```
ofrobots: It would be worth testing some cases where the `enabled` bit is off.

lib/trace-agent.js
```diff
@@ -163,8 +163,11 @@ TraceAgent.prototype.addTransactionLabel = function(key, value) {
  * Determines whether a trace of the given name should be recorded based
  * on the current tracing policy.
  */
-TraceAgent.prototype.shouldTrace = function(name) {
-  return this.policy.shouldTrace(Date.now(), name);
+TraceAgent.prototype.shouldTrace = function(name, options) {
+  if (isNaN(options)) {
```
ofrobots: Okay, so reading this my understanding of the semantics of whether we should trace or not are:

* If there is no trace context header; we use our sampling
* If there is a trace context header, but no options; we use our sampling
* If there is a trace context header with options, and options is a number; we use that to decide.

Can you edit the PR abstract to include the rationale (why we need to change this) and summarize the new semantics?

lib/trace-agent.js
matthewloring: Unfortunately, no.

lib/trace-agent.js
matthewloring: done.

lib/trace-agent.js

test/standalone/test-trace-options.js
matthewloring: Done.

test/standalone/test-trace-options.js
matthewloring: Done.

lib/trace-agent.js
matthewloring: done.

matthewloring:

lib/hooks/userspace/hook-express.js
```diff
@@ -70,14 +72,12 @@ function middleware(req, res, next) {
 /**
  * Creates and sets up a new root span for the given request.
  * @param {Object} req The request being processed.
+ * @param {Object} traceHeader The incoming trace header.
```
ofrobots: I think this ought to be `!Object` for traceHeader.

lib/hooks/userspace/hook-hapi.js
```diff
@@ -71,12 +73,9 @@ function middleware(request, reply) {
  * @param {Object} req The request being processed.
```
ofrobots: Similar to express, add type annotations for traceHeader.

lib/trace-agent.js
```diff
@@ -162,9 +162,15 @@ TraceAgent.prototype.addTransactionLabel = function(key, value) {
 /**
  * Determines whether a trace of the given name should be recorded based
  * on the current tracing policy.
+ *
+ * @param {string} name the url to trace
+ * @param {number} options the trace header options
```
ofrobots: I believe options can be `number|undefined`.

lib/hooks/userspace/hook-express.js
matthewloring: traceHeader should always be defined at this point. See https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/217/files#diff-aa20dec8ea838bb18b5bb09bd5b9fce8R45

test/standalone/test-trace-options.js
```diff
@@ -0,0 +1,68 @@
...
+    var app = express();
+    app.get('/', function (req, res) {
+      setTimeout(function() {
+        res.send('Hello World');
+      }, 50);
+    });
+    var server = app.listen(common.serverPort, function() {
+      var shouldTraceOptions = [1,3,5,7];
+      var shouldNotTraceOptions = [0,2,4,6];
+      sendRequests(shouldTraceOptions, shouldTraceOptions.length, function() {
```
ofrobots: Nit: It would be simpler to concatenate the two arrays and make a single call to `sendRequests`. Or, better, keep it inline.

lib/trace-agent.js
matthewloring: If the trace header options were `undefined` when parsed, they should be `NaN` at this point.

test/standalone/test-trace-options.js
matthewloring: We need validation that the traced headers correspond precisely to `[1,3,5,7]` and the same for the untraced headers. Combining the lists would require some other channel to surface these header results. I think this out ways the simplicity gained by concatenating these lists.

ofrobots:
LGTM with minor nits.

As a follow-on, this needs documentation updates about the sampling behaviour.

lib/hooks/userspace/hook-hapi.js
matthewloring: Done.

lib/trace-agent.js
ofrobots: You're right.

test/standalone/test-trace-options.js
ofrobots: SGTM.

lib/trace-agent.js
ofrobots: Actually, here's the path on which it is undefined: when parseContextFromHeader returns null, we assign `traceHeader` to `{}`.

PR #219, opened 2016-03-17
README clarification

matthewloring:
Specify that trace must start before requiring other modules.


matthewloring:
@ofrobots PTAL.

ofrobots:
LGTM.

PR #220, opened 2016-03-18
Support for hapi 9 + 13

matthewloring:


matthewloring:

JustinBeckwith:
This is great, but I'm actually on version 13 o.0

matthewloring:
Our unit tests are now passing for versions 9-13. I need to dig through the release notes for each of these versions just to ensure that there aren't new routing/handling mechanisms that need to be traced but the functionality that was in 8 is now preserved across 8-13.

test/hooks/test-trace-hapi.js
```diff
@@ -122,6 +123,9 @@ Object.keys(versions).forEach(function(version) {
     });
 
     it('should accurately measure get time, after + get', function(done) {
+      if (version.substring(4) > 10) {
```
ofrobots: What's the rationale for early done on hapi10?

ofrobots: er.. > 10.

matthewloring: https://github.com/hapijs/hapi/issues/2850 The `after` method was removed in 11.

ofrobots: Please add a comment here with the rationale.

matthewloring: done.

ofrobots:
LGTM.

Issue #221, opened 2016-03-23
publish new version


AdriVanHoudt:
I want to try this with hapi 13 but it seems like the module has not been published in a while.
Is this for a reason and if so when can I expect a new release? I really want to try this.

Also what is the command to log debug info to console (like `DEBUG=cloud-trace node server`)?

matthewloring:
You can get debug info by running `GCLOUD_TRACE_LOGLEVEL=4 node server.js`. We will be doing a release soon but you can try out hapi 13 functionality by installing from master: `npm install GoogleCloudPlatform/cloud-trace-nodejs`.

AdriVanHoudt:
Cool thanks!

AdriVanHoudt:
Ok did some testing on dev and looks nice!
Will enable on staging env for more testing.
Also Mysql did not seem to be logged due to no root. Logs:  `Untraced mysql query:` (I don't know what it means but just letting you know :P)
I am using https://www.npmjs.com/package/hapi-plugin-mysql for my sql connection, not sure if related.
Also I am using ioredis, is this package supported?

matthewloring:
Thank you for the log information. #222 should get mysql reporting working with your plugin. We do not currently have plans to support ioredis but do have support for [node redis](https://www.npmjs.com/package/redis).

AdriVanHoudt:
Very cool! Also awesome that you used a module I wrote to test :O 
Is there a reason you chose node_redis instead of ioredis? (they have plan to merge but no idea on the timeline on that)

matthewloring:
We went with node_redis due to its popularity but we hope to increase the number of frameworks we can support over time.

AdriVanHoudt:
Redis usage is not significant for atm so fine maybe I can do a shot at a PR after my vacation

AdriVanHoudt:
Thanks for the help and the great work!

matthewloring:
Thanks for trying it out. I'm glad the information is useful!

PR #222, opened 2016-03-24
Patch mysql connection pool

matthewloring:
If the mysql connection pool is accessed during a request we drop
context. By patching getConnection we can avoid this.

Fixes the mysql component of #221

### DIFF
--- a/test/standalone/test-mysql-pool.js
+++ b/test/standalone/test-mysql-pool.js
@@ -1,0 +1,67 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


matthewloring:
@ofrobots PTAL.

ofrobots:
How did you find this context loss?

test/standalone/test-mysql-pool.js
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ */
+'use strict';
+
+var common = require('../hooks/common.js');
+var assert = require('assert');
+var http = require('http');
+var semver = require('semver');
+
+// hapi 13 and hapi-plugin-mysql uses const
```
ofrobots: perhaps we should consider starting to use https://npm.im/feature-detect-es6.

matthewloring:
#221

ofrobots:
It would be good to include a 'Fixes' tag to the commit and PR.

test/standalone/test-mysql-pool.js
matthewloring: My understanding is that we would still need to manually audit each package for es6 use. We already know the features available on each of the versions we test against so I'm not sure what this module provides.

ofrobots: It is nicer to feature test rather than look for specific version numbers. The capability that we are really looking for is whether the version of node we are running on supports `const`.

This more of an observation at this point. If we start doing this in other places, it would be worth changing at that point.

ofrobots:
Looking at the source for mysql, there are indeed two calls to `process.domain.bind` that correspond to these two places we now patch mysql.

LGTM if the tests pass.

PR #227, opened 2016-04-01
Experimental support for grpc tracing

### LINKED ISSUE #226: Context is dropped when using gRPC

matthewloring:
Fixes #226

### DIFF
--- a/lib/hooks/userspace/hook-grpc.js
+++ b/lib/hooks/userspace/hook-grpc.js
@@ -1,0 +1,50 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+var cls = require('../../cls.js');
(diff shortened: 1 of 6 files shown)


matthewloring:
@ofrobots PTAL.

lib/hooks/userspace/hook-grpc.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+'use strict';
+
+var cls = require('../../cls.js');
+var shimmer = require('shimmer');
+var semver = require('semver');
+var agent;
+
+var SUPPORTED_VERSIONS = '^0.13.x';
```
ofrobots: Want to double check that we intend claim forward compatibility? We don't use `^` with most other libraries we support.

matthewloring: It is going to get broken sooner but unfortunately I think `0.13.x` is safer.

test/standalone/test-grpc-context.js
```diff
@@ -0,0 +1,66 @@
...
+      });
+    });
+    var server = app.listen(common.serverPort, function() {
+      var grpcServer = new grpc.Server();
+      grpcServer.addProtoService(test_proto.Tester.service, {
+        test: function(call, cb) {
+          cb(null, {message: 'world'});
+        }
+      });
+      grpcServer.bind('0.0.0.0:50051', grpc.ServerCredentials.createInsecure());
```
ofrobots: Bind to localhost only please. You don't want to open that port publicly even for the duration of the test.

ofrobots: nit: avoid hard-coded port number.

ofrobots:
LGTM w/ comment, nit, and if your have verified that the test loses context without the grpc patching.

test/standalone/test-grpc-context.js
matthewloring: Bound to localhost. We need to specify a port for grpc but I've moved it to a constant above.

matthewloring:
Verified without patch.

PR #228, opened 2016-04-06
Update supported framework versions in readme

matthewloring:


matthewloring:
@ofrobots PTAL.

README.md
```diff
@@ -87,14 +87,14 @@ Note: When you open on up the traces view under the monitoring header, you may s
...
-* hapi v8 (experimental)
+* express version 4
+* hapi versions 8 - 13
+* restify versions 3 - 4 (experimental)
 
 The agent will also automatic trace of the following kinds of RPCs:
 * Outbound HTTP requests
 * MongoDB v2+
-* Redis v0.12+ (experimental)
+* Redis versions 0.12 - 2 (experimental)
```
AdriVanHoudt: shouldn't this also include mysql?

matthewloring: Added, thanks for catching that!

AdriVanHoudt:
for the rpc's I would add what modules in specific are supported (and modules that use those packages like hapi-plugin-mysql), so people like me who happen to use redisio instead of node-redis don't get confused about why it is not logging

matthewloring:
@AdriVanHoudt I've updated the rpc section to link to the specific packages supported. Though things worked out for hapi-plugin-mysql, we can't guarantee generally that packages depending on supported rpc frameworks will not break our instrumentation. It is better to handle these dependent packages on a case by case basis.

ofrobots:
lgtm.

PR #233, opened 2016-04-11
Attempt to publish traces on unhandled exception

### LINKED ISSUE #232: Agent should attempt to publish traces on unhandled exception

matthewloring:
Fixes #232

### DIFF
--- a/config.js
+++ b/config.js
@@ -60,6 +60,13 @@ module.exports = {
 
     // The number of transactions we buffer before we publish to the trace
     // API, unless we hit `flushDelaySeconds` first.
-    bufferSize: 1000
+    bufferSize: 1000,
+
+    // Specifies the behavior of the trace agent in the case of an uncaught exception.
+    // Possible values are:
+    //   `ignore`: Take no action.
+    //   `flush`: Attempt to publish traces silence the exception.
+    //   `flushAndExit`: Attempt to publish traces and exit the process.
+    onUncaughtException: 'ignore'
   }
 };
--- a/index.js
+++ b/index.js
@@ -39,6 +39,8 @@ for (var i = 0; i < filesLoadedBeforeTrace.length; i++) {
   }
 }
 
+var onUncaughtExceptionValues = ['ignore', 'flush', 'flushAndExit'];
+
 /**
  * Phantom implementation of the trace agent. This allows API users to decouple
  * the enable/disable logic from the calls to the tracing API. The phantom API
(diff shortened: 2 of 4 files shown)


matthewloring:
@ofrobots PTAL.

test/standalone/test-trace-uncaught-exception.js
```diff
@@ -0,0 +1,79 @@
...
+var cls = require('../../lib/cls.js');
+var agent = require('../..');
+var request = require('request');
+
+nock.disableNetConnect();
+
+var uri = 'https://cloudtrace.googleapis.com';
+var path = '/v1/projects/0/traces';
+
+process.env.GCLOUD_PROJECT_NUM = 0;
```
ofrobots: Probably need to update this.

matthewloring: Done.

config.js
```diff
@@ -60,6 +60,18 @@ module.exports = {
...
     // The number of transactions we buffer before we publish to the Cloud Trace
     // API, unless we hit `flushDelaySeconds` first.
-    bufferSize: 1000
+    bufferSize: 1000,
+
+    // If true, the trace agent will attempt to publish any buffered traces if an
+    // uncaught exception occurs. This will result in the error being silenced. If
+    // you want your application to exit on uncaught exception you will have to
+    // add your own listener to the 'uncaughtException' event on the process.
+    publishOnUncaught: false,
```
ofrobots: Another idea would be to expose a single option `onUncaughtExcpetion` that has the following behaviours:
* falsy: we don't deal with uncaught exceptions
* 'flush': flush the buffer, and silence the error
* 'flushAndExit': flushes then exits.

WDYT?

matthewloring: I am not personally a fan of using strings as enums in configuration. If we want to reduce the configuration surface area, I would prefer offering a single `flushAndExit` boolean. If people are going to be recovering and silencing the errors then they would likely not care that we don't flush when the process would otherwise have gone down. Would this be acceptable?

matthewloring: Going with ternary enum approach.

index.js
```diff
@@ -133,6 +135,12 @@ var publicAgent = {
       logger.info('Locally provided ProjectId: ' + config.projectId);
     }
 
+    if (onUncaughtExceptionValues.indexOf(config.onUncaughtException) === -1) {
+      logger.error('The value of onUncaughtException should be one of ',
+        onUncaughtExceptionValues);
+      process.exit(1);
```
ofrobots: nit: throw instead of exiting.

config.js
```diff
@@ -60,6 +60,13 @@ module.exports = {
 
     // The number of transactions we buffer before we publish to the trace
     // API, unless we hit `flushDelaySeconds` first.
-    bufferSize: 1000
+    bufferSize: 1000,
+
+    // Specifies the behavior of the trace agent in the case of an uncaught exception.
+    // Possible values are:
+    //   `disregard`: Take no action.
```
ofrobots: 🚲🏡  time!

Can we use `'ignore'` instead of `'disregard'`?

ofrobots:
LGTM w/ nits.

config.js

index.js
matthewloring: Done.

PR #240, opened 2016-04-13
Remove trace span functions from stack traces

### LINKED ISSUE #229: Remove trace agent specific frames from reported stack traces
Right now the top ~4 frames are span logic. We should eliminate these.

matthewloring:
Fixes #229

### DIFF
--- a/lib/hooks/userspace/hook-express.js
+++ b/lib/hooks/userspace/hook-express.js
@@ -83,7 +83,7 @@ function startRootSpanForRequest(req, traceHeader) {
   // we use the path part of the url as the span name and add the full
   // url as a label
   var rootContext = agent.createRootSpanData(req.originalUrl, traceId,
-    parentSpanId);
+    parentSpanId, 3);
   rootContext.addLabel(TraceLabels.HTTP_METHOD_LABEL_KEY, req.method);
   rootContext.addLabel(TraceLabels.HTTP_URL_LABEL_KEY, url);
   rootContext.addLabel(TraceLabels.HTTP_SOURCE_IP, req.connection.remoteAddress);
--- a/lib/span-data.js
+++ b/lib/span-data.js
@@ -28,13 +28,13 @@ var uid = 1;
  * @param {string} name The name of the span.
  * @param {number} parentSpanId The id of the parent span, 0 for root spans.
  * @param {boolean} isRoot Whether this is a root span.
- * @param {number=} time The time when the associated TraceSpan begins.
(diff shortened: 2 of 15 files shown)


matthewloring:
@ofrobots PTAL.

lib/hooks/userspace/hook-express.js
```diff
@@ -83,7 +83,7 @@ function startRootSpanForRequest(req, traceHeader) {
   // we use the path part of the url as the span name and add the full
   // url as a label
   var rootContext = agent.createRootSpanData(req.originalUrl, traceId,
-    parentSpanId);
+    parentSpanId, 3); // Skip 3 frames
```
ofrobots: This mechanism will be brittle when we change something in how `createRootSpanData`. `createRootSpanData` should have an idea how how many frames it should skip regardless, and it could accept an argument that is the number of frames to skip *in addition* so that the caller of `createRootSpanData` can skip additional frames.

It would be good to reduce the unnecessary tight coupling here.

matthewloring: As discussed in person, this philosophy is applied inside the trace agent but doesn't work here in the hooks.

lib/span-data.js
```diff
@@ -28,10 +28,11 @@ var uid = 1;
  * @param {string} name The name of the span.
  * @param {number} parentSpanId The id of the parent span, 0 for root spans.
  * @param {boolean} isRoot Whether this is a root span.
+ * @param {number} skipFrames the number of frames to remove from the top of the stack.
  * @param {number=} time The time when the associated TraceSpan begins.
```
ofrobots: Aside: do we ever call this constructor with the `time` argument in the codebase? Perhaps it should be dropped (in a follow-on)

test/hooks/test-trace-restify.js
```diff
@@ -26,7 +26,7 @@ var versions = {
 
 var server;
 
-Object.keys(versions).forEach(function(version) {
+Object.keys(versions).forEach(function(version, i) {
```
ofrobots: Do you use the index anywhere? It is hard to search for `i` – I can't find it by eyeballing the code.

ofrobots:
lgtm w/ nits.

lib/span-data.js
matthewloring: Good catch. Wiping it out here since it is never used.

test/hooks/test-trace-restify.js
matthewloring: Nope, was used for testing but not needed anymore.

PR #241, opened 2016-04-14
Enforce sampling policy regardless of headers

### LINKED ISSUE #234: Cap trace sampling using configuration regardless of context headers

matthewloring:
This will lead to some empty root spans appearing in the UI if the app
server decides to trace an incoming request but the agent decides not to
sample it. This is deemed preferable to always accepting incoming trace
headers as we cannot always trust their origin.

Fixes #234


matthewloring:
@ofrobots PTAL.

ofrobots:
`shouldTrace` needs tests. Otherwise LGTM.

matthewloring:
`shouldTrace` is already generally tested by `test-trace-policy.js`. I've added additional testing for the case where we sample over top of incoming trace headers.

PR #243, opened 2016-04-18
Fix exception handler config to use ignore

matthewloring:


matthewloring:
@ofrobots PTAL.

ofrobots:
LGTM once tests pass.

PR #249, opened 2016-04-25
Document service account keys

### LINKED ISSUE #248: Full API access seems excessive
I have a cluster that holds data that falls under HIPAA.  That means a lot of security that a normal app does not get.  The cluster I am using only has open ports to the internet on the boxes with Node.js on them, our app servers.  If I give full API access I would also be giving full cluster control to those boxes, which seems really bad.  One of those boxes is compromised and my cluster is compromised.

My request is two fold:

1. Is there a way we can reduce the API requirements?
2. If the module does not need full access, can we accurately document what access it does need to function correctly?

matthewloring:
Add documentation to the README for the use of service account keys for
authorization as an alternative to logging in with the gcloud cli.

Fixes #248

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,9 +73,13 @@ If your application is running outside of Google Cloud Platform, such as locally
 
         GCLOUD_PROJECT=particular-future-12345 node myapp.js
 
-2. You will need to provide service account credentials to your application. The recommended way is via [Application Default Credentials](https://developers.google.com/identity/protocols/application-default-credentials). These can be provisioned by executing the following command:
+2. You need to provide service account credentials to your application. The recommended way is via [Application Default Credentials][app-default-credentials].
 
-        gcloud beta auth application-default login
+  1. [Create a new JSON service account key][service-account].
+  2. Copy the key somewhere your application can access it. Be sure not to expose the key publicly.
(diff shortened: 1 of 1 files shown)


matthewloring:

JustinBeckwith:
LGTM

README.md
```diff
@@ -73,9 +73,13 @@ If your application is running outside of Google Cloud Platform, such as locally
...
 
-2. You will need to provide service account credentials to your application. The recommended way is via [Application Default Credentials](https://developers.google.com/identity/protocols/application-default-credentials). These can be provisioned by executing the following command:
+2. You need to provide service account credentials to your application. The recommended way is via [Application Default Credentials][app-default-credentials].
 
-        gcloud beta auth application-default login
+  1. [Create a new JSON service account key][service-account].
+  2. Copy the key somewhere your application can access it. Be sure not to expose the key publicly.
+  3. Set the environment variable `GOOGLE_APPLICATION_CREDENTIALS` to the full path to the key. The debug agent will automatically look for this environment variable.
+
+3. Alternatively, if you are running your application on a machine where your are using the [`gcloud` command line tools][gcloud-sdk], and are logged using `gcloud auth login`, you already have sufficient credentials, and a service account key is not required.
```
ofrobots: Should this be `gcloud beta auth application-default login` instead of `gcloud auth login`?

matthewloring: Done.

ofrobots:
LGTM.

PR #251, opened 2016-05-02
Fix incorrect modulesLoadedBeforeTrace check

Salakar:
The `modulesLoadedBeforeTrace` check is incorrect, should be greater than 1 as requiring this module itself counts as a "moduleLoadedBeforeTrace", so will always log warnings.

Also fixed the double spacing on the logging.

**Logs:**
```
WARN :@google/cloud-trace: Tracing might not work as the following modules  were loaded before the trace agent was initialized: ["@google"]
```

The regex is also borked on this for extracting package names in orgs, but that can be another PR. 👍

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,10 +34,9 @@ var moduleRegex =
 for (var i = 0; i < filesLoadedBeforeTrace.length; i++) {
   var matches = moduleRegex.exec(filesLoadedBeforeTrace[i]);
   if (matches && matches.length > 1 &&
+      matches[1] !== '@google' &&
       modulesLoadedBeforeTrace.indexOf(matches[1]) === -1) {
-    if (matches[1] !== '@google') {
-      modulesLoadedBeforeTrace.push(matches[1]);
-    }
+    modulesLoadedBeforeTrace.push(matches[1]);
   }
 }
 
--- a/index.js
+++ b/index.js
@@ -29,12 +29,28 @@ var constants = require('./lib/constants.js');
 var path = require('path');
 
 var modulesLoadedBeforeTrace = [];
-var moduleRegex =
-  new RegExp('.*?node_modules\\' + path.sep + '([^\\' + path.sep + ']*).*');
+
+// Includes support for npm '@org/name' packages
+// Regex: .*?node_modules\/(@[^\/]*\/[^\/]*|[^\/]*).*
+// Tests: https://regex101.com/r/lW2bE3/4
+var moduleRegex = new RegExp(
+  '.*?node_modules'
+  + path.sep +
+  '(@[^'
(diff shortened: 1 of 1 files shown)


Salakar:
@googlebot CLA signed

Salakar:
Not sure why this fails on older node versions but fine on 5 and 6, was using v6 locally and had these issues, not sure how they could've been passing before?

Am looking into it.

index.js
```diff
@@ -35,7 +35,9 @@ for (var i = 0; i < filesLoadedBeforeTrace.length; i++) {
   var matches = moduleRegex.exec(filesLoadedBeforeTrace[i]);
   if (matches && matches.length > 1 &&
       modulesLoadedBeforeTrace.indexOf(matches[1]) === -1) {
-    modulesLoadedBeforeTrace.push(matches[1]);
+    if (matches[1] !== '@google' && matches[1] !=='cloud-trace') {
```
matthewloring: Thanks for submitting this PR. Have you observed `'cloud-trace'` appearing in the warning on startup? The `'@google'` check will eliminate the incorrect warning but I'm not seeing a code path where `'cloud-trace'` could appear as well.

Salakar: @matthewloring I added 'cloud-trace' as i was unsure how older node versions would handle the npm package orgs, I'll remove and we'll see on Travis I guess.

Salakar:
@matthewloring I've updated the PR:

- Regex now supports npm org modules, e.g. `@google/cloud-trace` instead of `@google` - see https://regex101.com/r/lW2bE3/4 for examples.
- `filesLoadedBeforeTrace` now ignores `@google/cloud-trace` instead of `@google`

All tests passing.

matthewloring:
@Salakar LGTM. Thank you for putting this together!

matthewloring:
Landed in 2167b16941b89a33aed6d9e3bebc3f9e9cb534d0.

Salakar:
@matthewloring just noticed that this will break in windows, needs correcting to:

```javascript
// Includes support for npm '@org/name' packages
// Regex: .*?node_modules(?!.*node_modules)\/(@[^\/]*\/[^\/]*|[^\/]*).*
// Tests: https://regex101.com/r/lW2bE3/6
var moduleRegex = new RegExp(
  '.*?node_modules(?!.*node_modules)\\' + path.sep +
  '(@[^\\' + path.sep +
  ']*\\' + path.sep +
  '[^\\' + path.sep +
  ']*|[^\\' + path.sep +
  ']*).*'
);
```

Forgot the `\\`

Want me to submit a new PR?

matthewloring:
I've opened it as a new PR. Thanks for catching this.

PR #261, opened 2016-06-13
Set outgoing trace enabled header always

matthewloring:
We currently only emit a trace header on the requests that we sample. This change causes all such outgoing traced requests to set the trace enabled bit.

### DIFF
--- a/lib/trace-agent.js
+++ b/lib/trace-agent.js
@@ -254,15 +254,17 @@ TraceAgent.prototype.parseContextFromHeader = function(str) {
  * @param {SpanData} spanData The span to be added to headers.
  * @param {Object} headers The http headers associated with the
  *   current request.
+ * @param {boolean} traced Whether this request was traced by the agent.
  */
-TraceAgent.prototype.addContextToHeaders = function(spanData, headers) {
+TraceAgent.prototype.addContextToHeaders = function(spanData, headers, traced) {
   if (spanData === SpanData.nullSpan) {
     return;
   }
   var header = spanData.trace.traceId + '/' + spanData.span.spanId;
-  if (spanData.options) {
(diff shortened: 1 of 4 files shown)


matthewloring:

lib/trace-agent.js
```diff
@@ -254,15 +254,17 @@ TraceAgent.prototype.parseContextFromHeader = function(str) {
...
+TraceAgent.prototype.addContextToHeaders = function(spanData, headers, traced) {
   if (spanData === SpanData.nullSpan) {
     return;
   }
   var header = spanData.trace.traceId + '/' + spanData.span.spanId;
-  if (spanData.options) {
-    header += (';o=' + spanData.options);
-  }
+  var options =
+    traced ? spanData.options | constants.TRACE_OPTIONS_TRACE_ENABLED :
```
ofrobots: [Style Guide](https://google.github.io/styleguide/javascriptguide.xml?showone=Code_formatting#Code_formatting): Put the `traced ?` on the preceeding line to make the two legs at similar indent.

test/test-trace-agent.js
```diff
@@ -60,7 +60,7 @@ describe('Trace Agent', function() {
         var options = {
           headers: {}
         };
-        agent.addContextToHeaders(spanData, options.headers);
+        agent.addContextToHeaders(spanData, options.headers, true);
```
cristiancavalli: Does the branch with false not get tested?

cristiancavalli: My bad @matthewloring already explained the difference here

matthewloring: Not a problem, tests for the false branch can be found on line 94.

ofrobots:
LGTM.

PR #263, opened 2016-06-14
Increase time bound for appveyor slowdown

matthewloring:


matthewloring:

PR #266, opened 2016-06-23
update the description for onUncaughtException

ofrobots:
Provide more details on what each of the options for config.trace.onUncaughtException mean for the user's application.

R=@matthewloring, @cristiancavalli.

### DIFF
--- a/config.js
+++ b/config.js
@@ -71,7 +71,9 @@ module.exports = {
     //            application, they may chose to terminate the process before the
     //            buffer has been flushed to the network. Also note that if you have no
     //            other terminating uncaught exception handlers in your application, the
-    //            error will get swallowed and the application will keep on running.
+    //            error will get swallowed and the application will keep on running. You
+    //            should use this option if you have other uncaught exception handlers
+    //            that you want to be responsible for terminating the application.
     //   `flushAndExit`: Handle the uncaught exception, make a best effort attempt to
     //            publish the traces to the API, and then terminate the application after
     //            a delay. Note that presence of other uncaught exception handlers may
--- a/config.js
+++ b/config.js
@@ -64,9 +64,19 @@ module.exports = {
 
     // Specifies the behavior of the trace agent in the case of an uncaught exception.
     // Possible values are:
-    //   `ignore`: Take no action.
-    //   `flush`: Attempt to publish traces silence the exception.
-    //   `flushAndExit`: Attempt to publish traces and exit the process.
+    //   `ignore`: Take no action. Note that the process may termiante before all the 
+    //            traces currently buffered have been flushed to the network.
+    //   `flush`: Handle the uncaught exception and attempt to publish the traces to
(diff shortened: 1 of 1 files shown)


config.js
```diff
@@ -64,9 +64,19 @@ module.exports = {
 
     // Specifies the behavior of the trace agent in the case of an uncaught exception.
     // Possible values are:
-    //   `ignore`: Take no action.
-    //   `flush`: Attempt to publish traces silence the exception.
-    //   `flushAndExit`: Attempt to publish traces and exit the process.
+    //   `ignore`: Take no action. Note that the process may termiante before all the
```
matthewloring: I feel like this would reach more people if it was in the readme.

config.js
```diff
@@ -64,9 +64,19 @@ module.exports = {
...
-    //   `ignore`: Take no action.
-    //   `flush`: Attempt to publish traces silence the exception.
-    //   `flushAndExit`: Attempt to publish traces and exit the process.
+    //   `ignore`: Take no action. Note that the process may termiante before all the 
+    //            traces currently buffered have been flushed to the network.
+    //   `flush`: Handle the uncaught exception and attempt to publish the traces to
+    //            the API. Note that if you have other uncaught exception handlers in your
+    //            application, they may chose to terminate the process before the
+    //            buffer has been flushed to the network. Also note that if you have no
+    //            other terminating uncaught exception handlers in your application, the
```
matthewloring: If there are other terminating uncaught exception handlers then the process should go down and the error will not be swallowed.

config.js
```diff
@@ -64,9 +64,19 @@ module.exports = {
 
     // Specifies the behavior of the trace agent in the case of an uncaught exception.
     // Possible values are:
-    //   `ignore`: Take no action.
-    //   `flush`: Attempt to publish traces silence the exception.
-    //   `flushAndExit`: Attempt to publish traces and exit the process.
+    //   `ignore`: Take no action. Note that the process may termiante before all the 
+    //            traces currently buffered have been flushed to the network.
+    //   `flush`: Handle the uncaught exception and attempt to publish the traces to
```
matthewloring: Maybe this should be renamed to `flushAndSilence`. The current name does not fully capture the action the user should be aware of.

config.js
ofrobots: README should be update too, but it is important for the config to be descriptive too, specially that we point people to the config.js file for documentation on the configuration options.

config.js
ofrobots: updated the comment to be more descriptive.

config.js
ofrobots: I thought about that too; but Silence is not accurate if there are other uncaught exception handlers.

matthewloring: Hmm, its tough because silence accurately describes the behavior of our module but may not describe the behavior of the overall application.

matthewloring:
LGTM.

PR #267, opened 2016-06-23
Support for gRPC time tracing

misterpoe:


misterpoe:
@matthewloring @cristiancavaili @ofrobots PTAL

misterpoe:
I signed it!

lib/hooks/userspace/hook-grpc.js
```diff
@@ -36,13 +79,15 @@ module.exports = function(version_, agent_) {
...
+    'src/node/src/client.js': {
+      patch: function(client) {
         agent = agent_;
-        shimmer.wrap(extension.Call.prototype, 'startBatch', startBatchWrap);
+        shimmer.wrap(client, 'makeClientConstructor',
+            makeClientConstructorWrap);
       },
       unpatch: function(extension) {
-        shimmer.unwrap(extension.Call.prototype, 'startBatch');
+        shimmer.unwrap(client, 'makeClientConstructor');
```
matthewloring: Does unwrapping restore the original grpc operations (makeUnaryRequest, etc).

lib/hooks/userspace/hook-grpc.js
```diff
@@ -19,14 +19,71 @@
...
 var agent;
 
 var SUPPORTED_VERSIONS = '0.13 - 0.15';
 
-function startBatchWrap(startBatch) {
-  return function startBatchTrace(thing, callback) {
-    // TODO: maybe we only want to do this if a root context exists.
-    return startBatch.call(this, thing, cls.getNamespace().bind(callback));
+function wrapClientMethod(name, method, class_options) {
+  return function() {
```
matthewloring: I think we will want to have this function named so the stack traces are meaningful.

lib/hooks/userspace/hook-grpc.js
```diff
@@ -19,14 +19,71 @@
...
+      // Depending on whether deprecatedArgumentOrder is set, the position
+      // of the callback function differs.
+      // We need to wrap the callback with the context, to propagate it.
+      var cbIndex;
+      if (class_options && class_options.deprecatedArgumentOrder) {
+        cbIndex = method.requestStream ? 0 : 1;
+      } else {
+        cbIndex = arguments.length - 1;
+      }
+      if (cbIndex >= 0 && cbIndex < arguments.length) {
```
matthewloring: Can this test ever fail? The above comment would suggest that the callback is always present somewhere in the arguments.

lib/hooks/userspace/hook-grpc.js
```diff
@@ -19,14 +19,71 @@
...
+      }
+    }
+    var call = method.apply(this, arguments);
+    if (agent.config_.enhancedDatabaseReporting) {
+      span.addLabel('call', name);
+    }
+    // The user might need the current context in listeners to this stream.
+    cls.getNamespace().bindEmitter(call);
+    if (method.responseStream) {
+      call.on('end', function() {
```
matthewloring: Is an 'end' event guaranteed to be emitted in the case of a stream api? If a callback is passed to a stream based call, will the span be ended twice?

misterpoe: For the bidirectional and server streams, an 'end' event is emitted when all the data has been read. If a callback is passed (in the case of client streams and unary requests), method.responseStream will not exist so no listener will be binded to the 'end' event (so the span won't be ended twice).

lib/hooks/userspace/hook-grpc.js
misterpoe: It might fail when the user does not give correct arguments to the function (e.g. missed the 'callback' parameter). In that case the call will throw an arguments Error on line 54. I was afraid there might be undefined behavior when accessing negative array indices, but it seems that in JavaScript arguments[-1] is perfectly valid. Should I remove the check then?

matthewloring: Nope, we can leave it. I think this test is better than introducing a callback when the user did not provide one. We want to let gRPC handle this error.

lib/hooks/userspace/hook-grpc.js
matthewloring: Makes sense!

matthewloring:
Everything looks good on my end. Since this is a bigger code review lets get at least one other LGTM before landing this.

lib/hooks/userspace/hook-grpc.js
```diff
@@ -19,14 +19,71 @@
...
+
 var agent;
 
 var SUPPORTED_VERSIONS = '0.13 - 0.15';
 
-function startBatchWrap(startBatch) {
-  return function startBatchTrace(thing, callback) {
-    // TODO: maybe we only want to do this if a root context exists.
-    return startBatch.call(this, thing, cls.getNamespace().bind(callback));
+function clientMethodWrap(name, method, class_options) {
```
ofrobots: nit: a better name for this function would be makeClientMethod

lib/hooks/userspace/hook-grpc.js
```diff
@@ -36,13 +93,15 @@ module.exports = function(version_, agent_) {
...
         agent = agent_;
-        shimmer.wrap(extension.Call.prototype, 'startBatch', startBatchWrap);
+        shimmer.wrap(client, 'makeClientConstructor',
+            makeClientConstructorWrap);
       },
-      unpatch: function(extension) {
-        shimmer.unwrap(extension.Call.prototype, 'startBatch');
+      unpatch: function(client) {
+        shimmer.unwrap(client, 'makeClientConstructor');
+        agent_.logger.info('gRPC makeClientConstructor: unpatched');
```
ofrobots: If I understand the semantics correct, any Client objects that have been already constructed may continue to be traced (i.e. will remain unaffected) by the unpatch.

ofrobots: Is the above correct?

lib/hooks/userspace/hook-grpc.js
```diff
@@ -19,14 +19,71 @@
...
+function clientMethodWrap(name, method, class_options) {
+  return function clientMethodTrace() {
+    var root = cls.getRootContext();
+    if (!root) {
+      agent.logger.debug('Untraced gRPC call: ', name);
+      return method.apply(this, arguments);
+    } else if (root === SpanData.nullSpan) {
+      return method.apply(this, arguments);
+    }
+    var span = agent.startSpan('grpc-call');
```
ofrobots: nit: it would be better to include `name` in the span.

lib/hooks/userspace/hook-grpc.js
```diff
@@ -19,14 +19,71 @@
...
+    done(err, res);
+  };
+  return cls.getNamespace().bind(fn);
+}
+
+function makeClientConstructorWrap(makeClientConstructor) {
+  return function makeClientConstructorTrace(methods, serviceName,
+                                             class_options) {
+    var Client = makeClientConstructor.apply(this, arguments);
+    Object.keys(methods).forEach(function(name) {
```
ofrobots: Use `shimmer.massWrap` on `Client.prototype` instead of doing the monkey patching yourself.

ofrobots: Hmm.. not sure if it works perfectly here, but please investigate if possible.

lib/hooks/userspace/hook-grpc.js
matthewloring: This is correct (and are the same semantics we have for redis which also has to patch client constructors). The way this can be avoided is by keeping a global list of clients we've patched to unpatch them appropriately. During the redis code review this was determined to be a memory leak because the clients could be unused by the application and we maintain a reference.

ofrobots: I agree with the semantics here; I just think they ought to be documented.

lib/hooks/userspace/hook-grpc.js
```diff
@@ -19,14 +19,71 @@
...
+      });
+    }
+    return call;
+  };
+}
+
+function wrapCallback(span, done) {
+  var fn = function(err, res) {
+    agent.endSpan(span);
+    // If we reach this point, done will always be well-defined.
```
ofrobots: What is the reasoning why done will always be well defined?

ofrobots: Okay, I see the path now. Nevermind.

ofrobots:
@murgatroid99 FYI. This is how we are starting to instrument grpc so that we get latency traces for Stackdriver trace. PTAL. A cleaner approach (for future gRPC versions) would be if you could give us an API to gather the data we need around RPCs.

matthewloring:
LGTM. Did you get a chance to verify that this traces the `gcloud` module?

lib/hooks/userspace/hook-grpc.js
```diff
@@ -36,13 +89,19 @@ module.exports = function(version_, agent_) {
...
+        var useDeprecatedArgumentOrder = semver.satisfies(version_, '<0.14');
+        shimmer.wrap(client, 'makeClientConstructor',
+            makeClientConstructorWrap.bind(null, useDeprecatedArgumentOrder));
       },
-      unpatch: function(extension) {
-        shimmer.unwrap(extension.Call.prototype, 'startBatch');
+      unpatch: function(client) {
+        // Note that already wrapped client methods will not be unwrapped here.
+        // Only the client constructor is unwrapped, so that future grpc.load's
+        // will not wrap client methods with tracing.
```
ofrobots: `s/client/Client`. Add: `Existing Client objects with wrapped prototype methods will continue tracing`.

lib/hooks/userspace/hook-grpc.js
ofrobots: Can you add a JSDoc that documents the requirement on the types of span and done? In the future when we get type-checking, it would be good to check against this assertion.

lib/hooks/userspace/hook-grpc.js
```diff
@@ -19,14 +19,73 @@
...
+    done(err, res);
+  };
+  return cls.getNamespace().bind(fn);
+}
+
+function makeClientConstructorWrap(useDeprecatedArgumentOrder,
+                                   makeClientConstructor) {
+  return function makeClientConstructorTrace(methods) {
+    var Client = makeClientConstructor.apply(this, arguments);
+    // If version < 0.14, we need to use the old argument order.
```
ofrobots: I am not sure what the comment has to do with the rest of the code here. Perhaps move it below to when you actually check `version_`.

lib/hooks/userspace/hook-grpc.js
misterpoe: No, I have added a comment for that.

lib/hooks/userspace/hook-grpc.js
misterpoe: Done

lib/hooks/userspace/hook-grpc.js
misterpoe: Done

lib/hooks/userspace/hook-grpc.js
misterpoe: Done

lib/hooks/userspace/hook-grpc.js
misterpoe: Done

lib/hooks/userspace/hook-grpc.js
misterpoe: Done

lib/hooks/userspace/hook-grpc.js
misterpoe: Done

lib/hooks/userspace/hook-grpc.js
misterpoe: Done

misterpoe:
@matthewloring I have tested this on one of the example [Bookshelf apps](https://cloud.google.com/nodejs/getting-started/using-cloud-datastore) and it shows tracing of gRPC calls when I `require('@google/cloud-trace')`.

ofrobots:
LGTM

PR #268, opened 2016-06-28
Add test fixture for gRPC 0.15

misterpoe:


misterpoe:

test/hooks/test-trace-grpc.js
```diff
@@ -230,6 +231,40 @@ Object.keys(versions).forEach(function(version) {
           stream.end();
         });
       });
+    } else {
+      it('should work for new argument orders (unary)', function(done) {
+        common.runInTransaction(function(endTransaction) {
+          client.testUnary({n: 42}, new grpc.Metadata(), {},
```
matthewloring: What is the empty object here?

misterpoe: The signature of the function is `makeUnaryRequest(argument, metadata, options, callback)`, where `metadata` and `options` are optional. So the empty object is `options`.

matthewloring:
LGTM.

PR #270, opened 2016-07-01
Support enhanced database reporting for gRPC

misterpoe:


misterpoe:

lib/hooks/userspace/hook-grpc.js
```diff
@@ -25,6 +25,15 @@ var agent;
 
 var SUPPORTED_VERSIONS = '0.13 - 0.15';
 
+function findFirstIndex(arr, predicate) {
```
matthewloring: Let's bring in [lodash.find](https://lodash.com/docs#find) for this. If only [Array.prototype.find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) was present in `0.12` 😿

test/hooks/test-trace-grpc.js
```diff
@@ -199,69 +217,125 @@ Object.keys(versions).forEach(function(version) {
...
+      });
+    });
+
+    it('should trace metadata for server requests', function(done) {
+      common.runInTransaction(function(endTransaction) {
+        var stream = client.testServerStream({n: 42}, metadata);
+        stream.on('data', function(data) {});
+        stream.on('status', function(status) {
+          endTransaction();
+          assert.strictEqual(status.code, grpc.status.OK);
```
matthewloring: Would it be possible to report this status code with `enhancedDatabaseReporting`?

lib/hooks/userspace/hook-grpc.js
misterpoe: Done

test/hooks/test-trace-grpc.js
misterpoe: Yes, we can listen for the 'status' event. I've added the test for that to the standard server stream request above (this test is for metadata).

matthewloring:
LGTM.

PR #271, opened 2016-07-01
Change callback lookup to not depend on gRPC version

misterpoe:


misterpoe:

matthewloring:
LGTM.

PR #276, opened 2016-07-06
Begin using test all versions for hooks

matthewloring:
This change moves express, mysql, and restify over to use
test-all-versions as opposed to committed fixtures.


matthewloring:

ofrobots:
As discussed, this looks interesting; but it still requires extra work to test *all* versions and to be able to support cross-product verification with multiple modules. Can this be closed?

matthewloring:
Yup!

PR #285, opened 2016-08-01
Fix mongoose promise deprecation

matthewloring:
Mongoose updated and dropped support for its own promise library
mpromise. The test will now rely on builtin promises.

### DIFF
--- a/test/hooks/test-trace-mongoose.js
+++ b/test/hooks/test-trace-mongoose.js
@@ -25,6 +25,7 @@ var traceLabels = require('../../lib/trace-labels.js');
 
 var assert = require('assert');
 var mongoose = require('./fixtures/mongoose4');
+mongoose.Promise = global.Promise;
 var Schema = mongoose.Schema;
 
 var simpleSchema = new Schema({


matthewloring:

ofrobots:
Please include some rationale in the commit message (and hopefully also in the PR) about the rationale. Why is this change being made? Mongoose has changed something that needs to be addressed or did the test have a problem as written?

matthewloring:
I updated the description. Mongoose updated and dropped support for its own promise library mpromise.

ofrobots:
Without looking at mongoose and how mongoose.Promise is used internally.. what breaks in our test if we don't monkey patch mongoose.Promise?

matthewloring:
Appveyor parses the deprecation as a failure. Travis/local runs are unaffected.

matthewloring:
Updated the commit message to link to: https://github.com/Automattic/mongoose/wiki/5.0-Deprecation-Warnings.

ofrobots:
LGTM

PR #288, opened 2016-08-03
Add trace context to traced http responses

matthewloring:


matthewloring:

lib/trace-agent.js
```diff
@@ -247,25 +247,22 @@ TraceAgent.prototype.parseContextFromHeader = function(str) {
...
  * @param {SpanData} spanData The span to be added to headers.
- * @param {Object} headers The http headers associated with the
- *   current request.
  * @param {boolean} traced Whether this request was traced by the agent.
  */
-TraceAgent.prototype.addContextToHeaders = function(spanData, headers, traced) {
+TraceAgent.prototype.generateTraceContext = function(spanData, traced) {
   if (spanData === SpanData.nullSpan) {
-    return;
+    return '';
```
ofrobots: What is the case where we reach here with `nullSpan`? In such a scenario we will be adding an empty string header.

matthewloring: We currently validate that we are tracing the request (`spanData !== SpanData.nullSpan`) before calling this method in all cases. The intent would be to check for empty string at any call site that cannot verify that `spanData` is not the null span before adding it to headers.

ofrobots: If such a call site doesn't exist today, then YAGNI.

matthewloring: Done. Added comment specifying no nullSpan.

matthewloring: Actually, I'm not sure this is a good idea. The nullspan is a valid input to this function but it will cause a crash. Additionally, it will be very easy to introduce a bug that will only manifest on sampled requests which will be less likely to be caught in unit tests. I would prefer to leave this in and comment specifying that the empty header should be checked before writing to headers. WDYT?

ofrobots: Why not throw?

matthewloring: Updated with checks and logs.

ofrobots:
LGTM

PR #295, opened 2016-09-02
Allow ignoring the requests Context header

jerryjj:
Hi,

Short background: I've been debugging for a week an issue that started occurring last week in our GCP projects after the Google Load Balancers got an update.
The LB doesn't respect my already set "x-cloud-trace-context" -header and overwrites it.
This makes our Traces to disappear because of our complex infrastructural setup.

This simple patch allows us to continue the usage of this module without any problems.
This will help others also who have similar infrastructural setup (I can write a description of it, if necessary).

Hopefully this could be included in the main package, so I do not need to keep installing this with npm from my github.

### DIFF
--- a/test/test-trace-agent.js
+++ b/test/test-trace-agent.js
@@ -163,11 +163,11 @@ describe('Trace Agent', function() {
 
     describe('when configured to ignore header', function() {
       it('should return expected value: null', function() {
-        var agent2 = agent;
-        agent2.config_.ignoreContextHeader = true;
-        var result = agent2.parseContextFromHeader(
+        agent.config_.ignoreContextHeader = true;
+        var result = agent.parseContextFromHeader(
           '123456/667;o=1');
         assert(!result);
+        agent.config_.ignoreContextHeader = false;
       });
     });
   });
--- a/test/test-trace-agent.js
+++ b/test/test-trace-agent.js
@@ -160,6 +160,16 @@ describe('Trace Agent', function() {
         });
       });
     });
+
+    describe('when configured to ignore header', function() {
+      it('should return expected value: null', function() {
+        var agent2 = agent;
+        agent2.config_.ignoreContextHeader = true;
+        var result = agent2.parseContextFromHeader(
+          '123456/667;o=1');
+        assert(!result);
+      });
+    });
   });
 
 });
--- a/config.js
+++ b/config.js
@@ -64,7 +64,7 @@ module.exports = {
 
     // Specifies the behavior of the trace agent in the case of an uncaught exception.
     // Possible values are:
-    //   `ignore`: Take no action. Note that the process may termiante before all the 
+    //   `ignore`: Take no action. Note that the process may termiante before all the
     //            traces currently buffered have been flushed to the network.
     //   `flush`: Handle the uncaught exception and attempt to publish the traces to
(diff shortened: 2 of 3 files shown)


matthewloring:
@jerryjj Thank you for putting together this patch! It looks good with a minor nit. I will check in with the server team to get a better understanding of why the change in LB behavior occurred. 

I would also be interested to learn more about your custom infrastructure setup just to better understand your usage of trace.

test/test-trace-agent.js
```diff
@@ -160,6 +160,16 @@ describe('Trace Agent', function() {
...
       });
     });
+
+    describe('when configured to ignore header', function() {
+      it('should return expected value: null', function() {
+        var agent2 = agent;
+        agent2.config_.ignoreContextHeader = true;
+        var result = agent2.parseContextFromHeader(
+          '123456/667;o=1');
+        assert(!result);
```
matthewloring: Could the value of `agent.config_. ignoreContextHeader` be reset to the default value `false` here so that it does not remain set for tests run after this one?

test/test-trace-agent.js
```diff
@@ -160,6 +160,16 @@ describe('Trace Agent', function() {
         });
       });
     });
+
+    describe('when configured to ignore header', function() {
+      it('should return expected value: null', function() {
+        var agent2 = agent;
```
matthewloring: Aliasing the agent here is not necessary.

test/test-trace-agent.js
jerryjj: Thanks for the feedback, fixed now.

matthewloring:
@jerryjj The changes look great. Would you be able to describe your infrastructure so we can understand the benefit of ignoring the incoming trace headers?

matthewloring:
I'm going to land this. Thank you for putting it together. I would still be interested in understanding your infrastructure that benefits from this change when you have time.

jerryjj:
@matthewloring Thanks for approving this.
Here is a simplified description of the infrastructure I have in place, that required this change (it used to work without it for a long time)

I have a Google Cloud Platform Project (lets call it "master-project") and then I have supporting projects for different environments (development, staging, production, etc.)
In the master-project I have a GKE (Container Engine) running pods in different namespaces (staging, production, etc). Also I have enabled HTTPS Load Balancing in that project, which terminates the incoming TLS communication and forwards the traffic to the underlying K8s cluster.
The micro-services running inside the K8s cluster sends their Cloud Traces to the respective supporting project (staging, production, etc) using Service Accounts. 

Before the Load Balancer -service update couple of weeks ago, this worked fine if we sent a empty "X-Cloud-Trace-Context" header in our incoming requests. But after the update, the LB started to overwrite this, which lead to the situation, that the Root span was written in the master-project's Cloud Trace and the child spans were written in the respective supporting projects Cloud Traces using the received Root span id as parent. Which of course made these traces never to be found from the Cloud Trace console, as none of them had a real Root span.
By ignoring the header coming from the LB, we are able to again see all the traces properly.

It would also be really great to have a view in the Cloud Trace console, for these kind of orphaned spans (would have saved me from days of work trying to figure out why my traces aren't visible anymore).

Hopefully this description helps and also the solution helps others that might end up in similar troubles.

PR #296, opened 2016-09-07
Document ignoreContextHeader configuration

matthewloring:


matthewloring:

config.js
```diff
@@ -81,7 +81,10 @@ module.exports = {
     //            the network.
     onUncaughtException: 'ignore',
 
-    // Allows to ignore the requests X-Cloud-Trace-Context -header if set
+    // EXPERIMENTAL:
+    // Allows to ignore the requests X-Cloud-Trace-Context -header if set. This will
```
ofrobots: s/This will/Setting this to true will/.

Why would one want this behaviour? 

Doesn't setting this to true also imply that a local sampling decision is made an the sampling/tracing decision from the incoming request is ignored?

matthewloring: Description questions addressed. @jerryjj Could you describe how this behavior is helpful to you so we can include it in the description of the config option?

jerryjj: @matthewloring I wrote a simplified description here: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/295#issuecomment-245531802

matthewloring: @jerryjj Thank you for the description!

config.js
```diff
@@ -81,7 +81,13 @@ module.exports = {
...
     onUncaughtException: 'ignore',
 
-    // Allows to ignore the requests X-Cloud-Trace-Context -header if set
+    // EXPERIMENTAL:
+    // Allows to ignore the requests X-Cloud-Trace-Context -header if set. Setting this
+    // to true will cause traces generated by this module to appear separately from other
+    // distributed work done by other services on behalf of the same income request. Setting
+    // this will also cause sampling decisions made by other distributed components to be
+    // ignored.
+    // This is useful for aggregating traces generated by different cloud platform projects.
```
ofrobots: nit: merge with previous paragraph.

ofrobots:
LGTM w/ nit

config.js

PR #298, opened 2016-09-12
Document using just the trace.append scope

zbjornson:
As far as I've tested today, it work fine to just use `https://www.googleapis.com/auth/trace.append` instead of the whole `cloud-platform` scope.

Ref #248

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ If you are using [Google App Engine flexible environment](https://cloud.google.c
 
 ### Google Compute Engine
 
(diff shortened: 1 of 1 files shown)


matthewloring:
\cc @JustinBeckwith

ofrobots:
LGTM. I agree that we should drop the reference to the cloud-platform scope, but this can be amended at landing time.

zbjornson:
Yeah, as long as folks don't then think that they have to have trace.append in addition to cloud-platform... (not that that's the end of the world).

matthewloring:
Landed as b9e4848c141425ef24e21d02bf95a9ed5f248168. Thank you for submitting this @zbjornson

PR #301, opened 2016-09-16
Implement gRPC Server Tracing

kjin:
This change introduces gRPC server function tracing. In addition to the timespan itself, the arguments, return results, and any metadata or errors are written in trace spans as well.

### DIFF
--- a/lib/hooks/userspace/hook-grpc.js
+++ b/lib/hooks/userspace/hook-grpc.js
@@ -17,6 +17,7 @@
 'use strict';
 
 var cls = require('../../cls.js');
+var TraceLabels = require('../../trace-labels.js');
 var shimmer = require('shimmer');
 var semver = require('semver');
 var SpanData = require('../../span-data.js');
@@ -176,7 +177,7 @@ function wrapBidi(namespace, handlerSet, requestName) {
       var endSpan = function () {
         if (!spanEnded) {
           spanEnded = true;
-          endRootSpanForRequest(rootContext, requestName);
+          endRootSpanForRequest(rootContext);
         }
       };
       stream.on('finish', function () {
@@ -227,7 +228,7 @@ function wrapServerStream(namespace, handlerSet, requestName) {
       var endSpan = function () {
         if (!spanEnded) {
           spanEnded = true;
-          endRootSpanForRequest(rootContext, requestName);
+          endRootSpanForRequest(rootContext);
         }
       };
       call.on('finish', function () {
@@ -273,8 +274,7 @@ function wrapClientStream(namespace, handlerSet, requestName) {
         if (agent.config_.enhancedDatabaseReporting) {
           if (err) {
             rootContext.addLabel('error', err);
-          }
-          else {
+          } else {
             rootContext.addLabel('result', JSON.stringify(result));
           }
           if (trailer) {
@@ -284,7 +284,7 @@ function wrapClientStream(namespace, handlerSet, requestName) {
         if (err) {
           // If there is an error, then no result will be serialized to
           // be sent back to the client. So end the root span here.
-          endRootSpanForRequest(rootContext, requestName);
+          endRootSpanForRequest(rootContext);
         }
         return callback(err, result);
       };
@@ -294,7 +294,7 @@ function wrapClientStream(namespace, handlerSet, requestName) {
   shimmer.wrap(handlerSet, 'serialize', function (serialize) {
     return function serializeTrace(result) {
       var serializeResult = serialize.apply(this, arguments);
-      endRootSpanForRequest(rootContext, requestName);
+      endRootSpanForRequest(rootContext);
       return serializeResult;
     };
   });
@@ -336,8 +336,7 @@ function wrapUnary(namespace, handlerSet, requestName) {
         if (agent.config_.enhancedDatabaseReporting) {
           if (err) {
             rootContext.addLabel('error', err);
-          }
-          else {
+          } else {
             rootContext.addLabel('result', JSON.stringify(result));
           }
           if (trailer) {
@@ -347,7 +346,7 @@ function wrapUnary(namespace, handlerSet, requestName) {
         if (err) {
           // If there is an error, then no result will be serialized to
           // be sent back to the client. So end the root span here.
-          endRootSpanForRequest(rootContext, requestName);
+          endRootSpanForRequest(rootContext);
         }
         return callback(err, result);
       };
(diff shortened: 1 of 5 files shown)


matthewloring:
Looks like there are a few lint errors to clean up: https://travis-ci.org/GoogleCloudPlatform/cloud-trace-nodejs/jobs/160330433.

lib/hooks/userspace/hook-grpc.js
```diff
@@ -129,6 +128,299 @@ function makeClientConstructorWrap(makeClientConstructor) {
...
+  // End right after serialization (which is done in the callback)
+  shimmer.wrap(handlerSet, 'func', function (func) {
+    return namespace.bind(function funcTrace(stream, callback) {
+      rootContext = startRootSpanForRequest(requestName);
+      if (agent.config_.enhancedDatabaseReporting) {
+        shimmer.wrap(stream, 'sendMetadata', sendMetadataWrapper(rootContext));
+      }
+      // Preserve context in stream event handlers.
+      namespace.bindEmitter(stream);
+      // arguments[1] is the callback
```
matthewloring: Are there any optional arguments that could cause the callback to be at a different position?

lib/hooks/userspace/hook-grpc.js
```diff
@@ -129,6 +128,299 @@ function makeClientConstructorWrap(makeClientConstructor) {
...
+    var requestName = 'grpc:' + name;
+    // Proceed to wrap methods that are invoked when a gRPC service call is made.
+    // In every case, the function 'func' is the user-implemented handling function.
+    if (method_type === 'bidi') {
+      wrapBidi(namespace, handlerSet, requestName);
+    } else if (method_type === 'server_stream') {
+      wrapServerStream(namespace, handlerSet, requestName);
+    } else if (method_type === 'client_stream') {
+      wrapClientStream(namespace, handlerSet, requestName);
+    } else { // method_type === 'unary'
```
matthewloring: It would be good to explicitly check that we are seeing `'unary'` and add an else case which warns if that isn't the case.

lib/hooks/userspace/hook-grpc.js
```diff
@@ -129,6 +128,299 @@ function makeClientConstructorWrap(makeClientConstructor) {
...
+      }
+      if (agent.config_.enhancedDatabaseReporting) {
+        rootContext.addLabel('argument', JSON.stringify(call.request));
+      }
+      // arguments[1] is the callback
+      arguments[1] = function (err, result, trailer) {
+        if (agent.config_.enhancedDatabaseReporting) {
+          if (err) {
+            rootContext.addLabel('error', err);
+          }
```
matthewloring: prefer `else` on the same line throughout this change.

lib/hooks/userspace/hook-grpc.js
```diff
@@ -129,6 +128,299 @@ function makeClientConstructorWrap(makeClientConstructor) {
...
+      };
+      stream.on('finish', function () {
+        // End the span unless there is an error.
+        if (stream.status.code === 0) {
+          endSpan();
+        }
+      });
+      stream.on('error', function (err) {
+        if (agent.config_.enhancedDatabaseReporting) {
+          rootContext.addLabel('error', err);
```
matthewloring: Do you know how human readable these error objects are?

lib/hooks/userspace/hook-grpc.js
```diff
@@ -129,6 +128,299 @@ function makeClientConstructorWrap(makeClientConstructor) {
...
+          endRootSpanForRequest(rootContext, requestName);
+        }
+        return callback(err, result);
+      };
+      return func.apply(this, arguments);
+    }, context);
+  });
+  shimmer.wrap(handlerSet, 'serialize', function (serialize) {
+    return function serializeTrace() {
+      var serializeResult = serialize.apply(this, arguments);
```
matthewloring: Is serialization done synchronously here?

lib/hooks/userspace/hook-grpc.js
kjin: From what I've seen on the GCP Trace UI it displays the error in plain text (as opposed to JSON). It should be the same string as on the client side. I believe the tests confirm this.

lib/hooks/userspace/hook-grpc.js
```diff
@@ -129,6 +128,299 @@ function makeClientConstructorWrap(makeClientConstructor) {
...
+}
+
+/**
+ * Creates and sets up a new root span for the given request.
+ * @param {Object} req The request being processed.
+ * @returns {!SpanData} The new initialized trace span data instance.
+ */
+function startRootSpanForRequest(req) {
+  var rootContext = agent.createRootSpanData(req, null, null, 1);
+  rootContext.span.kind = 'RPC_SERVER';
```
kjin: My mistake. Though I think root span kind should default to `RPC_SERVER`, only being set otherwise when a custom span is created - what do you think?

lib/hooks/userspace/hook-grpc.js
kjin: That is a good point - yes, both the docs and usage of serialize suggest that it's synchronous.

lib/hooks/userspace/hook-grpc.js
kjin: Also a good point - it seems like arguments[1] is always the callback.

lib/hooks/userspace/hook-grpc.js
```diff
@@ -129,6 +129,298 @@ function makeClientConstructorWrap(makeClientConstructor) {
...
+          if (trailer) {
+            rootContext.addLabel('trailing_metadata', JSON.stringify(trailer.getMap()));
+          }
+        }
+        if (err) {
+          // If there is an error, then no result will be serialized to
+          // be sent back to the client. So end the root span here.
+          endRootSpanForRequest(rootContext);
+        }
+        return callback(err, result);
```
matthewloring: Will this result in trailer being dropped?

lib/hooks/userspace/hook-grpc.js
```diff
@@ -129,6 +129,298 @@ function makeClientConstructorWrap(makeClientConstructor) {
...
```
matthewloring: And here.

lib/hooks/userspace/hook-grpc.js
```diff
@@ -129,6 +129,298 @@ function makeClientConstructorWrap(makeClientConstructor) {
...
+ * Wraps a bidirectional streaming function in order to record trace spans.
+ * @param {cls.Namespace} namespace The CLS namespace.
+ * @param {Object} handlerSet An object containing references to the function handle,
+ * as well as serialize and deserialize handles.
+ * @param {string} requestName The human-friendly name of the request.
+ */
+function wrapBidi(namespace, handlerSet, requestName) {
+  var rootContext;
+  // CLS context object
+  var context = {};
```
matthewloring: Could we add a brief comment here on why we need to explicitly chain contexts using bind instead of the run pattern used in the other hooks?

lib/hooks/userspace/hook-grpc.js
kjin: Good catch - also a fourth parameter that I had overlooked... thanks!

PR #309, opened 2016-10-03
Change connect trace URL prefix and add tests for connect hook

kjin:
Connect spans now record the protocol and hostname as a prefix to the original URL.

Also, I've added some tests that are pretty similar to the Express ones.


kjin:

matthewloring:
*LGTM pending travis passing, looks like a few nits.

PR #311, opened 2016-10-05
More thorough gRPC unpatching and added corresponding tests

kjin:
gRPC unpatching previously only unwrapped the register function which is only called once. Now it unwraps the functions that were wrapped for each service.

### DIFF
--- a/lib/hooks/userspace/hook-grpc.js
+++ b/lib/hooks/userspace/hook-grpc.js
@@ -169,7 +169,9 @@ function wrapUnary(namespace, handlerSet, requestName) {
   // End right after serialization
   shimmer.wrap(handlerSet, 'deserialize', function (deserialize) {
     return namespace.bind(function deserializeTrace() {
-      rootContext = startRootSpanForRequest(requestName);
+      if (agent.isActive()) {
+        rootContext = startRootSpanForRequest(requestName);
+      }
       return deserialize.apply(this, arguments);
     }, context);
   });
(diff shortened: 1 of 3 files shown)


kjin:

lib/hooks/userspace/hook-grpc.js
```diff
@@ -365,43 +365,79 @@ function wrapBidi(namespace, handlerSet, requestName) {
...
+        wrapServerStream(namespace, handlerSet, requestName);
+      } else if (method_type === 'client_stream') {
+        wrapClientStream(namespace, handlerSet, requestName);
+      } else if (method_type === 'bidi') {
+        wrapBidi(namespace, handlerSet, requestName);
+      } else {
+        agent.logger.warn('gRPC Server: Unrecognized method_type ' +
+          method_type);
+      }
+      unwrapFnList.push(function() {
```
matthewloring: I believe storing these functions will introduce a memory leak as they will hold reference to pieces of the server after the application code no longer has need of them. I don't think this is as big a problem as if it were a client library patch since servers shouldn't be created/destroyed frequently. @ofrobots Is this a leak we should be concerned about?

kjin: Discussed in person and arrived at the conclusion that we could instead have patched service methods lazily unpatch themselves the next time they're invoked.

ofrobots:
I am not sure if I am the biggest fan of this style of unpatching.

1. First of all, I don't think we necessarily need to have a strong contract of completely removing performance when un-patching is performed. 
2. The current PR does unpatching lazily which means that there is a performance overhead on each trace (the primary use case of this module). A preferable design would be to register functions to be invoked once when agent is deactivated. These could be queued at the time the patch is applied.
3. Lack of symmetry in calls to `shimmer.unwrap` is bothersome. Similarly, the current patch depends on `shimmer` internals in expecting a `__wrapped` property.

I am not sure if it is worth implementing 2.

kjin:
@ofrobots The shimmer API states that "All monkeypatched functions have an attribute, `__wrapped`, set to true on them."

When you say lack of symmetry, are you referring to the fact that the set of `shimmer.wrap` calls is different for each service method type is different?

In either case - I'd be fine with not unpatching the service methods - this would look more or less like the second-to-last commit.

ofrobots:
The asymmetry is that I noticed was that `deserialize` seems to be responsible for unwrapping `func` and `serialize` as well.

matthewloring:
It sounds like we've decided not to move forward with this for now.

PR #315, opened 2016-10-27
Implemented reading keyFile/credentials field from config object

kjin:
Clients may use `config.keyFile` and `config.credentials` (or `config.keyFilename` - an alias for the former) to supply a key file or its contents in lieu of `GCLOUD_APPLICATION_CREDENTIALS`.


kjin:

README.md
```diff
@@ -84,7 +84,9 @@ If your application is running outside of Google Cloud Platform, such as locally
   2. Copy the key somewhere your application can access it. Be sure not to expose the key publicly.
   3. Set the environment variable `GOOGLE_APPLICATION_CREDENTIALS` to the full path to the key. The debug agent will automatically look for this environment variable.
```
matthewloring: Do we still want this to be the recommended way or is the config preferable?

config.js
```diff
@@ -88,6 +88,21 @@ module.exports = {
...
+    ignoreContextHeader: false,
+
+    // A path to a key file relative to the current working directory. If this
+    // field is set, the contents of the pointed file will be used for
+    // authentication instead of your application default credentials.
+    keyFile: null,
+
+    // An alias for keyFile. If both keyFile and keyFilename are set, the value
+    // of keyFilename will be ignored.
+    keyFilename: null,
```
matthewloring: Was it a requirement to support this alias? I would prefer not to have duplicate config options unless there is a strong use case for different names?

test/standalone/test-config-credentials.js
```diff
@@ -0,0 +1,96 @@
...
+
+var queueSpans = function(n, privateAgent) {
+  for (var i = 0; i < n; i++) {
+    privateAgent.createRootSpanData('name', 1, 0).close();
+  }
+};
+
+process.env.GCLOUD_PROJECT = 0;
+
+describe('test-config-credentials', function() {
```
matthewloring: We should have at least one test that sets both `credentials` and `keyFilename` to make sure precedence is handled correctly.

lib/trace-writer.js
```diff
@@ -40,7 +42,10 @@ function TraceWriter(logger, config) {
   this.config_ = config;
 
   /** @private {function} authenticated request function */
-  this.request_ = utils.authorizedRequestFactory(SCOPES);
+  this.request_ = utils.authorizedRequestFactory(SCOPES, {
+    credentials: config.credentials,
+    keyFile: config.keyFile || config.keyFilename
```
matthewloring: Will passing in a `null` value for `keyFile` result in `GOOGLE_APPLICATION_CREDENTIALS` being used?

README.md
kjin: I believe setting `GOOGLE_APPLICATION_CREDENTIALS` is preferable.

config.js
kjin: Not a requirement - `keyFilename` is part of google-cloud's config options so we should use that. I don't remember where I saw `keyFile` from, I thought it was part of a public interface in another library. I can remove it.

README.md
```diff
@@ -84,7 +84,9 @@ If your application is running outside of Google Cloud Platform, such as locally
...
 
-3. Alternatively, if you are running your application on a development machine or test environment where you are using the [`gcloud` command line tools][gcloud-sdk], and are logged using `gcloud beta auth application-default login`, you already have sufficient credentials, and a service account key is not required.
+If you are running your application on a development machine or test environment where you are using the [`gcloud` command line tools][gcloud-sdk], and are logged using `gcloud beta auth application-default login`, you already have sufficient credentials, and a service account key is not required.
+  
+Alternatively, you may set the `keyFilename` or `credentials` configuration field to the full path or contents to the key file, respectively. Setting either of these fields will override both of the above options. (See the [default configuration](config.js) for more details.)
```
ofrobots: Reading this, it is not clear to me what `both of the above options` are.

ofrobots: Secondly, IMO, the environment variable should override the credentials provided in the source code (config) or obtained through the metadata service.

What is the behaviour of the [google-cloud](http://npm.im/google-cloud) modules here in this regard?

README.md
ofrobots: Rationale? [google-cloud](http://npm.im/google-cloud) modules suggest config as the default mechanism.

kjin: I think authentication options set in `config` should only be used as an explicit override in case users don't want to use application-default credentials. Going with the idea that the library should be as simple to use as possible, the config object should be minimal (and ideally, unnecessary at all).

This [document](https://google-cloud-python.readthedocs.io/en/latest/google-cloud-auth.html) (from the Python google-cloud) also seems to suggest that explicit credentials should be reserved for specific use cases. The Node.js version doesn't mention `GOOGLE_APPLICATION_CREDENTIALS` anywhere.

README.md
ofrobots: Yikes. I think we should have consistency in behaviour of the environment variables between all the diagnostics agents and google-cloud. @matthewloring thoughts?

matthewloring: My intuition is that an env_var change is easier than a code change so it would be nice if `GOOGLE_APPLICATION_CREDENTIALS` overrode `keyFilename` and `credentials`. Maybe we could open an issue and see if auto auth will change their precedence? If not, we should change our stuff to match them for consistency.

kjin: @ofrobots I don't think my changes are inconsistent with google-cloud - in both cases, config options take precedence over other ways to supply credentials.

@matthewloring Though there's no direct equivalent to the way the Node gcloud library handle config options, it seems that options specified in-code overrides anything other ways of supplying credentials consistently among the gcloud libraries in other languages (see [Java](https://github.com/GoogleCloudPlatform/google-cloud-java#specifying-a-project-id) and [Python](https://google-cloud-python.readthedocs.io/en/latest/google-cloud-auth.html#explicit-credentials).

ofrobots: I think this opens a broader question about consistent behaviour that should be addressed independently in a semver major change. The current PR seems like a consistent addition of behaviour.

kjin:
Appveyor failure seems to be due to an unrelated issue (gRPC 1.0.1 compilation on Windows).

config.js
```diff
@@ -88,6 +88,16 @@ module.exports = {
...
+
+    // A path to a key file relative to the current working directory. If this
+    // field is set, the contents of the pointed file will be used for
+    // authentication instead of your application default credentials.
+    keyFilename: null,
+
+    // The contents of a key file. If this field is set, its contents will be
+    // used for authentication instead of your application default credentials.
+    // If keyFilename is also set, the value of credentials will be ignored.
+    credentials: null
```
DominicKramer: Perhaps it could be easier for users if there was just a single `credentials` attribute.  If the value of the attribute was a file path, it would be assumed to the filename to the key file.  Otherwise, the value would be assumed to be the contents of the key file.

The advantage is that there is one attribute that specifies credentials so that if the user encounters problems, there are less interrelated attributes to consider, and the single attribute handles priority.

The disadvantage is that a single attribute no longer has a single interpretation.

Just an idea.

PR #320, opened 2016-11-11
Remove query parameters from span names

matthewloring:


matthewloring:
PTAL.

lib/hooks/core/hook-http.js
```diff
@@ -63,7 +63,7 @@ function requestWrap(request) {
     var labels = {};
     labels[TraceLabels.HTTP_METHOD_LABEL_KEY] = options.method;
     labels[TraceLabels.HTTP_URL_LABEL_KEY] = uri;
-    var span = agent.startSpan(uri, labels);
+    var span = agent.startSpan(options.hostname || options.host || 'localhost', labels);
```
ofrobots: It would be better to extract this into a function similar to `uriFromOptions` along with the documentation on why the order makes sense.

ofrobots: Actually, why isn't `uri.host` the thing we should use here?

ofrobots: D'oh. I was misreading `uriFromOptions` it returns a formatted uri.

ofrobots: A simple comment here would be good
// c.f. _http_client.js ClientRequest constructor

matthewloring: Done.

lib/hooks/userspace/hook-hapi.js
```diff
@@ -88,7 +89,7 @@ function startRootSpanForRequest(req, traceHeader) {
 
   // we use the path part of the url as the span name and add the full
   // url as a label
-  var rootContext = agent.createRootSpanData(req.url, traceId, parentSpanId, 3);
+  var rootContext = agent.createRootSpanData(urlParse(req.url).pathname, traceId, parentSpanId, 3);
```
ofrobots: Wouldn't `req.path` work here?

Doc: http://hapijs.com/api#request-properties
Available in hapi@8 as well as hapi@latest as per the docs.

lib/hooks/userspace/hook-koa.js
```diff
@@ -91,7 +92,7 @@ function startRootSpanForRequest(req, traceHeader) {
 
   // we use the path part of the url as the span name and add the full
   // url as a label
-  const rootContext = agent.createRootSpanData(req.url, traceId,
+  const rootContext = agent.createRootSpanData(urlParse(req.url).pathname, traceId,
```
ofrobots: Would `req.path` work here? Doc: http://koajs.com/#request

lib/hooks/userspace/hook-hapi.js
matthewloring: It is not set at the time our code runs. I've added a comment to that effect.

lib/hooks/userspace/hook-koa.js
matthewloring: Same.

PR #322, opened 2016-11-14
Begin testing against v7 on travis

matthewloring:


matthewloring:
PTAL

matthewloring:
There have been some changes since the LGTM because restify3 and grpc <1 are not supported for node 7. PTAL.

ofrobots:
Is restify3 never going to support node 7, or is it a point in time
statement?

matthewloring:
The last commit on the 3 line was 5/19/15 and there doesn't appear to be an active branch on github corresponding to 3.x. I couldn't find any official documentation on their support.

PR #331, opened 2016-12-09
Remove unncessary environment variables

matthewloring:


test/standalone/test-config-priority.js
```diff
@@ -1,42 +0,0 @@
...
-process.env.GCLOUD_DIAGNOSTICS_CONFIG =
-  path.join(__dirname, '..', 'fixtures', 'test-config.js');
-
-process.env.GCLOUD_TRACE_LOGLEVEL = 2;
-
-var agent = require('../..').start({logLevel: 3, stackTraceLimit: 2,
-  flushDelaySeconds: 31});
-
-describe('should respect config load order', function() {
-  it('should order Default -> start -> env config -> env specific', function() {
```
ofrobots: You're getting rid of 'env config' from the list, but the rest of priority should still be tested, right?

matthewloring: I believe the other priority combinations are validated by the appropriate `standalone/test-env-*.js` files. WDYT?

ofrobots: sgtm

PR #333, opened 2016-12-10
Rename the `lib` directory to `src`

DominicKramer:
Renamed the `lib` directory to `src` to more closely align this module with the other Google Cloud API Client Libraries via the `google-cloud` modules in [1].


DominicKramer:
PTAL

PR #335, opened 2016-12-14
Specify package name in version string

matthewloring:


src/trace-writer.js
```diff
@@ -67,7 +67,7 @@ function TraceWriter(logger, config) {
   that.getHostname(function(hostname) {
     that.getInstanceId(function(instanceId) {
       var labels = {};
-      labels[traceLabels.AGENT_DATA] = 'node ' + pjson.version;
+      labels[traceLabels.AGENT_DATA] = 'node ' + pjson.name + ' v' + pjson.version;
```
DominicKramer: Is it possible for `pjson.name` and/or `pjson.version` to be undefined?

matthewloring: https://docs.npmjs.com/files/package.json#name

DominicKramer: I see.  Then LGTM

PR #337, opened 2016-12-17
The module now exports a constructor

DominicKramer:
Now this module exports a constructor.  In addition, the `start`
method has been renamed to `startAgent`.  These changes were made
to more closely align this module with the other Google Cloud API
Client Libraries via the `google-cloud` modules in [1].

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ This module provides Stackdriver Trace support for Node.js applications. [Stackd
 
 3. Include and start the library *as the very first action in your application*:
 
-        require('@google/cloud-trace').start();
+        require('@google/cloud-trace')().startAgent();
 
   If you use `--require` in your start up command, make sure that the trace agent is --required first.
 
@@ -47,7 +47,7 @@ If you are running somewhere other than the Google Cloud Platform, see [running 
 
 See [the default configuration](config.js) for a list of possible configuration options. These options can be passed to the agent through the object argument to the start command shown above:
 
-        require('@google/cloud-trace').start({samplingRate: 500});
+        require('@google/cloud-trace')().startAgent({samplingRate: 500});
 
 Alternatively, you can provide configuration through a config file. This can be useful if you want to load our module using `--require` on the command line instead of editing your main script. You can start by copying the default config file and modifying it to suit your needs. The `GCLOUD_DIAGNOSTICS_CONFIG` environment variable should point to your configuration file.
 
@@ -142,7 +142,7 @@ For any of the web frameworks listed above (`express`, `hapi`, and `restify`), a
 The API is exposed by the `agent` returned by a call to `start`:
 
 ```javascript
-  var agent = require('@google/cloud-trace').start();
+  var agent = require('@google/cloud-trace')().startAgent();
 ```
 
 For child spans, you can either use the `startSpan` and `endSpan` API, or use the `runInSpan` function that uses a callback-style. For root spans, you must use `runInRootSpan`.
--- a/index.js
+++ b/index.js
@@ -82,25 +82,6 @@ var initConfig = function(projectConfig) {
   return config;
 };
 
-function _printStack(name) {
-  /*
-  var stack = require('callsite');
-  console.log('Invoking ' + name + ' at:');
-  stack().forEach(function(site){
-    var file = site.getFileName();
-    if (file.indexOf('cloud-trace-nodejs') !== -1 &&
-          file.indexOf('node_modules') === -1 &&
-            !file.endsWith('index.js')){
-      console.log('  %s:%d',
-        //site.getFunctionName() || 'anonymous',
-        site.getFileName(),
-        site.getLineNumber());
-    }
-  });
-  console.log();
-  */
-}
-
 /**
  * The singleton public agent. This is the public API of the module.
  */
@@ -202,8 +183,6 @@ var publicAgent = {
   },
 
   stop: function() {
-    _printStack('publicAgent.stop with isActive=' + this.isActive());
-
     if (this.isActive()) {
       agent.stop();
       agent = phantomTraceAgent;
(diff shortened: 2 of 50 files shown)


index.js
```diff
@@ -198,9 +198,23 @@ var publicAgent = {
...
+  publicAgent.startAgent(projectConfig);
+  return publicAgent;
+};
+
+module.exports = global._google_trace_agent = Trace;
 
 // If the module was --require'd from the command line, start the agent.
 if (module.parent && module.parent.id === 'internal/preload') {
-  module.exports.start();
+  module.exports().startAgent();
```
kjin: Why do we need to call a function in order to get an object through which the agent can be started?

DominicKramer: The module now exports a constructor that is used to make `Trace` objects.  Each `Trace` object then has a single method, `startAgent()`, that is used to start the agent (and the agent is returned by `startAgent()`).

DominicKramer:
PTAL

test/standalone/test-trace-uncaught-exception.js
```diff
@@ -62,7 +67,9 @@ describe('tracewriter publishing', function() {
       }, 20);
     });
     process.nextTick(function() {
-      var privateAgent = agent.start({
+      // TODO: (DK) Check if stop() should be called here
+      agent.stop();
```
matthewloring: Does this test fail without this stop?

index.js
```diff
@@ -198,9 +198,24 @@ var publicAgent = {
   private_: function() { return agent; }
 };
 
-module.exports = global._google_trace_agent = publicAgent;
+/**
+ * @param {object} options - These are currently ignored
```
matthewloring: Could we add a todo here so the comment can be updated when we refactor configuration.

test/hooks/test-trace-grpc.js
```diff
@@ -16,7 +16,8 @@
 'use strict';
 
 var common = require('./common.js');
-require('../..').private_().config_.enhancedDatabaseReporting = true;
+var trace = require('../..')();
+trace.startAgent().private_().config_.enhancedDatabaseReporting = true;
```
matthewloring: Can this be passed through the config object to `startAgent`? I know this was mirroring previous style but passing it the official way would be better.

test/hooks/test-trace-mysql.js
```diff
@@ -17,7 +17,8 @@
 
 var common = require('./common.js');
 var traceLabels = require('../../src/trace-labels.js');
-require('../..').private_().config_.enhancedDatabaseReporting = true;
+var agent = require('../..')().startAgent();
+agent.private_().config_.enhancedDatabaseReporting = true;
```
matthewloring: And here.

index.js
DominicKramer: Done

index.js
```diff
@@ -114,7 +114,7 @@ var publicAgent = {
     return agent.addTransactionLabel(key, value);
   },
 
-  start: function(projectConfig) {
+  startAgent: function(projectConfig) {
```
matthewloring: It seems odd for the constructor to return an object with a `startAgent` property that in turn produces an object with another `startAgent` property.

index.js
```diff
@@ -198,9 +198,25 @@ var publicAgent = {
...
+/**
+ * @param {object} options - These are currently ignored
+ */
+function Trace(options) {
+  if (!(this instanceof Trace)) {
+    return new Trace(options);
+  }
+}
+
+Trace.prototype.startAgent = function(projectConfig) {
```
matthewloring: @ofrobots Is the preferred mechanism to retrieve an already initialized agent to call `startAgent` again? I believe we originally introduced the `get` function above to avoid this. Could the same be done here?

test/hooks/test-trace-grpc.js
DominicKramer: Yes, and I have changed it.

test/standalone/test-trace-uncaught-exception.js
DominicKramer: The test fails without the stop, but the cause has been found and resolved.

src/trace-agent.js
```diff
@@ -35,6 +35,7 @@ var traceAgent;
 function TraceAgent(config, logger) {
   this.config_ = config;
   this.logger = logger;
+  this.running = true;
```
matthewloring: Could we add a TODO to remove this?

test/hooks/common.js
```diff
@@ -42,24 +40,32 @@ var SERVER_CERT = fs.readFileSync(path.join(__dirname, 'fixtures', 'cert.pem'));
...
-function getTraces() {
+function getTraces(agent) {
   return agent.traceWriter.buffer_.map(JSON.parse);
 }
 
-function getMatchingSpan(predicate) {
-  var spans = getMatchingSpans(predicate);
+function getMatchingSpan(agent, predicate) {
+  if (arguments.length !== 2){
+    throw new Error('getMatchingSpan() not called with 2 arguments');
```
matthewloring: Is this still needed or was it just used during the refactor?

README.md
```diff
@@ -33,7 +33,7 @@ This module provides Stackdriver Trace support for Node.js applications. [Stackd
 
 3. Include and start the library *as the very first action in your application*:
 
-        require('@google/cloud-trace').start();
+        require('@google/cloud-trace')().startAgent();
```
ofrobots: nit: it would be better if the examples always save the return value from the constructor into a variable to make it obvious that the `startAgent` call is available once an API object has been constructed.

DominicKramer:
I've created PR [361](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/361) that consolidates all of the changes in this PR to make future merging to master easier.  I will close this PR and address the comments on this PR in the new PR.

PR #338, opened 2016-12-28
Removed dependency on cloud-diagnostics-common

kjin:
Removed all dependencies on cloud-diagnostics-common, replacing with equivalent functionality in cloud-common.

### DIFF
--- a/config.js
+++ b/config.js
@@ -90,14 +90,14 @@ module.exports = {
     // projects.
     ignoreContextHeader: false,
 
+    // The contents of a key file. If this field is set, its contents will be
+    // used for authentication instead of your application default credentials.
+    credentials: null,
+
     // A path to a key file relative to the current working directory. If this
     // field is set, the contents of the pointed file will be used for
     // authentication instead of your application default credentials.
-    keyFilename: null,
-
-    // The contents of a key file. If this field is set, its contents will be
-    // used for authentication instead of your application default credentials.
-    // If keyFilename is also set, the value of credentials will be ignored.
-    credentials: null
+    // If credentials is also set, the value of keyFilename will be ignored.
+    keyFilename: null
   }
 };
--- a/index.js
+++ b/index.js
@@ -23,7 +23,8 @@ var filesLoadedBeforeTrace = Object.keys(require.cache);
 require('continuation-local-storage');
 
 var SpanData = require('./src/span-data.js');
-var common = require('@google/cloud-diagnostics-common');
+var common = require('@google-cloud/common');
+var gcpMetadata = require('gcp-metadata');
 var semver = require('semver');
 var constants = require('./src/constants.js');
 var util = require('./src/util.js');
@@ -125,7 +126,18 @@ var publicAgent = {
     if (!config.enabled) {
       return this;
     }
-    var logger = common.logger.create(config.logLevel, '@google/cloud-trace');
+
+    var logLevel = config.logLevel || 2;
+    if (logLevel < 0) {
+      logLevel = 0;
+    } else if (logLevel > 4) {
+      logLevel = 4;
+    }
+    var logger = common.logger({
+      level: common.logger.LEVELS[logLevel],
+      tag: '@google/cloud-trace'
+    });
+
     if (!semver.satisfies(process.versions.node, '>=0.12')) {
       logger.error('Tracing is only supported on Node versions >=0.12');
       return this;
@@ -153,7 +165,10 @@ var publicAgent = {
     if (typeof config.projectId === 'undefined') {
       // Queue the work to acquire the projectNumber (potentially from the
       // network.)
-      common.utils.getProjectNumber(headers, function(err, project) {
+      gcpMetadata.project({
+        property: 'project-id',
+        headers: headers
+      }, function(err, response, projectId) {
         if (err) {
           // Fatal error. Disable the agent.
           logger.error('Unable to acquire the project number from metadata ' +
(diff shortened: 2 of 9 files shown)


index.js
```diff
@@ -125,7 +126,18 @@ var publicAgent = {
     if (!config.enabled) {
       return this;
     }
-    var logger = common.logger.create(config.logLevel, '@google/cloud-trace');
+
+    var logLevel = config.logLevel || 2;
```
ofrobots: I think `initConfig` and config.js guarantee that config.logLevel will be defined by this point. At any rate, we shouldn't duplicate the default logLevel that is also in the config.

Aside: should the default logLevel in config.js continue to be an integer, or should it be `'error'`.

index.js
```diff
@@ -125,7 +126,18 @@ var publicAgent = {
...
+    if (logLevel < 0) {
```
ofrobots: We have a hazard here if `config.logLevel` came from the environment. Numbers would show up as strings. That's not fatal however since the expression `'4' < 0` does evaluates correctly due to type coersion `false`.

However `'monkey' < 0` is a valid comparison too that evaluates to `false`. In other words, if the user had an invalid log level, we disable logging for them.

index.js
```diff
@@ -125,7 +126,18 @@ var publicAgent = {
...
+      logLevel = 0;
+    } else if (logLevel > 4) {
```
ofrobots: hard-coded constant

index.js
kjin: Is the specification of `logLevel` is not baked in? If not, I think having `logLevel` accept the name of the level makes more sense and I'd be all for it - though it would involve changing across the other agents as well to preserve uniformity.

ofrobots: baked-in where?

kjin: As in, whether it's set in stone that it should be a number rather than the log level itself.

kjin:
Addressed inconsistencies in log levels here: https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1900

index.js
```diff
@@ -74,7 +75,11 @@ var initConfig = function(projectConfig) {
   util._extend(config, require('./config.js').trace);
   util._extend(config, projectConfig);
   if (process.env.hasOwnProperty('GCLOUD_TRACE_LOGLEVEL')) {
-    config.logLevel = process.env.GCLOUD_TRACE_LOGLEVEL;
+    try {
+      config.logLevel = parseInt(process.env.GCLOUD_TRACE_LOGLEVEL);
+    } catch (e) {
+      // Just use the original log level if GCLOUD_TRACE_LOGLEVEL is malformed.
```
AdriVanHoudt: why not print out some sort of warning because this is probably not what the user wants right?

kjin: Makes sense - though the logger isn't initialized at this point (since it depends on the value of `config.logLevel`).

Do you think it'd make sense to defer printing of the warning until after the logger is initialized?

AdriVanHoudt: Not familiar with internals but makes sense to defer when the logger isn't there.
I assume this is not breaking anything just failsafe. So just letting the user know (even a bit later) seems fine

index.js
```diff
@@ -121,11 +108,47 @@ var publicAgent = {
...
+    var logLevelValid = true;
+
+    // Initialize config object
+    var util = require('util');
+    var config = {};
+    util._extend(config, require('./config.js').trace);
+    util._extend(config, projectConfig);
+    if (process.env.hasOwnProperty('GCLOUD_TRACE_LOGLEVEL')) {
+      if (process.env.GCLOUD_TRACE_LOGLEVEL.match(/^\d+$/)) {
+        config.logLevel = parseInt(process.env.GCLOUD_TRACE_LOGLEVEL);
```
ofrobots: nit: always specify a radix for parseInt.

index.js
```diff
@@ -121,11 +108,47 @@ var publicAgent = {
...
-    var config = initConfig(projectConfig);
+    var logLevelValid = true;
+
+    // Initialize config object
+    var util = require('util');
+    var config = {};
+    util._extend(config, require('./config.js').trace);
+    util._extend(config, projectConfig);
+    if (process.env.hasOwnProperty('GCLOUD_TRACE_LOGLEVEL')) {
+      if (process.env.GCLOUD_TRACE_LOGLEVEL.match(/^\d+$/)) {
```
ofrobots: Could you avoid this by doing `parseInt` and checking if the result is `NaN`?

index.js
```diff
@@ -121,11 +109,46 @@ var publicAgent = {
...
+      logLevelValid = false;
+    }
+    var logger = common.logger({
+      level: common.logger.LEVELS[logLevel],
+      tag: '@google/cloud-trace'
+    });
+    if (!logLevelValid) {
+      var desiredLogLevel = process.env.GCLOUD_TRACE_LOGLEVEL ||
+        config.logLevel;
+      logger.warn('Provided log level ' + desiredLogLevel + ' is invalid, ' +
```
ofrobots: I think you can use `console.error` here to avoid issues with the default logLevel not allowing `warn` to actually be printed. This would be a valid use of `console.error`.

This should also simplify the logic here as you wouldn't need to carry around the `logLevelValid` flag and you could print the error as you encounter it.

It would be good to get this code back into a initConfig function.

config.js
```diff
@@ -90,14 +90,14 @@ module.exports = {
     // projects.
     ignoreContextHeader: false,
 
+    // The contents of a key file. If this field is set, its contents will be
+    // used for authentication instead of your application default credentials.
+    credentials: null,
+
```
ofrobots: Why is this change here? It don't seem to to have anything to do with the rest of the changes in this PR.

kjin: The authentication library used in `cloud-common` checks `credentials` and `keyFilename` in reverse order to ours.

ofrobots: Ah, I missed that part. Thanks for clarifying.

index.js
kjin: Done

index.js
kjin: Done.

index.js
kjin: OK, I've used console.error instead.

kjin:
Appveyor test failures seem to be unrelated to these changes.

PR #343, opened 2017-01-12
Ensure agent is always set for mysql/grpc

matthewloring:
If either of these modules switch to lazily importing any of our patched
files, we would fail to set the agent in some cases.


matthewloring:
PTAL

kjin:
I'm not sure I understand what the impact of this change is. It seems like now `agent` is set only if it hasn't been set before, but does this necessarily make it any more certain that it's actually set?

matthewloring:
None of the code in these hooks can execute without a call to one of the `patch` functions. This change ensures that the first patch function called always sets `agent` as opposed to assuming that all `patch` functions are called before any other code in the hook file runs.

PR #344, opened 2017-01-13
Trace Plugin API

kjin:
This PR is a summary of changes so far for the Trace Plugin API. It's at the moment by no means complete.

It currently has an "external" express plugin.

Since transactions have to be stored per-continuation, there is some intersection in using CLS to store both root context and transaction objects.

### DIFF
--- a/src/cls.js
+++ b/src/cls.js
@@ -53,6 +53,17 @@ module.exports = {
 
   setRootContext: function setRootContext(rootContext) {
     getNamespace().set('root', rootContext);
+  },
+
+  getTransaction: function getTransaction() {
+    if (getNamespace() && getNamespace().get('transaction')) {
+      return getNamespace().get('transaction');
+    }
+    return null;
+  },
+
+  setTransaction: function setTransaction(transaction) {
+    getNamespace().set('transaction', transaction);
   }
 };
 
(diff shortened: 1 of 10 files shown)


src/cls.js
```diff
@@ -53,6 +53,17 @@ module.exports = {
 
   setRootContext: function setRootContext(rootContext) {
     getNamespace().set('root', rootContext);
+  },
+
+  getTransaction: function getTransaction() {
```
matthewloring: It would be good to unify the 'transaction' and 'root' as there is a 1-1 relationship between them.

src/trace-plugin-interface.js
```diff
@@ -0,0 +1,123 @@
...
+  if (header) {
+    context = this.agent.parseContextFromHeader(header);
+  }
+  context = context || {};
+  if (!this.agent.shouldTrace(url, context.options)) {
+    return null;
+  }
+  var transaction = new Transaction(this.agent, context);
+  cls.getNamespace().run(function() {
+    cls.setTransaction(transaction);
```
matthewloring: This transaction will only be set in the scope of the function passed to run. We will need to structure this differently for it to be retrievable by child spans.

kjin:
See #356

PR #345, opened 2017-01-13
Limit span names/labels to service limits

matthewloring:


matthewloring:
PTAL.

src/trace-span.js
```diff
@@ -37,7 +37,13 @@ function TraceSpan(name, spanId, parentSpanId) {
  * @param {string} value The new value of the label.
  */
 TraceSpan.prototype.setLabel = function(key, value) {
-  this.labels[key] = '' + value;
+  var val = '' + value;
+  // The official limit imposed by the service is 16k per label value.
+  // We truncate smaller than that to keep requests small.
+  if (val.length > 5000) {
```
ofrobots: It would be better to make this an (internal) config property rather than a hard-coded constant.

matthewloring: Moved to an external constant.

src/constants.js
```diff
@@ -24,5 +24,15 @@ module.exports = {
...
+
+  /** See: cloud.google.com/trace/api/reference/rest/v1/projects.traces for limits. */
+  /** Maximum size of a span name in bytes. */
+  TRACE_SERVICE_SPAN_NAME_LIMIT: 128,
+
+  /** Maximum size of a label key in bytes. */
+  TRACE_SERVICE_LABEL_KEY_LIMIT: 128,
+
+  /** Maximum size of a label value in bytes. */
+  TRACE_SERVICE_LABEL_VALUE_LIMIT: 16 * 1000,
```
ofrobots: The actual limit is 16KiB, you have 16KB here.

src/trace-span.js
```diff
@@ -16,12 +16,18 @@
...
 
+var constants = require('./constants.js');
+
 /**
  * Creates a trace span object.
  * @constructor
  */
 function TraceSpan(name, spanId, parentSpanId) {
-  this.name = name;
+  if (name.length > constants.TRACE_SERVICE_SPAN_NAME_LIMIT) {
```
ofrobots: `name.length` is the number of unicode code units in the string `name` not bytes.

Can you add a testcase where the code units are < 128, but the byte length is greater?

src/trace-span.js
```diff
@@ -37,7 +43,14 @@ function TraceSpan(name, spanId, parentSpanId) {
  * @param {string} value The new value of the label.
  */
 TraceSpan.prototype.setLabel = function(key, value) {
-  this.labels[key] = '' + value;
+  if (key.length > constants.TRACE_SERVICE_LABEL_KEY_LIMIT) {
```
ofrobots: likewise

src/trace-span.js
matthewloring: Done.

src/constants.js
matthewloring: Done.

src/trace-span.js
```diff
@@ -17,13 +17,26 @@
 'use strict';
 
 var util = require('util');
+var constants = require('./constants.js');
+
+function truncate(string, length) {
```
ofrobots: You do need a big comment here describing the rationale for this. Also, what are the semantics of this function, it gatuantees that the returns string will have no more than `length` bytes, or does it guarantee that it will have less than `length` bytes?

src/constants.js
```diff
@@ -24,5 +24,15 @@ module.exports = {
   TRACE_AGENT_REQUEST_HEADER: 'x-cloud-trace-agent-request',
 
   /** @const {number} bitmask to determine whether trace is enabled in trace options. */
-  TRACE_OPTIONS_TRACE_ENABLED: 1 << 0
+  TRACE_OPTIONS_TRACE_ENABLED: 1 << 0,
+
+  /** See: cloud.google.com/trace/api/reference/rest/v1/projects.traces for limits. */
+  /** Maximum size of a span name in bytes. */
```
ofrobots: This comment is now wrong. `Size of a span name (in bytes) must be less than this limit`.

src/trace-span.js
matthewloring: Done.

src/constants.js
matthewloring: Done.

PR #346, opened 2017-01-13
Display contents of label values of object type

matthewloring:


matthewloring:
PTAL.

src/trace-span.js
```diff
@@ -37,7 +39,7 @@ function TraceSpan(name, spanId, parentSpanId) {
  * @param {string} value The new value of the label.
  */
 TraceSpan.prototype.setLabel = function(key, value) {
-  this.labels[key] = '' + value;
+  this.labels[key] = typeof value === 'object' ? util.inspect(value) : '' + value;
```
ofrobots: Interesting. It is curious why we are doing a `toString` at this point in the first place. 

Later on at https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/src/trace-writer.js#L196 we do `JSON.stringify` which would have done the right thing.

matthewloring: `JSON.stringify(5) === "5"` `JSON.stringify('5') === "\"5\""` It is my understanding that the API will reject these if they do not decode to string values.

PR #350, opened 2017-01-19
Fix gRPC Server root span reference issue

kjin:
Calling gRPC server functions would clobber references to root span objects (causing them to either never end, or end at the wrong time) because they are stored in the scope of the patching function (`wrapUnary`, `wrapClientStream`, etc.).

This change fixes that by putting a reference to the root span in the scope of the `func` handler, so a root span can exist per gRPC call.

In addition, unit tests were added/modified:
- to ensure that interleaved gRPC call traces don't interfere with one another.
- to ensure that context is being propagated correctly to call handlers, by creating child spans in these handlers.
- to ensure that gRPC server call traces include the correct subset of stack frames.

### DIFF
--- a/src/hooks/userspace/hook-grpc.js
+++ b/src/hooks/userspace/hook-grpc.js
@@ -164,6 +164,7 @@ function wrapUnary(namespace, handlerSet, requestName) {
     return function funcTrace(call, callback) {
       var that = this;
       var args = arguments;
+      // Running in the namespace here propagates context to func.
       return namespace.runAndReturn(function() {
         var rootContext = startRootSpanForRequest(requestName);
         if (agent.config_.enhancedDatabaseReporting) {
@@ -172,7 +173,9 @@ function wrapUnary(namespace, handlerSet, requestName) {
         if (agent.config_.enhancedDatabaseReporting) {
           rootContext.addLabel('argument', JSON.stringify(call.request));
         }
-        // arguments[1] is the callback
+        // args[1] is the callback.
+        // Here, we patch the callback so that the span is ended immediately
+        // beforehand.
         args[1] = function (err, result, trailer, flags) {
           if (agent.config_.enhancedDatabaseReporting) {
             if (err) {
@@ -205,6 +208,7 @@ function wrapServerStream(namespace, handlerSet, requestName) {
     return function funcTrace(stream) {
       var that = this;
       var args = arguments;
+      // Running in the namespace here propagates context to func.
       return namespace.runAndReturn(function() {
         var rootContext = startRootSpanForRequest(requestName);
         if (agent.config_.enhancedDatabaseReporting) {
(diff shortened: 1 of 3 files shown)


kjin:

src/hooks/userspace/hook-grpc.js
```diff
@@ -160,58 +160,38 @@ function sendMetadataWrapper(rootContext) {
...
         if (agent.config_.enhancedDatabaseReporting) {
-          if (err) {
-            rootContext.addLabel('error', err);
-          } else {
-            rootContext.addLabel('result', JSON.stringify(result));
-          }
-          if (trailer) {
-            rootContext.addLabel('trailing_metadata', JSON.stringify(trailer.getMap()));
-          }
+          shimmer.wrap(call, 'sendMetadata', sendMetadataWrapper(rootContext));
```
matthewloring: Is `call` a unique object per request or is it shared?

matthewloring:
This looks good on my end but it would be good to get one more pair of eyes at least on the test logic due to the complexity.

src/hooks/userspace/hook-grpc.js
kjin: It's per request; specifically, it's passed into the request handler and contains information about the request (including the incoming request data, namely `call.request`).

src/hooks/userspace/hook-grpc.js
```diff
@@ -160,58 +160,38 @@ function sendMetadataWrapper(rootContext) {
...
-  });
-  shimmer.wrap(handlerSet, 'serialize', function (serialize) {
-    return function serializeTrace() {
-      var serializeResult = serialize.apply(this, arguments);
-      endRootSpanForRequest(rootContext);
-      return serializeResult;
+        // args[1] is the callback.
+        // Here, we patch the callback so that the span is ended immediately
+        // beforehand.
+        args[1] = function (err, result, trailer, flags) {
```
cristiancavalli: Disclaimer: personal opinion
Overwriting arguments make me nervous one could achieve the same result:

```
var applyArgs = [].concat(args);
applyArgs[1]  = function () {...}
```

src/hooks/userspace/hook-grpc.js
```diff
@@ -224,46 +204,48 @@ function wrapUnary(namespace, handlerSet, requestName) {
...
+        // event handlers.
+        stream.on('finish', function () {
+          // End the span unless there is an error. (If there is, the span will
+          // be ended in the error event handler. This is to ensure that the
+          // 'error' label is applied.)
+          if (stream.status.code === 0) {
+            endSpan();
+          }
+        });
+        stream.on('error', function (err) {
```
cristiancavalli: `on` actually returns the instance for chaining so one could do:

`eventEmitter.on(...).on(...);`

src/hooks/userspace/hook-grpc.js
kjin: Sounds good, will do

src/hooks/userspace/hook-grpc.js
kjin: Did not know this - but I think I think I would prefer the slight clarity of two `stream.on`s to the slight succinctness of having one.

PR #352, opened 2017-01-19
ignore labels unless it is an object

ofrobots:


package.json
```diff
@@ -40,6 +40,7 @@
     "@google-cloud/common": "^0.11.0",
     "continuation-local-storage": "^3.1.4",
     "gcp-metadata": "^0.1.0",
+    "is": "^3.2.0",
```
matthewloring: We `lodash` for the `isObject` check in other libraries and `is` uses a different `isObject` check than `lodash`. It would be good to have only one notion of what an object is.

test/standalone/test-index.js
```diff
@@ -164,6 +164,36 @@ describe('index.js', function() {
...
+      agent.start();
+      cls.getNamespace().run(function() {
+        agent.private_().createRootSpanData('root', 1, 2);
+        var stringWithProperty = 'foo';
+        var spanData = agent.startSpan('sub', 'foo');
+        agent.endSpan(spanData);
+        var traceSpan = spanData.span;
+        assert.equal(traceSpan.name, 'sub');
+        assert.ok(traceSpan.labels);
+        assert.strictEqual(traceSpan.labels[0], undefined);
```
matthewloring: Is there a reason this isn't a length check? I think the first label being `undefined` doesn't strictly capture the correctness condition.

test/standalone/test-index.js
```diff
@@ -164,6 +164,36 @@ describe('index.js', function() {
...
+        assert.ok(traceSpan.labels);
+        assert.equal(traceSpan.labels.test1, 'value');
+      });
+    });
+
+    it('should ignore non-object labels', function() {
+      agent.start();
+      cls.getNamespace().run(function() {
+        agent.private_().createRootSpanData('root', 1, 2);
+        var stringWithProperty = 'foo';
```
matthewloring: Is this used?

test/standalone/test-index.js
ofrobots: Length check would be quite brittle. By default we include a stack-trace property on labels, but checking against 1 would put a tight dependency on the default labels that the agent adds. I do also want a stronger check here, but haven't figured out a better one yet. Suggestions welcome.

ofrobots: BTW, note that this code isn't checking for the 'first label'. `labels` isn't an array. This currently checks for a property named `'0'`.

matthewloring: Yup, I missed that fact on the first read through. Maybe we could add a short comment about the previous failure mode?

test/standalone/test-index.js
ofrobots: fixed already

package.json
ofrobots: I think we should use `is` consistently everywhere as that is what is used on `google-code-node` as well. This module doesn't use `lodash.isObject` anywhere.

test/standalone/test-index.js
ofrobots: Okay, I think tight coupling is still better than the weak check here. Changed.

package.json
matthewloring: I think it was used in errors. If `is` is the best way going forward we should switch things over there.

AdriVanHoudt: Also note that `is.object() !== _.isObject()`, they will result in different results. e.g. `is.object([]) === false` while `_.isObject([]) === true`

ofrobots: @AdriVanHoudt Indeed – it was my intention to not accept arrays as a valid 'bag of properties'. The `is.object` semantics were needed here.

PR #356, opened 2017-01-21
Trace Plugin API

kjin:
Here's the Trace Plugin API with a more sensible commit order than #344.

The changes in `src/hooks/index.js` are temporary, pending the conversion of hooks to plugins.


src/trace-agent.js
```diff
@@ -35,6 +35,7 @@ var traceAgent;
 function TraceAgent(config, logger) {
   this.config_ = config;
   this.logger = logger;
+  this.plugins = config ? config.plugins : {};
```
DominicKramer: Should this line be:
```
this.plugins = config && config.plugins ? config.plugins : {};
```
?  Otherwise, its possible for `this.plugins` to still be `undefined` which will cause `src/hooks.js#activate` to fail.

kjin: Yep. Good catch

test/fixtures/plugin-express.js
```diff
@@ -0,0 +1,76 @@
+'use strict';
+var TraceLabels = require('../../src/trace-labels.js');
+var shimmer = require('shimmer');
+var semver = require('semver');
+var methods = require('methods').concat('use', 'route', 'param', 'all');
+
+var SUPPORTED_VERSIONS = '4.x';
+
+module.exports = function(version, api) {
+  if (!semver.satisfies(version, SUPPORTED_VERSIONS)) {
```
matthewloring: This plugin api work is an opportunity to change the flow of these functions if you see room for improvement. All plugins should be doing a supported versions check, maybe we could accept the plugins supported versions and do the check inside the agent so that every plugin doesn't have to depend on `semver` (and is guaranteed to do the check).

test/fixtures/plugin-mongodb-core.js
```diff
@@ -0,0 +1,125 @@
...
+function wrapCallback(transaction, done) {
+  return function(err, res) {
+    if (api.config.enhancedDatabaseReporting) {
+      if (err) {
+        // Errors may contain sensitive query parameters.
+        transaction.addLabel('mongoError', err);
+      }
+      if (res) {
+        var result = res.result ? res.result : res;
+        transaction.addLabel('results', traceUtil.stringifyPrefix(
```
matthewloring: We should figure out whether/how to expose this through the API. We recently added truncation of label values to `addLabel`. I don't think we need two mechanisms for this stacked on top of each other. Maybe we can drop `databaseResultReportingSize` and summarize using `stringPrefix` and the server label value limit instead of truncating as done today. This would also simplify the plugin API.

test/test-plugins.js
```diff
@@ -0,0 +1,73 @@
+'use strict';
```
matthewloring: It was useful to see the test instrumentation along side the api but I would prefer holding off on submitting these test files as they should only be need while both patching mechanisms are in place.

src/trace-plugin-interface.js
```diff
@@ -0,0 +1,242 @@
...
+ */
+Transaction.prototype.runInChildSpan = function(options, fn) {
+  var that = this;
+  options = options || {};
+  var childContext = that.agent.startSpan(options.name, {},
+    options.skipFrames ? options.skipFrames + 1 : 1);
+  // If the options object passed in has the setHeader field set,
+  // use it to set trace metadata in an outgoing request.
+  if (typeof(options.setHeader) === 'function') {
+    var outgoingTraceContext = that.agent.generateTraceContext(childContext, true);
```
cristiancavalli: If you're not using this variable any where else can you inline its invocation value?
ex.

```JS
options.setHeader('x-cloud-trace-context',
  that.agent.generateTraceContext(childContext, true));
```

src/trace-plugin-interface.js
```diff
@@ -0,0 +1,242 @@
...
+ * @returns A new Transaction object, or null if the trace agent's policy has
+ * disabled tracing for the given set of options.
+ */
+Plugin.prototype.createTransaction = function(options) {
+  options = options || {};
+  var that = this;
+  // If the options object passed in has the getHeader field set,
+  // try to retrieve the header field containing incoming trace metadata.
+  var incomingTraceContext;
+  if (typeof(options.getHeader) === 'function') {
```
cristiancavalli: Since you're doing this comparison a fair amount I would suggest using `is`

ex. `is.fn(options.getHeader)`

test/fixtures/plugin-grpc.js
```diff
@@ -0,0 +1,414 @@
...
+    // register(n, h, s, d, m) is called in addService once for each service method.
+    // Its role is to assign the serialize, deserialize, and user logic handlers
+    // for each exposed service method. Here, we wrap these functions depending on the
+    // method type.
+    var result = register.apply(this, arguments);
+    var handlerSet = this.handlers[name];
+    var requestName = 'grpc:' + name;
+    // Proceed to wrap methods that are invoked when a gRPC service call is made.
+    // In every case, the function 'func' is the user-implemented handling function.
+    if (method_type === 'unary') {
```
cristiancavalli: This might be more concisely represented as a switch statement

```JS
switch (method_type) {
  case 'unary':
    wrapUnary(handlerSet, requestName);
    break;
  case 'server_stream':
    wrapServerStream(handlerSet, requestName);
    break;
  case 'client_stream':
    wrapClientStream(handlerSet, requestName);
    break;
  case 'bidi':
    wrapBidi(handlerSet, requestName);
    break;
  default:
    api.logger.warn('gRPC Server: Unrecognized method_type ' + method_type);
}

return result;
```

test/test-plugins.js
```diff
@@ -0,0 +1,73 @@
...
+    // Run tests used for express hook and make sure there are no failures
+    mocha.run(function(numFailures) {
+      assert(numFailures === 0);
+      setImmediate(done);
+    });
+  });
+
+  it('should make a gRPC plugin capable of running correctly', function(done) {
+    this.timeout(10000);
+    ['1'].forEach(function(version) {
```
cristiancavalli: This seems a little redundant since it's a constant singular entry.

src/hooks/index.js
```diff
@@ -123,6 +124,15 @@ function activate(agent) {
 
   logger = agent.logger;
 
+  // Plugin stuff
+  var api = new Plugin(agent);
+  Object.keys(agent.plugins).forEach(function(pluginName) {
+    if (toInstrument[pluginName]) {
```
DominicKramer: With this check, a plugin will only be loaded if its name is already listed in `toInstrument`, but new plugins will most likely be for modules not listed in `toInstrument`.

test/fixtures/plugin-mongodb-core.js
DominicKramer: I agree.  This method seems like it will be used by a lot of plugins.  Another approach could be to have a small set of core functionality in `trace-plugin-interface.js` and have another helper class that accepts an `api` object and has methods that perform common complex tasks using the `api` object.

This could allow the helper to be tested independent of the api's core functionality by mocking the api.

test/fixtures/plugin-express.js
kjin: Makes sense, it does seem boilerplate - I'm taking this to mean that plugins should be able to provide a field with that is passed to `semver.satisfies` internally.

kjin: What do you think about flipping paths and `patch/unpatch`? What I mean is instead of accessing the patch function through plugin['']['patch'], we get it with plugin['patch'][''] - so the plugin object has these top-level fields: `versions`, `patch`, `unpatch`.

test/fixtures/plugin-grpc.js
kjin: Sure thing, fixed here: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/357/files#diff-8a4a73c6b4ea6be6a9022b3e6846829bR385

The gRPC plugin will be re-written from `hook-grpc.js` as the linked PR is merged.

test/test-plugins.js
kjin: Yep - just an artifact of realizing that 0.13-0.15 aren't supported anymore.

src/hooks/index.js
kjin: Good catch - in fact enough logic has changed that I'm going to add a new file specifically for plugins. Change pending

src/trace-plugin-interface.js
```diff
@@ -0,0 +1,242 @@
...
+  this.context.close();
+};
+
+/**
+ * Runs the given function in a child span, passing it an object that
+ * exposes an interface for adding labels and closing the span.
+ * @param {object} options An object that specifies options for how the child
+ * span is created and propogated.
+ * @param {string} options.name The name to apply to the child span.
+ * @param {?function(string, string)} options.setHeader A function describing
```
kjin: I'm hoping for the latter; it's true that we only ever set one header field, but that header field name can only be exposed in one of two ways: either it's supplied as a constant (like api.TRACE_HEADER_NAME), or passed in as an argument to `(set)(get)Header`. So the two methods of setting the trace context header are:

```js
setTraceContextHeader(headerValue) {
  req.set(api.TRACE_HEADER_NAME, headerValue);
}
```

or

```js
setHeader(headerName, headerValue) {
  req.set(headerName, headerValue);
}
```

Between these I like the second because it seems like boilerplate to have to use a provided constant on an API in a function that is being provided to that same API. In addition, for servers with express-like request objects, you can pass in `req.set` and `req.get` directly for `(set)(get)Header` respectively, which is nice.

src/trace-plugin-interface.js
```diff
@@ -0,0 +1,242 @@
...
+ * once. If the incoming request should be traced, a root span will be created,
+ * and this function will be called with a Transaction object exposing functions
+ * operating on the root span; otherwise, it will be called without any
+ * arguments.
+ * @returns The return value of calling fn.
+ */
+Plugin.prototype.runInRootSpan = function(options, fn) {
+  var that = this;
+  return this.agent.namespace.runAndReturn(function() {
+    var oldSkipFrames = options.skipFrames;
```
kjin: Sorry, not sure what you mean here.

test/fixtures/plugin-express.js
```diff
@@ -0,0 +1,76 @@
+'use strict';
+var TraceLabels = require('../../src/trace-labels.js');
```
kjin: Done

test/fixtures/plugin-express.js
kjin: Done - using patch object sets

src/hooks/index.js
kjin: Added new file `trace-plugin-loader.js`, and this file was reverted to original.

test/test-plugins.js
kjin: OK - will leave out of PR merge.

src/trace-plugin-interface.js
kjin: Done

src/trace-plugin-interface.js
kjin: Done

test/fixtures/plugin-mongodb-core.js
kjin: @matthewloring Do we want to prefer `stringifyPrefix` or `truncate`? If we always prefer one over another then the other should be removed.

src/trace-plugin-interface.js
```diff
@@ -0,0 +1,246 @@
...
+ * for arbitrary modules.
+ */
+
+/**
+ * An object that is associated with a single child span. It exposes
+ * functions for adding labels to or closing the associated span.
+ */
+function ChildSpan(agent, context) {
+  this.agent = agent;
+  this.context = context;
```
matthewloring: Could we rename this to `privateSpan` or maybe `span`. Context is a pretty overloaded term. (Span is too but maybe it's better?)

src/trace-plugin-interface.js
```diff
@@ -0,0 +1,246 @@
...
+};
+
+/**
+ * An object that is associated with a single root span. It exposes
+ * functions for adding labels to or closing the associated span.
+ */
+function Transaction(agent, context) {
+  this.agent = agent;
+  this.context = context;
+  this.closed = false;
```
matthewloring: Is this used?

src/trace-plugin-interface.js
matthewloring: Sounds good, I'm happy with the current naming.

src/trace-plugin-interface.js
```diff
@@ -0,0 +1,246 @@
...
+ * @param {?number} options.skipFrames The number of stack frames to skip when
+ * collecting call stack information for the root span, starting from the top;
+ * this should be set to avoid including frames in the PluginAPI. Defaults to 0.
+ * @param {function(ChildSpan)} fn A function that will be called exactly
+ * once, with a ChildSpan object exposing an interface operating on the child
+ * span.
+ * @returns The return value of calling fn.
+ */
+Transaction.prototype.runInChildSpan = function(options, fn) {
+  var that = this;
```
matthewloring: Do we need to alias `this` in this context?

src/trace-plugin-interface.js
```diff
@@ -0,0 +1,246 @@
...
+  // If the options object passed in has the getHeader field set,
+  // try to retrieve the header field containing incoming trace metadata.
+  var incomingTraceContext;
+  if (is.fn(options.getHeader)) {
+    var header = options.getHeader('x-cloud-trace-context');
+    if (header) {
+      incomingTraceContext = that.agent.parseContextFromHeader(header);
+    }
+  }
+  incomingTraceContext = incomingTraceContext || {};
```
matthewloring: We can do this initialization at the declaration point above.

src/trace-plugin-interface.js
```diff
@@ -0,0 +1,246 @@
...
+      incomingTraceContext.options !== undefined) ?
+      incomingTraceContext.options : constants.TRACE_OPTIONS_TRACE_ENABLED;
+    outgoingTraceContext += (';o=' + outgoingHeaderOptions);
+    options.setHeader('x-cloud-trace-context', outgoingTraceContext);
+  }
+  return new Transaction(that.agent, rootContext);
+};
+
+PluginAPI.prototype.getTransaction = function() {
+  if (cls.getRootContext()) {
```
matthewloring: Does `getRootContext` return null if it isn't defined? Maybe we can get rid of this branch.

test/fixtures/plugin-mongodb-core.js
matthewloring: That's a good point. I'm fine leaving it as is and ripping one out as a follow on. I would be a fan of using `truncate` everywhere since it is unclear that `stringifyPrefix` actually results in more intuitive output.

src/trace-plugin-interface.js
kjin: Yes, it's a misnomer here.

I'll also make these private, ideally plugins should not access them other than in exceptional circumstances.

src/trace-plugin-interface.js
kjin: Nope, will unalias

src/trace-plugin-interface.js
kjin: It does return null, but there still needs to be a branch, since we don't want `getTransaction` to return a `Transaction` object if there isn't one.

src/trace-plugin-interface.js
kjin: This catches `incomingTraceContext` being null if from the return value of `parseContextFromHeader`.

src/trace-plugin-interface.js
kjin: Nope, removed

src/trace-plugin-interface.js
```diff
@@ -0,0 +1,244 @@
...
+ * and this function will be called with a Transaction object exposing functions
+ * operating on the root span; otherwise, it will be called without any
+ * arguments.
+ * @returns The return value of calling fn.
+ */
+PluginAPI.prototype.runInRootSpan = function(options, fn) {
+  var that = this;
+  return this.agent.namespace.runAndReturn(function() {
+    var oldSkipFrames = options.skipFrames;
+    options.skipFrames = options.skipFrames ? options.skipFrames + 2 : 2;
```
ofrobots: nit: It would be nice not to permute the user provided object, even temporarily.

src/trace-plugin-interface.js
```diff
@@ -0,0 +1,244 @@
...
+ * span is created and propogated. @see Transaction.prototype.runInChildSpan
+ * @param {function(?Transaction)} fn A function that will be called exactly
+ * once. @see Transaction.prototype.runInChildSpan
+ * @returns The return value of calling fn.
+ */
+PluginAPI.prototype.runInChildSpan = function(options, fn) {
+  var transaction = this.getTransaction();
+  if (transaction) {
+    var oldSkipFrames = options.skipFrames;
+    options.skipFrames = options.skipFrames ? options.skipFrames + 1 : 1;
```
ofrobots: nit: it would be better if we didn't permute the user-supplied object, even temporarily.

src/trace-agent.js
```diff
@@ -35,8 +36,10 @@ var traceAgent;
 function TraceAgent(config, logger) {
   this.config_ = config;
   this.logger = logger;
+  this.plugins = config && config.plugins ? config.plugins : {};
```
DominicKramer: If a user has an existing configuration that doesn't have a `plugins` entry in their config, using `config.plugins` could cause an error to be thrown when operating on `config.plugins`.

Do we want the agent to fail-fast or to gracefully handle `plugins` not being defined as specifying not to use any plugins?

src/trace-plugin-loader.js
```diff
@@ -0,0 +1,215 @@
...
+    var first = process._preload_modules[0];
+    if (first !== '@google/cloud-trace') {
+      logger.error('Tracing might not work as ' + first +
+            ' was loaded with --require before the trace agent was initialized.');
+    }
+  }
+}
+
+function activate(agent) {
+  logger = agent.logger;
```
kjin: I think the name `activate` does already imply this, but I'll add a check to ensure that activate is only called once. I suppose this is doubly important since we wouldn't want to double-wrap `Module._load`.

kjin: Actually - in most cases it doesn't hurt to log an error when something bad happens, but `activate` is only called from `startAgent` which itself can only be called once, so I think a code path where `activate` is called twice is already not possible. In this case does it make sense to log an error?

src/trace-agent.js
kjin: @DominicKramer I think we should not fail if `plugins` aren't defined. We don't want to prohibit custom trace spans in the absence of tracing well-known libraries.

kjin: @ofrobots I'll keep plugins a part of `config`. The only place that uses the field `this.plugins` is in the plugin loader itself, so I'll make a change there.

src/trace-plugin-interface.js
```diff
@@ -0,0 +1,244 @@
...
+var is = require('is');
+
+/**
+ * This file describes an interface for third-party PluginAPIs to enable tracing
+ * for arbitrary modules.
+ */
+
+/**
+ * An object that is associated with a single child span. It exposes
+ * functions for adding labels to or closing the associated span.
```
kjin: Sure, left them out initially because this function is not publicly available.

src/trace-plugin-loader.js
ofrobots: You're making a good case for an assertion.

src/trace-agent.js
ofrobots: We should normalize config when we received it from the user, and assume in the rest of the code that all expected properties are defined.

src/trace-plugin-interface.js
kjin: @ofrobots The hope is that `setHeader` and `getHeader` are always set in plugins when the concept of a header exists, i.o.w cross-service requests. So this applies to anything HTTP-related, including the HTTP core module itself. I agree that the documentation doesn't clearly detail exactly when or why this is necessary, so I'll try to come up with a better description.

The only example so far is this:

https://github.com/kjin/cloud-trace-nodejs/blob/b1e1ec4df89e8324de240375a493a53baa927bfb/test/fixtures/plugin-express.js#L20

src/trace-plugin-interface.js
kjin: I think the most correct solution here is to copy the options object entirely - though I worry about the performance impact. Alternatively, we could document the number of additional frames to skip when one calls this function, though that puts more burden on developers... any preference here? For now, I'll do the former.

ofrobots: Yes, I am concerned about performance too. Is there any way to pull the shouldTrace check up?

src/trace-plugin-interface.js
```diff
@@ -0,0 +1,244 @@
...
+
+/**
+ * PluginAPI constructor. Don't call directly - a PluginAPI object will be passed to
+ * PluginAPIs themselves
+ * TODO(kjin): Should be called something else
+ */
+function PluginAPI(agent) {
+  this.agent = agent;
+  this.logger = agent.logger;
+  this.config = {
```
kjin: Done - exposed `enhancedReportingEnabled()` on API that returns the value of `agent.config.enhancedDatabaseReporting`.

src/trace-plugin-interface.js
```diff
@@ -0,0 +1,244 @@
...
+ * @param {object} options An object that specifies options for how the child
+ * span is created and propogated.
+ * @param {string} options.name The name to apply to the child span.
+ * @param {?function(string, string)} options.setHeader A function describing
+ * how to set a header field with the given field name to a given value,
+ * respectively. If supplied, it will be called to set the header field in an
+ * outgoing request associated with this child span to the field name
+ * 'x-cloud-trace-context'.
+ * @param {?number} options.skipFrames The number of stack frames to skip when
+ * collecting call stack information for the root span, starting from the top;
```
kjin: Done

src/trace-plugin-interface.js
```diff
@@ -0,0 +1,244 @@
+'use strict';
```
kjin: Done

src/trace-agent.js
kjin: Seems like it is already being normalized.

src/trace-agent.js
```diff
@@ -18,6 +18,7 @@
 
 var cls = require('./cls');
 var hooks = require('./hooks/index.js');
+var plugins = require('./trace-plugin-loader.js');
```
kjin: Changed to `pluginLoader`.

src/trace-plugin-loader.js
```diff
@@ -0,0 +1,215 @@
...
+
+  // Plugin stuff
```
kjin: Elaborated to describe what this line is doing.

src/trace-plugin-interface.js
kjin: I've changed it so that this (and the equivalent function for child spans) call on a private function that accepts `skipFrames` separately. This allows us not to create/mutate an `options` object at all, in exchange for some clutter in the interface file (though this can be folded away in a separate file).

src/trace-plugin-loader.js
```diff
@@ -0,0 +1,235 @@
...
+function activate(agent) {
+  if (logger) {
+    logger.error('Plugins activated more than once.');
+  }
+  logger = agent.logger;
+
+  // Create a new object exposing functions to create trace spans and propagate
+  // context. This relies on functions currently exposed by the agent.
+  var api = new PluginAPI(agent);
+  var pluginConfig = agent.config().plugins;
```
evanlucas: so as a consumer of this new plugin api, I would have to add my module to be wrapped to the `plugins` object in the config? Would there also be a way to just get a reference to the `api` and manually call `api.runInRootSpan()`?

kjin: We'd definitely be able to expose the `api` object through the agent itself, in which case you could certainly manually patch your own module or application however you wanted, outside of the plugin loader system. The plugin loader is meant to make it easy for application developers to bring in plugins for modules they'd want to trace.

matthewloring: @evanlucas I believe one of your current concerns with the existing [custom tracing api](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs#custom-tracing-api) was that a new trace context was always created making it impossible to provide your own from another source. We will also be updating the the api of the trace agent to be consistent with this plugin api (which has this functionality).

evanlucas: Yes, that is my biggest concern. We have our own framework and my goal is to simply add the creation of the root context when we route initially without having to monkey patch anything. But yes, being able to correlate contexts would be a big win for us, so we are really excited about this. Thanks!

src/trace-plugin-interface.js
kjin: After discussing the role of the trace context header with Matt I'm now convinced that it's better to generalize the functionality of `setHeader` to `setTraceContext` since it's not necessarily true that trace context is only propagated over the network through HTTP headers. See 2629880.

I think it'd be best to discuss the importance of providing this optional parameter in the plugin implementation guide rather than the JSDoc.

ofrobots:
I think naming is important and neither setHeader nor setTraceContext leave
me fully satisfied.

Let's first talk about the incoming context. Currently we are asking the
user to provide a getHeader/getTraceContext functions that we immediately
call synchronously. I think it would be better if they passed in the
incoming context as a field on options. Yes, this would require us to
expose a constant for the header name we expect – I think that is totally
fine. Right now we are asking them to provide a function with a lot more
capability that we don't need.

Once you have created a Transaction object (again, synchronously), the
context property must be defined on that object. I think you should just
add a function/getter on Transaction to allow people to retrive the context
from the transaction.

kjin:
@ofrobots That looks to me as a more sensible way to go - I've changed it so that `options.traceContext` can contain the (serialized) trace context, while `(Transaction|ChildSpan)`.getTraceContext() extracts it in 269f1af.

kjin:
Coverage decrease is due to new files related to plugins not being tested; these tests will be added in a future PR.

src/trace-plugin-loader.js
```diff
@@ -0,0 +1,235 @@
...
+function resolvePluginPath(pluginFile) {
+  // If path doesn't start with '/', './', or '../',
+  // it must refer to a module in node_modules
+  if (!pluginFile.match(/^(\/|\.\/|\.\.\/)/)) {
+    return pluginFile;
+  } else {
+    if (path.isAbsolute(pluginFile)) {
+      return pluginFile;
+    } else {
+      var entryPoint = process.argv[1];
```
ofrobots: As discussed, it would be better to not support relative paths in the first version of this. Absolute paths or module names are fine as a starting point.

src/trace-plugin-loader.js
```diff
@@ -0,0 +1,235 @@
...
+    if (fs.existsSync(pjson)) {
+      return load(pjson).version;
+    }
+  }
+  return process.version;
+}
+
+function resolvePluginPath(pluginFile) {
+  // If path doesn't start with '/', './', or '../',
+  // it must refer to a module in node_modules
```
ofrobots: but I think this comment and test goes away if you get rid of relative path support.

PR #357, opened 2017-01-24
Follow-Up Fixes for gRPC

kjin:
* Slight `wrap`-function doc change to reflect previous changes
* Comments summarizing start and end span points
* Check `agent.shouldTrace` to determine whether span should actually be started
* Renamed `func` object to `serverMethod` internally (less confusion)
* No longer dealing with `arguments` - `apply` calls changed to `call` with actual arguments
* Collapse consecutive `if` statements with identical conditions
* `switch` instead of `if...else if` statement for picking `wrap` function

### DIFF
--- a/src/hooks/userspace/hook-grpc.js
+++ b/src/hooks/userspace/hook-grpc.js
@@ -144,8 +144,6 @@ function sendMetadataWrapper(rootContext) {
     return function sendMetadataTrace(responseMetadata) {
       if (rootContext) {
         rootContext.addLabel('metadata', JSON.stringify(responseMetadata.getMap()));
-      } else {
-        agent.logger.info('gRPC: No root context found in sendMetadata');
       }
       return sendMetadata.apply(this, arguments);
     };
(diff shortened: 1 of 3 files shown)


src/hooks/userspace/hook-grpc.js
```diff
@@ -144,8 +144,6 @@ function sendMetadataWrapper(rootContext) {
     return function sendMetadataTrace(responseMetadata) {
       if (rootContext) {
         rootContext.addLabel('metadata', JSON.stringify(responseMetadata.getMap()));
-      } else {
-        agent.logger.info('gRPC: No root context found in sendMetadata');
```
matthewloring: Why is this removed?

ofrobots:
Given that 'shouldTrace' sampling was a bugfix (the rest are mostly cosmetic, structure), would it be possible to add a test to make sure we do in fact have functional sampling that we don't regress in the future?

src/hooks/userspace/hook-grpc.js
kjin: `rootContext` cannot be null; its value here is always obtained directly from a call to `createRootSpanData`.

matthewloring: If this is the case, the if statement above should be removed too and a brief comment added.

kjin:
@ofrobots Added a test to show that `shouldTrace` is being called. Do you think I should also test this for the other hooks?

test/hooks/common.js
```diff
@@ -136,6 +147,10 @@ function createChildSpan(cb, duration) {
   };
 }
 
+function getAndResetShouldTraceArguments() {
```
matthewloring: Is this used?

kjin: Nope, will remove

PR #358, opened 2017-01-26
hapi versions in README.md were out of sync

nowen094:


matthewloring:
This has already landed on master I believe. Let me know if there is another concern here. Closing for now.

PR #361, opened 2017-02-01
The module now exports a constructor

DominicKramer:
Now this module exports a constructor.  In addition, the `start`
method has been renamed to `startAgent`, and the agent must be
stopped before it can be started again.

These changes were made to more closely align this module with
the other Google Cloud API Client Libraries via the
`google-cloud` modules in [1].

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ This module provides Stackdriver Trace support for Node.js applications. [Stackd
 
 3. Include and start the library *as the very first action in your application*:
 
-        require('@google/cloud-trace').start();
+        require('@google/cloud-trace')().startAgent();
 
   If you use `--require` in your start up command, make sure that the trace agent is --required first.
 
@@ -47,7 +47,7 @@ If you are running somewhere other than the Google Cloud Platform, see [running 
 
 See [the default configuration](config.js) for a list of possible configuration options. These options can be passed to the agent through the object argument to the start command shown above:
 
-        require('@google/cloud-trace').start({samplingRate: 500});
+        require('@google/cloud-trace')().startAgent({samplingRate: 500});
 
 Alternatively, you can provide configuration through a config file. This can be useful if you want to load our module using `--require` on the command line instead of editing your main script. You can start by copying the default config file and modifying it to suit your needs. The `GCLOUD_DIAGNOSTICS_CONFIG` environment variable should point to your configuration file.
 
@@ -142,7 +142,7 @@ For any of the web frameworks listed above (`express`, `hapi`, and `restify`), a
 The API is exposed by the `agent` returned by a call to `start`:
 
 ```javascript
-  var agent = require('@google/cloud-trace').start();
+  var agent = require('@google/cloud-trace')().startAgent();
 ```
 
 For child spans, you can either use the `startSpan` and `endSpan` API, or use the `runInSpan` function that uses a callback-style. For root spans, you must use `runInRootSpan`.
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ This module provides Stackdriver Trace support for Node.js applications. [Stackd
 
 3. Include and start the library *as the very first action in your application*:
 
-        require('@google/cloud-trace')().startAgent();
+        var agent = require('@google/cloud-trace')().startAgent();
 
   If you use `--require` in your start up command, make sure that the trace agent is --required first.
 
--- a/bin/run-test.sh
+++ b/bin/run-test.sh
@@ -39,27 +39,11 @@ function run {
 }
 
 # Run test/coverage
-run test/test-constants.js
-run test/test-span-data.js
+run `find test -name '*.js' -depth 1 -not -name test-trace-agent.js`
 run test/test-trace-agent.js
-run test/test-trace-policy.js
-run test/test-trace-span.js
-run test/test-trace.js
-run test/test-util.js
 
-run test/hooks/common.js
-run test/hooks/test-hooks-index.js
-run test/hooks/test-hooks-interop-mongo-express.js
-run test/hooks/test-trace-connect.js
+run `find test/hooks -name '*.js' -depth 1 -not -name test-trace-express.js`
 run test/hooks/test-trace-express.js
-run test/hooks/test-trace-grpc.js
-run test/hooks/test-trace-hapi.js
-run test/hooks/test-trace-http.js
-run test/hooks/test-trace-mongodb.js
-run test/hooks/test-trace-mongoose.js
-run test/hooks/test-trace-mysql.js
-run test/hooks/test-trace-redis.js
-run test/hooks/test-trace-restify.js
 
 for test in test/standalone/test-*.js ;
 do
--- a/bin/run-test.sh
+++ b/bin/run-test.sh
@@ -39,11 +39,8 @@ function run {
 }
 
 # Run test/coverage
-run `find test -maxdepth 1 -name '*.js' -not -name test-trace-agent.js`
-run test/test-trace-agent.js
-
-run test/hooks/*.js
-
+run test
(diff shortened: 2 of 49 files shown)


DominicKramer:
PTAL

I still need to address an issue where the tests fail if `test-hooks-interop-mongo-express.js` and `test-trace-express.js` are run together.  Similarly, I still need to address an issue where the tests fail if `test-trace-agent.js` and `test-span-data.js` are run together.

index.js
```diff
@@ -94,8 +94,20 @@ var initConfig = function(projectConfig) {
  * The singleton public agent. This is the public API of the module.
  */
 var publicAgent = {
+  // FOR_TESTING_ONLY
+  __DISABLE_START_CHECK__: false,
```
kjin: Not advocating for one way or the other, but for the sake of uniformity you could consider what I've done in the debug agent: https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/blob/master/src/index.js#L44

It would minimize the surface of the public agent, even if the additions are debug-only.

test/hooks/common.js
```diff
@@ -45,32 +37,66 @@ var SERVER_RES = '1729';
...
+    return shouldTrace.apply(this, arguments);
+  };
+}
+
 /**
  * Cleans the tracer state between test runs.
  */
-function cleanTraces() {
+function cleanTraces(agent) {
+  if (arguments.length !== 1) {
```
kjin: Will these `argument` length checks only be around until you're sure all `common` functions have been substituted correctly?

Also, checking the shape of `arguments[0]` may be more effective here...

test/standalone/test-trace-header-context.js
```diff
@@ -16,18 +16,26 @@
...
+    express = require('../hooks/fixtures/express4');
+  });
+
+  after(function() {
+    agent.stop();
   });
 
   afterEach(function() {
-    require('../..').stop();
+    cls.destroyNamespace();
```
kjin: Why do you need to do this?

matthewloring:
Try this patch for the test-trace-agent fix: http://pastebin.com/sbHXAktB.

bin/run-test.sh
```diff
@@ -39,7 +39,11 @@ function run {
 }
 
 # Run test/coverage
-run test test/hooks
+run `find test -maxdepth 1 -name '*.js' -not -name test-trace-agent.js`
+run test/test-trace-agent.js
+
+run test/hooks/*.js
```
matthewloring: Let's see if we can revert this before landing. I linked a patch that should take care of this last test.

index.js
```diff
@@ -122,11 +134,14 @@ var publicAgent = {
     return agent.addTransactionLabel(key, value);
   },
 
-  start: function(projectConfig) {
+  startAgent: function(projectConfig) {
     if (this.isActive()) { // already started.
-      agent.logger.warn('Calling start on already started agent.' +
-        'New configuration will be ignored.');
-      return this;
+      var message = 'Cannot call start on an already started agent.';
```
matthewloring: Can this be lifted to the top of the file as a constant?

index.js
```diff
@@ -229,9 +244,62 @@ var publicAgent = {
...
+ *
+ * @constructor
+ * @alias module:trace
+ *
+ * @resource [What is Stackdriver Trace]{@link
+ *   https://cloud.google.com/cloud-trace/}
+ *
+ * @param {object} options - [Configuration object](#/docs)
+ */
+function Trace(options) {
```
matthewloring: This constructor can be removed in a follow on PR as per the API design discussion.

test/hooks/common.js
```diff
@@ -45,32 +37,66 @@ var SERVER_RES = '1729';
 var SERVER_KEY = fs.readFileSync(path.join(__dirname, 'fixtures', 'key.pem'));
 var SERVER_CERT = fs.readFileSync(path.join(__dirname, 'fixtures', 'cert.pem'));
 
+function init(agent) {
```
matthewloring: Where is this function called again?

test/standalone/test-grpc-context.js
```diff
@@ -15,9 +15,10 @@
...
 'use strict';
 
-var agent = require('../..').start({ samplingRate: 0 }).private_();
-
 var common = require('../hooks/common.js');
+
+var agent = require('../..')().startAgent({ samplingRate: 0 }).private_();
+
 var assert = require('assert');
 var express = require('../hooks/fixtures/express4');
```
matthewloring: Does this loading of express/starting of the agent have to move to the before of the test below?

test/standalone/test-default-ignore-ah-health.js
```diff
@@ -15,7 +15,7 @@
  */
 'use strict';
 
-var agent = require('../..').start({samplingRate: 0});
+var agent = require('../..')().startAgent({samplingRate: 0});
 
 var assert = require('assert');
 var http = require('http');
```
matthewloring: Does this express loading need to move into the test below?

test/standalone/test-ignore-url-express.js
```diff
@@ -15,7 +15,10 @@
...
 'use strict';
 
-var agent = require('../..').start({ignoreUrls: ['/test'], samplingRate: 0});
+var agent = require('../..')().startAgent({
+  ignoreUrls: ['/test'],
+  samplingRate: 0
+});
 
 var assert = require('assert');
 var http = require('http');
```
matthewloring: Should this express be loaded inside the test?

bin/run-test.sh
DominicKramer: I've applied the patch and changed this script.

index.js
DominicKramer: I like your solution better that what I had.  I've changed the trace agent to align with the debug agent.

test/hooks/common.js
DominicKramer: Yes, I was only planning on having them around until this PR and any follow-up PRs have landed.

index.js
DominicKramer: With the change I made to align the trace agent with the debug agent with respect to forcing a new agent to be started, this comment is only used once.

index.js
DominicKramer: I've added a TODO in the code.

test/hooks/common.js
DominicKramer: This is used in `test-trace-grpc.js`.

test/standalone/test-ignore-url-express.js
DominicKramer: I've updated this test to start the agent from within a `before()` call.

test/standalone/test-default-ignore-ah-health.js
DominicKramer: I've updated this test to start the agent from within a `before()` call.

test/standalone/test-grpc-context.js
DominicKramer: I've updated this test to start the agent from within a `before()` call.

test/standalone/test-trace-header-context.js
DominicKramer: This change was needed for the tests to pass.  Matt helped me identify this fix.  I've added a TODO in the code to investigate this further (probably in a follow-up PR).

README.md
```diff
@@ -33,7 +33,7 @@ This module provides Stackdriver Trace support for Node.js applications. [Stackd
 
 3. Include and start the library *as the very first action in your application*:
 
-        require('@google/cloud-trace').start();
+        var agent = require('@google/cloud-trace')().startAgent();
```
ofrobots: Just a reminder that the `start` to `startAgent` rename will need to be undone in a follow-on.

PR #362, opened 2017-02-06
Added Plugin API docs

kjin:


plugins.md
```diff
@@ -0,0 +1,218 @@
+# Trace Plugin API Developer Guide
+
+**Note: The Trace Plugin API is experimental and is subject to change without notice.**
+
+The trace agent is driven by a set of plugins that describe how to patch a
+module to generate trace spans when that module is used. We provide plugins for
+some well-known modules such as express, mongodb, and http, and provide a plugin
+API for third-party developers to create their own.
```
matthewloring: I would use a different word from "third-party" here. Maybe just: "for developers to create their own."

plugins.md
```diff
@@ -0,0 +1,218 @@
...
+## Outline
+
+A plugin consists of a set of *patch objects*. A patch object gives information
+about how a file in a module should be patched in order to create trace spans.
+
+Each patch object should contain the following fields:
+
+* `file`: The path to the file whose exports should be patched, relative to the
+  root of the module. You can specify an empty string, or omit this field
+  entirely, to specify the root itself.
```
matthewloring: I think `root` is ambiguous here. Maybe program entrypoint or main? (https://docs.npmjs.com/files/package.json#main)

plugins.md
```diff
@@ -0,0 +1,218 @@
...
+* `versions`: A `semver` expression which will be used to control whether the
+  specified file will be patched based on the module version; the patch will
+  only be applied if the module version satisfies this expression. This might be
```
matthewloring: "if the loaded module version"

plugins.md
```diff
@@ -0,0 +1,218 @@
...
+of the module, in which case the `file` field can set to `''` or omitted. Based
+on how the module being patched is implemented, however, it may be necessary to
+patch other parts of the module as well.
+
+### Enabling plugins in the Trace Agent
+
+Developers wishing to trace their applications may specify, in the
+[configuration][config-js] object passed to the trace agent, a `plugins` field
+with a key-value pair (module name, path to plugin). For
+example, to patch just `express` with a plugin named `TODO`:
```
matthewloring: We should pick a name other than `TODO`. Maybe `express-patch`?

plugins.md
```diff
@@ -0,0 +1,218 @@
...
+
+require('@google-cloud/trace-agent').start({
+  config: {
```
DominicKramer: Is this supposed to be `plugins` instead of `config`?

plugins.md
```diff
@@ -0,0 +1,218 @@
...
+
+Creates and returns a new `Transaction` object. A `Transaction` object
+represents a root-level trace span, and exposes functions for adding labels,
+ending the span, and serializing the current trace context. Note that the
+underlying root span is started when this function is called.
+
+This function consults the trace agent's tracing policy to determine whether a
+trace span should actually be started or not. If the tracing policy prohibits
+a new span to be created under the current circumstances, this function will
+return `null` instead.
```
matthewloring: Maybe we should consider returning a dummy `Transaction` with noop operations so that users don't need to embed `null` checks or worry about sampling (they should still know it may occur).

plugins.md
```diff
@@ -0,0 +1,218 @@
...
+
+#### `api.getTransaction()`
+* Returns `Transaction`
+
+Returns a Transaction object that corresponds to a root span started earlier
+in the same context, or `null` if one doesn't exist.
+
+#### `api.runInRootSpan(options, fn)`
```
matthewloring: I think we want a description here.

plugins.md
```diff
@@ -0,0 +1,218 @@
...
+requests and propagate it in outgoing requests. The Plugin API accepts the
+serialized trace context as an [option](#trace-span-options) when creating new
+trace spans.
+
+#### `transaction.getTraceContext()` | `childSpan.getTraceContext()`
+* Returns `string`
+
+Gets the trace context serialized as a string.
+
+### Context Propagation
```
matthewloring: It would be good to include a little bit here on context propagation (like a sentence or two) even for people familiar with CLS.

plugins.md
kjin: I didn't want to pick a module name here before we decided on one. Using `./plugins/plugin-express.js`. We have to figure out what users would do if they want to include some (but not all) of the packaged plugins, because they need to get the absolute path files in our node_module.

plugins.md
kjin: Fixed

plugins.md
kjin: "root" refers to the object exported by the module itself, I'll update that.

plugins.md
kjin: Sounds good, changed.

plugins.md
kjin: Good catch, fixed

plugins.md
kjin: Oops, moved this header from somewhere below but not its description. Thanks for the catch.

kjin:
@GoogleCloudPlatform/node-team Mentioning well-known labels and constants here (forgot to do so in initial change): a40c762

plugins.md
```diff
@@ -0,0 +1,225 @@
...
+#### `transaction.addLabel(key, value) | childSpan.addLabel(key, value)`
+* `key`: `string`
+* `value`: `any`
+
+Add a label to the span associated with the calling object.
+
+#### `transaction.endSpan() | childSpan.endSpan()`
+
+Ends the span associated with the calling object. This function should only be
+called once.
```
DominicKramer: Is an error thrown if it is called more than once, or is it silently ignored?

plugins.md
```diff
@@ -0,0 +1,225 @@
...
+label keys and constants through `api.labels` and `api.constants` respectively.
+
+### Trace Spans
+
+These functions provide the capability to create trace spans, add labels to
+them, and close them. `transaction` and `childSpan` are instances of
+`Transaction` and `ChildSpan`, respectively.
+
+#### `api.createTransaction(options)`
+* `options`: [`TraceOptions`](trace-span-options)
```
DominicKramer: The links for `TraceOptions` weren't working for me.

plugins.md
```diff
@@ -0,0 +1,225 @@
...
+## What the API provides
+
+The `api` object, in short, provides functions that facilitate the following:
+
+- Creating trace spans and add labels to them.
+- Getting information about how the trace agent was configured in the
+  current application.
+- Parsing and serializing trace contexts for the sake of propagating them over
+  the network.
+- Binding callbacks and event emitters with a CLS namespace in order to
```
DominicKramer: Perhaps we should write out "continuation local storage" followed by "(CLS)" for completeness.

ofrobots:
Would it be possible to generate these docs as a result of running jsdoc on the `trace-plugin-api.js` file. i.e. Can this doc be contained in comments in the source, or is there something intrinsic that requires this to be an externally maintained documentation?

plugins.md
ofrobots: You don't need to mention the mechanism (CLS) since we may want to change that in the future. 'Binding callbacks and event emitters in order to propagate...'

plugins.md
```diff
@@ -0,0 +1,225 @@
...
+the function should proceed as if nothing is being traced.
+
+#### `transaction.runInChildSpan(options, fn)`
+* `options`: [`TraceOptions`](trace-span-options)
+* `fn`: `function(ChildSpan): any`
+* Returns `any`
+
+Runs the given function in a child span corresponding to an incoming request.
+The provided function should accept a `ChildSpan` object, which represents a
+child span.
```
ofrobots: Can the childSpan object be null here?

plugins.md
```diff
@@ -0,0 +1,225 @@
...
+
+* `name`: `string`
+  * Required
+  * The name that should be given to the newly created span.
+* `traceContext`: `string`
+  * Optional for root spans, ignored for child spans
+  * A serialized trace context. If the module being traced is a web framework,
+    the plugin that patches this module should attempt to extract this from
+    an incoming request header and set this field. Omitting this field will
+    never cause an error, but may cause trace spans that correspond to a single
```
ofrobots: 'When this property is omitted, the trace agent will generate a brand trace context for you. For standalone servers this is fine, but in a distributed (e.g. microservices) environment your single service will appear disassociated from the rest. If an incoming trace context is available, you should always provide it here. See also...'

plugins.md
kjin: I think giving `null` has its merits in that it encourages plugins to short-circuit if a trace isn't generated, to avoid the overhead of many no-op `addLabel`s, and in some cases wrapping a span-ending handler. I'm open to changing the API to give a dummy object, though.

kjin:
@ofrobots I did indeed find that I was copy-pasting a lot from the JSdocs. It's true that it would be a hassle to update both at once -- the merits I saw in a separate doc was that it could be organized differently (I'm grouping functions here by their purpose, while in the implementation they're grouped by prototype.). I think it's important to have a doc that developers can read from top-to-bottom (this file as it is might not satisfy this though.).

But since you mention it, I now think it would be best for me to remove the specifics of the functions from this doc (making sure that they're written in the JSDoc file), but I think the sections describing the plugin loader, config, and high-level overview of what the API does should stay, for the sake of having at least something centralized that can help plugin developers get started quickly.

plugins.md
kjin: Oops, forgot `#` - thanks.

plugins.md
kjin: Your replacement is addressed towards application developers but this document should address plugin developers. So while I understand the need to stress that omitting the trace context could cause services to appear dis-associated, I think this needs to be worded in a different way. I've revised the original text slightly to include some of the changes.

plugins.md
kjin: No, as of right now it's guaranteed not to be... noted.

plugins.md
```diff
@@ -0,0 +1,225 @@
...
+#### `api.enhancedReportingEnabled()`
+* Returns `boolean`
+
+Returns a boolean value describing whether the trace agent was started with
+an enhanced level of reporting. See the [configuration][config-js] object
+definition for more details.
+
+### Cross-Service Trace Contexts
+
+The Trace Agent supports distributed tracing, so that in supported web
```
kjin: Didn't know where to look for this - thanks!

plugins.md
kjin: It's silently ignored right now. I would like to add an error message if it is, though.

plugins.md
```diff
@@ -0,0 +1,230 @@
...
+`api.constants.TRACE_CONTEXT_HEADER_NAME`.
+
+#### `transaction.getTraceContext()` | `childSpan.getTraceContext()`
+* Returns `string`
+
+Gets the trace context serialized as a string.
+
+### Context Propagation
+
+#### `api.bind(fn)`
```
DominicKramer: Did you mean `api.wrap(fn)`?

plugins.md
```diff
@@ -0,0 +1,230 @@
...
+* `fn`: `function`
+* Returns `function` (same signature as `fn`)
+
+Binds the given function to the current context.
+
+#### `api.bindEmitter(emitter)`
```
DominicKramer: Did you mean `api.wrapEmitter(emitter)`?

plugins.md
```diff
@@ -0,0 +1,230 @@
...
+
+A plugin consists of a set of *patch objects*. A patch object gives information
+about how a file in a module should be patched in order to create trace spans.
+
+Each patch object should contain the following fields:
+
+* `file`: The path to the file whose exports should be patched, relative to the
+  root of the module. You can specify an empty string, or omit this field
+  entirely, to specify the export of the module itself.
+* `versions`: A `semver` expression which will be used to control whether the
```
DominicKramer: Will the plugin apply to all versions if the `versions` field is omitted?

plugins.md
```diff
@@ -0,0 +1,230 @@
...
+  specified file will be patched based on the module version; the patch will
+  only be applied if the loaded module version satisfies this expression. This
+  might be useful if your plugin only works on some versions of a module, or if
+  you are patching internal mechanisms that are specific to a certain range of
+  versions.
+* `patch`: A function describing how the given file should be patched. It will
```
DominicKramer: What happens if this field is omitted?

plugins.md
kjin: Elaborated, thanks for catching.

plugins.md
kjin: Done, good catch

kjin:
Updated: Newest changes @ 532e371

README.md
```diff
@@ -137,82 +137,179 @@ The trace configuration additionally exposes the `samplingRate` option which set
...
+
+These functions provide the capability to create trace spans, add labels to them, and close them. `transaction` and `childSpan` are instances of `Transaction` and `ChildSpan`, respectively.
+
+* `TraceApi#api.runInRootSpan(options, fn)`
+  * `options`: [`TraceOptions`](#trace-span-options)
+  * `fn`: `function(?Span): any`
+  * Returns `any`
+  * Attempts to create a root span, run the given callback, and pass it a `Span` object if the root span was successfuly created. Otherwise, the given function is run with `null` as an argument. This may be for one of two reasons:
+    * The trace policy, as specified by the user-given configuration, disallows a root span from being created under the current circumstances.
+    * The trace agent is disabled, either because it wasn't started at all, started in disabled mode, or started in an environment where the GCP project ID could not be obtained.
```
matthewloring: I would leave the last element of this list more general. Maybe: "or encountered an initialization error"?

README.md
```diff
@@ -137,82 +137,179 @@ The trace configuration additionally exposes the `samplingRate` option which set
...
+In addition to the above, `TraceApi` also provides a number of well-known label keys and constants through its `labels` and `constants` fields respectively.
+
+#### Trace Spans
+
+These functions provide the capability to create trace spans, add labels to them, and close them. `transaction` and `childSpan` are instances of `Transaction` and `ChildSpan`, respectively.
+
+* `TraceApi#api.runInRootSpan(options, fn)`
+  * `options`: [`TraceOptions`](#trace-span-options)
+  * `fn`: `function(?Span): any`
+  * Returns `any`
```
matthewloring: We should mention that this return value is that of `fn`.

README.md
```diff
@@ -137,82 +137,179 @@ The trace configuration additionally exposes the `samplingRate` option which set
...
+  * Attempts to create a root span, run the given callback, and pass it a `Span` object if the root span was successfuly created. Otherwise, the given function is run with `null` as an argument. This may be for one of two reasons:
+    * The trace policy, as specified by the user-given configuration, disallows a root span from being created under the current circumstances.
+    * The trace agent is disabled, either because it wasn't started at all, started in disabled mode, or started in an environment where the GCP project ID could not be obtained.
+* `TraceApi#createChildSpan(options)`
+  * `options`: [`TraceOptions`](#trace-span-options)
+  * Returns `?Span`
+  * Attempts to create a child span, and returns a `Span` object if this is successful. Otherwise, it returns `null`. This may be for one of several reasons:
+    * A root span wasn't created beforehand because an earlier call to `runInRootSpan` didn't generate one.
```
matthewloring: Sorry to go back and forth on this but reading through the level of detail you have here maybe it would make sense to pull this into a separate doc from the readme (leaving a basic overview of the api in the readme). WDYT?

README.md
```diff
@@ -137,82 +137,179 @@ The trace configuration additionally exposes the `samplingRate` option which set
...
+    incoming request header and set this field; omitting this field may cause
+    trace spans that correspond to a single request across several services in a
+    distributed environment (e.g. microservices) to appear disassociated with
+    one another.
+    See also [Cross-Service Trace Contexts](#cross-service-trace-contexts).
+* `url`: `string`
+  * Optional for root spans, ignored for child spans
+  * The URL of the incoming request. This only applies if the module being
+    traced is a web framework. If given, a label will automatically be created
+    for the new span for the url (under the key `url`). This field will also be
```
matthewloring: Our plugins add this label but I don't think the api will do this automatically.

README.md
```diff
@@ -137,82 +137,179 @@ The trace configuration additionally exposes the `samplingRate` option which set
...
+    compared against the trace agent's URL filtering policy to check whether a
+    span should be created.
+  * Plugin developers should favor populating this field over using
```
matthewloring: This is no longer true after increasing the default label size.

README.md
```diff
@@ -137,82 +137,179 @@ The trace configuration additionally exposes the `samplingRate` option which set
...
+  * Optional; defaults to `0`
+  * Trace spans include the call stack at the moment of creation as part of the
+    information gathered. The call stack may include undesirable frames such as
+    frames within the plugin itself. This field specifies the number of stack
+    frames to skip when writing the call stack to the trace span. Frames within
+    the trace agent implementation are automatically skipped.
+
+#### Trace Agent Configuration
+
+* `TraceApi#enhancedDatabaseReportingEnabled()`
```
matthewloring: This has been removed.

README.md
```diff
@@ -137,82 +137,179 @@ The trace configuration additionally exposes the `samplingRate` option which set
...
+  * Returns `boolean`
+  * Returns whether the trace agent was started with an enhanced level of reporting. See the [configuration][config-js] object definition for more details.
+
+#### Cross-Service Trace Contexts
+
+The Trace Agent supports distributed tracing, so that in supported web frameworks, incoming requests that are known to come from other services that are also integrated with Stackdriver Trace (through the ['x-cloud-trace-context'][stackdriver-trace-faq] field in request headers) should build spans that are aware of the information serialized in this field, known as the trace context. (For more information, see the [Dapper][dapper-paper] paper describing the distributed tracing system.)
```
matthewloring: I think this can be reworded to improve readability.

README.md
```diff
@@ -137,82 +137,179 @@ The trace configuration additionally exposes the `samplingRate` option which set
...
+- Creating trace spans and add labels to them.
+- Getting information about how the trace agent was configured in the current application.
+- Parsing and serializing trace contexts to support distributed tracing between microservices.
+- Binding callbacks and event emitters in order to propagate trace contexts across asynchronous boundaries.
+
+In addition to the above, `TraceApi` also provides a number of well-known label keys and constants through its `labels` and `constants` fields respectively.
+
+#### Trace Spans
+
+These functions provide the capability to create trace spans, add labels to them, and close them. `transaction` and `childSpan` are instances of `Transaction` and `ChildSpan`, respectively.
```
kjin: Oops, should have checked that references to `transaction` were removed.

I'm going to delete this sentence altogether, since the notation changed and I no longer refer to canonical instances of `*Span` at all.

README.md
```diff
@@ -137,82 +137,179 @@ The trace configuration additionally exposes the `samplingRate` option which set
 
 The custom tracing API can be used to add custom spans to trace. A *span* is a particular unit of work within a trace, such as an RPC request. Spans may be nested; the outermost span is called a *root span*, even if there are no nested child spans. Root spans typically correspond to incoming requests, while *child spans* typically correspond to outgoing requests, or other work that is triggered in response to incoming requests.
 
-For any of the web frameworks listed above (`express`, `hapi`, and `restify`), a root span is automatically started whenever an incoming request is received (in other words, all middleware already runs within a root span). If you wish to record a span outside of any of these frameworks, any traced code must run within a root span that you create yourself.
+For any of the web frameworks listed above (`express`, `hapi`, `koa` and `restify`), a root span is automatically started whenever an incoming request is received (in other words, all middleware already runs within a root span). If you wish to record a span outside of any of these frameworks, any traced code must run within a root span that you create yourself.
```
kjin: Done: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/362/commits/fdc7605082a8d0f4f3bc6d05ca90ab4ea2d811cc

README.md
kjin: That sounds good!

README.md
matthewloring: Still there. Ignore this.

README.md
kjin: Yeah, seems very run-on-y on a re-read. I attempted a rewrite here: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/362/commits/4c59a3c45d8151861657e966ad7e46f6ffd21131

README.md
kjin: Removed in https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/362/commits/4c59a3c45d8151861657e966ad7e46f6ffd21131

README.md
kjin: Removed in https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/362/commits/4c59a3c45d8151861657e966ad7e46f6ffd21131

README.md
kjin: I'd like to move stuff after this PR lands, for the sake of reviewability (some links will have to change). It should be straightforward.

README.md
kjin: Done

README.md
```diff
@@ -137,82 +137,179 @@ The trace configuration additionally exposes the `samplingRate` option which set
...
+  // after start() is called
+  var traceApi = require('@google/cloud-trace').get();
 ```
 
-Note that the labels parameter must be an object. Other types (e.g. string) will be silently ignored.
+The object returned by both of these calls is guaranteed to have the surface described below, even if the agent is disabled.
+
+### The `TraceApi` Object
+
+A `TraceApi` instance, in short, provides functions that facilitate the following:
```
kjin: Done

matthewloring:
LGTM

PR #363, opened 2017-02-07
Add Express Trace Plugin

kjin:
* Add Express Trace Plugin
* Remove Express Hook


kjin:
Dependent on #366 - PTAL at that first

src/plugins/plugin-express.js
```diff
@@ -0,0 +1,70 @@
+'use strict';
```
ofrobots: add copyright

src/plugins/plugin-express.js
```diff
@@ -0,0 +1,70 @@
...
+
+  methods.forEach(function(method) {
+    shimmer.wrap(express.application, method, applicationActionWrap);
+  });
+  express._plugin_patched = true;
+}
+
+module.exports = [
+  { versions: SUPPORTED_VERSIONS, patch: patchModuleRoot }
+];
```
matthewloring: '\n'

kjin:
Test failures are due to us no longer unpatching. So this is blocked on `stop` (and related tests) being removed completely.

kjin:
CI failures are due to differences in context propagation in mocha in Node v0.12. This is apparent if we add

```js
if (this.isActive() && config.forceNewAgent_) {
  agent.stop();
}
```

to the public agent's `start` method, and is the same reason why #378 was also failing in Node v0.12.

While the fix is small, I think this may a compelling reason to drop support for v0.12. @ofrobots @matthewloring What are your thoughts on this?

kjin:
Blocked on #388.

src/plugins/plugin-express.js
```diff
@@ -0,0 +1,93 @@
...
+        this.use(middleware);
+      }
+      return method.apply(this, arguments);
+    };
+  }
+
+  function middleware(req, res, next) {
+    var options = {
+      name: req.path,
+      traceContext: req.get(api.constants.TRACE_CONTEXT_HEADER_NAME),
```
matthewloring: do we need to preserve the `|| {}` here?

kjin: `traceContext` now accepts a string, which it parses. It doesn't parse it if it's not a string, so it's OK for `req.get` to return null.

PR #366, opened 2017-02-07
Plugin API: Additional comments, logging, handling no namespace

kjin:
I believe this is needed for some tests to pass since the agent might be started with no namespace. Also, added closed flag to be able to log errors if spans are ended multiple times.

### DIFF
--- a/src/trace-plugin-interface.js
+++ b/src/trace-plugin-interface.js
@@ -35,7 +35,6 @@ function ChildSpan(agent, span) {
   this.agent_ = agent;
   this.span_ = span;
   this.serializedTraceContext_ = agent.generateTraceContext(span, true);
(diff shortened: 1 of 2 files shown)


src/trace-plugin-interface.js
```diff
@@ -88,6 +94,10 @@ Transaction.prototype.addLabel = function(key, value) {
  * Ends the underlying span. This function should only be called once.
  */
 Transaction.prototype.endSpan = function() {
+  if (this.closed_) {
+    this.agent_.logger.error('A span ended more than once.');
+  }
+  this.closed_ = true;
```
matthewloring: I'm hesitant to introduce per state transaction on the transaction objects as they are not unique to the underlying root span.

PR #367, opened 2017-02-07
Remove the trace constructor and rename startAgent

DominicKramer:
This PR is a follow-up to [PR 361](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/361).

### DIFF
--- a/index.js
+++ b/index.js
@@ -126,7 +126,7 @@ var publicAgent = {
     return agent.addTransactionLabel(key, value);
   },
 
-  startAgent: function(projectConfig) {
+  start: function(projectConfig) {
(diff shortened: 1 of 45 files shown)


index.js
```diff
@@ -269,25 +241,29 @@ function Trace(options) {
...
+function get() {
   return publicAgent.get();
-};
+}
 
 global._google_trace_agent = publicAgent;
-module.exports = Trace;
+module.exports = {
+  start: start,
+  isActive: isActive,
```
matthewloring: Do we need to expose `isActive` on the public interface or can people ask the agent they receive from `start`/`get`?

PR #368, opened 2017-02-07
Remove agent isRunning/trace isActive

matthewloring:


ofrobots:
.. once the tests pass

matthewloring:
This needed some more refactoring to get the tests passing. Could I get another review before landing?

ofrobots:
OMG github is so bad at showing this diff.

Pro-tip: add `?w=1` to the URL when looking at the diff to make life 100x better.

matthewloring:
`?w=1` is sooooo magical!!!!!!! why is this not the default??!?!

PR #369, opened 2017-02-07
Run all tests stand alone

matthewloring:
This will prevent state changes in one test from affecting any others


kjin:
Looks like `test-grpc-context.js` is failing. Taking a look at it I think this test is now unnecessary altogether, since `test-trace-grpc.js` now creates child spans in gRPC server root spans..

bin/run-test.sh
```diff
@@ -39,9 +39,14 @@ function run {
 }
 
 # Run test/coverage
-run test
-run test/hooks
-for test in test/standalone/test-*.js ;
+for test in test/*.js ;
```
DominicKramer: To avoid duplication we should be able to write:
```
for test in test/*.js test/hooks/*.js ;
do
    if [[ ! $(node --version) =~ v0\.12\..* || ! "${test}" =~ .*trace\-koa\.js ]]
    then
        run "${test}"
    fi
done
```

matthewloring: Thanks! fixed

PR #370, opened 2017-02-07
Add http plugin

cristiancavalli:
Basics which could be built upon


src/hooks/index.js
```diff
@@ -43,8 +43,6 @@ var toInstrument = Object.create(null, {
       patches: {} } },
   'hapi': { enumerable: true, value: { file: './userspace/hook-hapi.js',
       patches: {} } },
-  'http': { enumerable: true, value: { file: './core/hook-http.js',
```
AdriVanHoudt: should `core/hook-http.js` not be deleted?

kjin: This replaces `hook-http.js`, so it should be deleted.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
```
ofrobots: Add back the copyright notice.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,52 @@
...
+var shimmer = require('shimmer');
+var url = require('url');
+var isString = require('is').string;
+var merge = require('lodash.merge');
+var httpAgent = require('_http_agent');
+
+function patchModuleRoot (request, api) {
```
ofrobots: style nit: omit space after function name.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,52 @@
...
+      if (!this._google_trace_patched && options) {
+        // Don't keep wrapping our same request
+        this._google_trace_patched = true;
+        return patchHTTPRequest(options, callback);
+      }
+      return request.apply(this, arguments);
+    }
+  }
+
+  function parseRequestOptions (requestOptions) {
```
ofrobots: I don't believe this needs to be in the closure as it doesn't need access to anything held by the closure. Can you pull this out of patchModuleRoot?

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,52 @@
...
+    return isString(requestOptions) ? url.parse(requestOptions) :
+      // Don't mutate user given objects
+      merge({}, requestOptions);
+  }
+
+  function extractUrl (parsedOptions) {
```
ofrobots: Likewise. I believe this can be pulled out to the top level.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+var shimmer = require('shimmer');
+var url = require('url');
+var isString = require('is').string;
+var merge = require('lodash.merge');
+var httpAgent = require('_http_agent');
+
+function patchModuleRoot (request, api) {
+  var labels = api.labels;
```
ofrobots: nit: given that there is another variable below named `labels` that shadows this, it would be better not to have this variable to begin with and use `api.labels` everywhere.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,52 @@
...
+  function patchHTTPRequest (requestOptions, callback) {
+    var parsedOptions = parseRequestOptions(requestOptions);
+    var uri = extractUrl(parsedOptions);
+    var labels = {};
+    api.runInRootSpan(options, function (transaction) {
+      if (!transaction) {
+        return request.apply(this, arguments);
+      }
+      merge(labels, {
+        [api.labels.HTTP_METHOD_LABEL_KEY]: parsedOptions.method,
```
ofrobots: Computed property names don't work with 0.12. I'm fine with dropping support for 0.12, but this would need to be an explicit decision.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,52 @@
...
+
+  function patchHTTPRequest (requestOptions, callback) {
+    var parsedOptions = parseRequestOptions(requestOptions);
+    var uri = extractUrl(parsedOptions);
+    var labels = {};
+    api.runInRootSpan(options, function (transaction) {
+      if (!transaction) {
+        return request.apply(this, arguments);
+      }
+      merge(labels, {
```
ofrobots: The merged value doesn't seem to be used.

src/plugins/plugin-http.js
ofrobots: nit: `patchHttp` would be a better name.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,52 @@
...
+var url = require('url');
+var isString = require('is').string;
+var merge = require('lodash.merge');
+var httpAgent = require('_http_agent');
+
+function patchModuleRoot (request, api) {
+  var labels = api.labels;
+  function wrapInvocation (options, callback) {
+    return function httpActionTrace () {
+      if (!this._google_trace_patched && options) {
```
ofrobots: The original version of the hook doesn't seem to have the `_google_trace_patched` logic. Is this necessary?

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+var shimmer = require('shimmer');
+var url = require('url');
+var isString = require('is').string;
+var merge = require('lodash.merge');
+var httpAgent = require('_http_agent');
+
+function patchModuleRoot (request, api) {
+  var labels = api.labels;
+  function wrapInvocation (options, callback) {
```
ofrobots: Missing call to shimmer?

src/hooks/index.js
kjin: Oh, I just realized what @AdriVanHoudt actually meant by this - @cristiancavalli, could you remove `hook-http.js`?

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,118 @@
...
+    });
+    return req;
+  }
+}
+
+module.exports = [
+  {
+    file: 'http',
+    patch: function (http, api) {
+      shimmer.wrap(http, 'request', patchModuleRoot)
```
kjin: The third argument should be a function that returns a function that replaces `http.request`. `patchModuleRoot` doesn't seem to have a return value as is.

src/plugins/plugin-http.js
kjin: You're assigning `api.labels` as `labels` here but everywhere else you're referencing it as `TraceLabels` (which is not defined) -- so maybe you meant to write TraceLabels here.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,118 @@
...
+        return patchHTTPRequest(options, callback, request);
+      }
+      return request.apply(this, arguments);
+    }
+  }
+
+  function setTraceHeader (parsedOptions, context) {
+    if (context) {
+      return merge(parsedOptions, {
+        headers: {
```
kjin: Does this overwrite `parsedOptions.headers` entirely? Maybe it should be `merge(parsedOptions.headers, {[TraceLabels.TRACE_CONTEXT_HEADER_NAME]: context})` instead?

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,118 @@
...
+    var transaction = api.createTransaction({
+      name: 'http',
+      url: uri
+    });
+    var downloadSpan;
+    var connectSpan = api.createChildSpan({name: 'http', url: uri})
+      .addLabel(TraceLabels.HTTP_METHOD_LABEL_KEY, parsedOptions.method)
+      .addLabel(TraceLabels.HTTP_URL_LABEL_KEY, uri);
+    parsedOptions = setTraceHeader(parsedOptions, transaction.getTraceContext());
+    var req = api.wrapEmitter(
```
kjin: `api.wrapEmitter` doesn't return anything; I think it mutates its argument.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,118 @@
...
+      options = url.parse(requestOptions);
+    }
+    // c.f. _http_client.js ClientRequest constructor
+    return options.hostname || options.host || 'localhost';
+  }
+
+  function patchHTTPRequest (requestOptions, callback, request) {
+    var parsedOptions = parseRequestOptions(requestOptions);
+    var uri = extractUrl(parsedOptions);
+    var transaction = api.createTransaction({
```
kjin: Since this is an outgoing request, we do not want to create a new transaction. In general we don't want the same code path to create both a root and child span.

src/plugins/plugin-http.js
cristiancavalli: I don' think so, using @kjin methodology here:
https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/363/files#diff-95b59fa789b256afddff8050472b50a2R23

src/plugins/plugin-http.js
cristiancavalli: Yes using @kjin methodology here https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/363/files#diff-95b59fa789b256afddff8050472b50a2R27

src/plugins/plugin-http.js
cristiancavalli: @kjin 
It does not:

```JS
var i = {headers: {x: 1}};
var j = {headers: {}};
_.merge(i, j)
/*

{
  "headers": {
    "x": 1
  }
}

*/
```

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,114 @@
...
+      var numBytes = 0;
+      res.on('data', function (chunk) {
+        numBytes += chunk.length;
+      });
+      res.on('end', function () {
+        requestLifecycleSpan
+          .addLabel(api.labels.HTTP_RESPONSE_SIZE_LABEL_KEY, numBytes);
+        requestLifecycleSpan
+          .addLabel(api.labels.HTTP_RESPONSE_CODE_LABEL_KEY, res.statusCode);
+        requestAnimationFrame.endSpan();
```
kjin: I'm assuming auto-complete took over here -- should this be `requestLifecycleSpan`?

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,114 @@
...
+var merge = require('lodash.merge');
+var httpAgent = require('_http_agent');
+
+function requestWrap(api, request) {
+  var labels = api.labels;
+  function setTraceHeader(parsedOptions, context) {
+    if (context) {
+      return merge(parsedOptions, {
+        headers: {
+          [TraceLabels.TRACE_CONTEXT_HEADER_NAME]: context
```
kjin: `TraceLabels` -> `api.labels` or `labels`, or vice-versa

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,114 @@
...
+    });
+    api.wrapEmitter(req);
+    req.on('error', function (e) {
+      var labels = {};
+      if (e) {
+        requestLifecycleSpan.addLabel(TraceLabels.ERROR_DETAILS_NAME, e.name);
+        requestLifecycleSpan
+          .addLabel(TraceLabels.ERROR_DETAILS_MESSAGE, e.message);
+      } else {
+        console.error('HTTP request error was null or undefined');
```
kjin: I know this is in `hook-http.js` as well, but I don't think it's the responsibility of the trace agent to print an error if an `'error'` event is emitted without an argument. In addition, I'm not sure if `req` can even do this. Maybe we can omit this entirely?

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,114 @@
...
+    return options.hostname || options.host || 'localhost';
+  }
+  function patchedHTTPRequest(requestOptions, callback, request) {
+    var parsedOptions = parseRequestOptions(requestOptions);
+    var uri = extractUrl(parsedOptions);
+    var requestLifecycleSpan = api.createChildSpan({name: 'http', url: uri})
+      .addLabel(api.labels.HTTP_METHOD_LABEL_KEY, parsedOptions.method)
+      .addLabel(api.labels.HTTP_URL_LABEL_KEY, uri);
+    parsedOptions = setTraceHeader(parsedOptions, transaction.getTraceContext());
+    var req = request.call(request, requestOptions, function (res) {
```
kjin: I think the first argument here should be either `this` or `null` (in the hook it was `this`).

src/plugins/plugin-http.js
kjin: I would suggest ensuring that the HTTP hook test passes locally, since that would easily catch minor things like this.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,114 @@
...
+      } else {
+        console.error('HTTP request error was null or undefined');
+      }
+      requestLifecycleSpan.endSpan();
+    });
+    return req;
+  }
+  return function (options, callback) {
+    return function () {
+      if (!this._google_trace_patched && options) {
```
ofrobots: I don't think this property `_google_trace_patched` should be installed. It is necessary on express, but not needed here.

src/plugins/plugin-http.js
cristiancavalli: @ofrobots done

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,114 @@
...
+      }
+      return request.apply(this, arguments);
+    };
+  };
+}
+
+module.exports = [
+  {
+    file: 'http',
+    patch: function (http, api) {
```
cristiancavalli: @kjin updated

cristiancavalli:
@GoogleCloudPlatform/node-team @AdriVanHoudt  will remove `hook-http.js` before merge. Right now I'm leaving it in so that reviewers may compare & contrast against new plugin.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,137 @@
...
+  if (!requestOptions) {
+    return request.call(request, requestOptions, callback);
+  } else if (isTraceAgentRequest(requestOptions)) {
+    return request.call(request, requestOptions, callback);
+  } else if (!api.getRootSpan()) {
+    if (isString(requestOptions)) {
+      requestOptions = url.parse(requestOptions);
+    }
+    // What's the new logger target?
+    // console.log('Untraced http uri:', extractUrl(requestOptions));
```
kjin: You can remove this line -- the agent prints a message if `getRootSpan()` returns null.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,137 @@
...
+    merge({headers: {}}, requestOptions);
+}
+
+function patchedHTTPRequest(requestOptions, callback, request, api) {
+  if (!requestOptions) {
+    return request.call(request, requestOptions, callback);
+  } else if (isTraceAgentRequest(requestOptions)) {
+    return request.call(request, requestOptions, callback);
+  } else if (!api.getRootSpan()) {
+    if (isString(requestOptions)) {
```
kjin: Since the logger doesn't print anymore, there's no need for this if block since `url.parse(requestOptions)` isn't used anywhere subsequently.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,137 @@
...
+      requestOptions = url.parse(requestOptions);
+    }
+    // What's the new logger target?
+    // console.log('Untraced http uri:', extractUrl(requestOptions));
+    return request.call(request, requestOptions, callback);
+  }
+  var parsedOptions = parseRequestOptions(requestOptions);
+  var uri = extractUrl(parsedOptions);
+  var requestLifecycleSpan = api.createChildSpan({name: getSpanName(parsedOptions)});
+  if (!requestLifecycleSpan) {
```
kjin: `requestLifecycleSpan` is guaranteed to be non-null because `api.createChildSpan` depends on `api.getRootSpan`, the latter already being verified as non-null earlier. Just FYI... I'm not 100% if it is a best practice to check anyway.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,138 @@
...
+
+function getSpanName(requestOptions) {
+  if (isString(requestOptions)) {
+    requestOptions = url.parse(requestOptions);
+  }
+  // c.f. _http_client.js ClientRequest constructor
+  return requestOptions.hostname || requestOptions.host || 'localhost';
+}
+
+function setTraceHeader(parsedOptions, context, api) {
```
kjin: `api` is unused here.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,138 @@
...
+  requestLifecycleSpan.addLabel(api.labels.HTTP_URL_LABEL_KEY, uri);
+  parsedOptions = setTraceHeader(parsedOptions, requestLifecycleSpan.getTraceContext(), api);
+  var req = request.call(request, parsedOptions, function (res) {
+    api.wrapEmitter(res);
+    var numBytes = 0;
+    res.on('data', function (chunk) {
+      numBytes += chunk.length;
+    });
+    res.on('end', function () {
+      if (requestLifecycleSpan) {
```
kjin: I think this check is not necessary; requestLifecycleSpan must be non-null.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,138 @@
...
+  api.wrapEmitter(req);
+  req.on('error', function (e) {
+    if (e && requestLifecycleSpan) {
+      requestLifecycleSpan.addLabel(api.labels.ERROR_DETAILS_NAME, e.name);
+      requestLifecycleSpan
+        .addLabel(api.labels.ERROR_DETAILS_MESSAGE, e.message);
+      requestLifecycleSpan.endSpan();
+    } else if (!e) {
+      // What's the new logger target?
+      // console.error('HTTP request error was null or undefined', e);
```
kjin: I may have mentioned this earlier, but I don't think we need to log an error here at all. The trace agent shouldn't be responsible for notifying users that an http request emitted a malformed error event.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,138 @@
...
+        requestLifecycleSpan.endSpan();
+      }
+    });
+    if (callback) {
+      return callback(res);
+    }
+  });
+  api.wrapEmitter(req);
+  req.on('error', function (e) {
+    if (e && requestLifecycleSpan) {
```
kjin: Same as above; I don't think `requestLifecycleSpan`'s existence needs to be checked here.

src/trace-plugin-interface.js
```diff
@@ -224,12 +225,14 @@ PluginAPI.prototype.wrap = function(fn) {
...
  * @param {EventEmitter} emitter An event emitter whose handlers should have
  * the trace context binded to them.
+ * @returns {EventEmitter} - returns the given event emitter
  */
 PluginAPI.prototype.wrapEmitter = function(emitter) {
   if (!this.agent_.namespace) {
     this.logger_.warn('Trace agent: No CLS namespace to bind emitter to');
   }
   this.agent_.namespace.bindEmitter(emitter);
+  return emitter;
```
kjin: I would want this to be added once we reach a consensus that it would improve the interface, so please hold off on this change for now.

cristiancavalli:
Failures on 0.12 are due to bracket syntax for object construction. Since we don't need support 0.12 anymore I'm leaving it in.

cristiancavalli:
Not sure what the appveyor failure is about, since that same test throws the same exception on v4 and it still keeps going. Removed old hook file, tests are green for versions officially supported.

ofrobots:
Let me fix the long lines.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,124 @@
...
+          return callback(res);
+        }
+      });
+      api.wrapEmitter(req);
+      req.on('error', function (e) {
+        if (e) {
+          requestLifecycleSpan.addLabel(api.labels.ERROR_DETAILS_NAME, e.name);
+          requestLifecycleSpan
+            .addLabel(api.labels.ERROR_DETAILS_MESSAGE, e.message);
+        } else {
```
kjin: Empty `else` statement.

src/plugins/plugin-http.js
```diff
@@ -0,0 +1,124 @@
...
+          // console.error('HTTP request error was null or undefined', e);
+        }
+        requestLifecycleSpan.endSpan();
+      });
+      return req;
+    };
+  });
+
+  function isTraceAgentRequest (options) {
+    return options && options.headers &&
```
matthewloring: This will only work if `options` is an object and not a string. I think we're fine because all of the requests made by our agent use an object but it may be worth generalizing or parsing string options in a follow on (especially since the old http hook had this same logic).

test/test-hooks-sample-warning.js
```diff
@@ -92,7 +92,7 @@ describe('express + dbs', function() {
         http.get({port: common.serverPort + 1}, function(res) {
           server.close();
           common.cleanTraces(agent);
-          assert.equal(debugCount, 2);
+          assert.equal(debugCount, 0);
```
matthewloring: As discussed in person, we should remove dead code from this test on the log checks and open a bug to track plugin specific log messages.

PR #372, opened 2017-02-07
Plugin API: Loading Improvements

kjin:
- Warn if module version is unsatisfied by plugin
- Fix possible bad behavior if `patch.versions` doesn't exist
- Fix programming error related to patch sets
- Store multiple module versions
- Add plugin loader tests

### DIFF
--- a/src/trace-plugin-loader.js
+++ b/src/trace-plugin-loader.js
@@ -116,7 +116,7 @@ function activate(agent) {
       if (!patchSet) {
         // Load the plugin object
         var plugin = originalModuleLoad(instrumentation.file, module, false);
-        patchSet = [];
+        patchSet = {};
         plugin.forEach(function(patch) {
           if (semver.satisfies(version, patch.versions)) {
             patchSet[patch.file] = {
@@ -126,6 +126,10 @@ function activate(agent) {
             };
           }
         });
+        if (Object.keys(patchSet).length === 0) {
+          logger.warn(moduleRoot + ': version ' + version + ' not supported ' + 
+            'by plugin.');
+        }
         instrumentation.patches[moduleRoot][version] = patchSet;
       }
       Object.keys(patchSet).forEach(function(file) {
--- a/src/trace-plugin-loader.js
+++ b/src/trace-plugin-loader.js
@@ -118,7 +118,7 @@ function activate(agent) {
         var plugin = originalModuleLoad(instrumentation.file, module, false);
         patchSet = {};
         plugin.forEach(function(patch) {
-          if (semver.satisfies(version, patch.versions)) {
+          if (!patch.versions || semver.satisfies(version, patch.versions)) {
             patchSet[patch.file] = {
               file: patch.file || '',
               patch: patch.patch,
@@ -146,7 +146,7 @@ function activate(agent) {
         }
         patch.active = true;
       });
-      var rootPatch = patchSet.filter(function(patch) { return !patch.name; })[0];
+      var rootPatch = patchSet[''];
       if (rootPatch && rootPatch.intercept) {
         return rootPatch.module;
       } else {
--- a/src/trace-plugin-loader.js
+++ b/src/trace-plugin-loader.js
@@ -201,19 +201,6 @@ function activate(agent) {
 }
 
 function deactivate() {
-  for (var moduleName in plugins) {
-    var instrumentation = plugins[moduleName];
-    for (var moduleRoot in instrumentation.patches) {
-      var modulePatch = instrumentation.patches[moduleRoot];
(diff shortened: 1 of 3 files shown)


src/trace-plugin-loader.js
```diff
@@ -116,16 +116,20 @@ function activate(agent) {
       if (!patchSet) {
         // Load the plugin object
         var plugin = originalModuleLoad(instrumentation.file, module, false);
-        patchSet = [];
+        patchSet = {};
```
matthewloring: Was `instrumentation.patches[moduleRoot][version]` previously an object or array? If it was an array we should update its documentation as well.

kjin: I don't think the original hook loading mechanism had anything stored as an array. I think I introduced `patchSet` as an array in the plugin loader and changed my mind about it later.

kjin: Oh - I think `[version]` is a mistake. It's not seen anywhere else. Can we guarantee that there's only one version of each module we load? I know there are things like `npm-install-package` but I hope that can count as a pathological case.

kjin: It seems like `intrumentation.patches[moduleRoot]` is treated as an object everywhere else.

matthewloring: We cannot assume a single version unfortunately. Things like `redis` that are used for caching may be required at different versions for by different dependencies of the primary project.

kjin: OK, got it - I put it back in 9fab021

kjin:
I realize that the amount of change the plugin loader needs to go through (due to bugginess) doesn't install confidence. So I have added tests that should hopefully showcase that things work smoothly now, and also to bump coverage up.

src/trace-plugin-loader.js
```diff
@@ -140,49 +109,30 @@ function activate(agent) {
...
+      return instrumentation.patches[moduleRoot] &&
+        instrumentation.patches[moduleRoot][version];
     }
 
     // Future requires get patched as they get loaded.
     return function Module_load(request, parent, isMain) {
       var instrumentation = plugins[request];
 
-      if (instrumentation &&
-          agent.config().excludedHooks.indexOf(request) === -1) {
```
matthewloring: Do we still respect `excludedHooks`?

src/trace-plugin-loader.js
```diff
@@ -197,19 +147,6 @@ function activate(agent) {
 }
 
 function deactivate() {
-  for (var moduleName in plugins) {
```
matthewloring: Is this combining the `stop` changes alongside other improvements? I feel like the removal of `unpatch` would better align with that PR. It's fine to leave it here but I am curious if it is needed outside of the context of the `stop` changes.

test/test-plugin-loader.js
```diff
@@ -0,0 +1,317 @@
...
+    exports: mock,
+    version: version
+  };
+}
+
+// This function creates an object with just enough properties to appear to the
+// plugin loader as the trace agent. It accepts the list of plugins that the
+// plugin loader reads.
+function createFakeAgent(plugins) {
+  function writeToLog(log, data) {
```
matthewloring: We do comparable custom logging to a string elsewhere in the trace agent. Maybe we could generalize this across all of our tests as opposed to having each one patch the logging methods differently.

src/trace-plugin-loader.js
kjin: I suppose since loading plugins is technically a "new" feature this should not have been included to begin with, which is why I removed it here. Also, other changes in this PR change how plugins are stored, and I felt that instead of modifying `deactivate` to reflect this change, I might as well remove it altogether if it will be removed eventually.

src/trace-plugin-loader.js
kjin: It's still respected in the `hooks` loader. I think the `excludedHooks` option would have to be deprecated since the name is outdated.

matthewloring: That makes sense, once plugins are included explicitly we won't have need of this configuration any more.

src/trace-plugin-loader.js
matthewloring: Sounds good.

test/test-plugin-loader.js
matthewloring: I don't think this has to be done as part of this pull request but it would be a nice clean up thing to do in the future.

kjin:
@matthewloring Thanks for taking a look at this again!

test/test-plugin-loader.js
kjin: Sounds good, but I'll leave that to a future PR.

PR #373, opened 2017-02-07
Plugin API: Add createChildSpan

kjin:
Introduces `Transaction.prototype.createChildSpan` (and `api.prototype.createChildSpan`). Would like to rip out `runInChildSpan` if possible, it no longer has a reason to be used.

### DIFF
--- a/src/trace-plugin-interface.js
+++ b/src/trace-plugin-interface.js
@@ -121,23 +121,6 @@ Transaction.prototype.createChildSpan = function(options, fn) {
 };
 
 /**
- * Runs the given function in a child span, passing it an object that
- * exposes an interface for adding labels and closing the span.
- * @param {object} options An object that specifies options for how the child
- * span is created and propogated. @see Transaction.prototype.createChildSpan
- * @param {function(ChildSpan)} fn A function that will be called exactly
- * once, with a ChildSpan object exposing an interface operating on the child
- * span.
- * @returns The return value of calling fn.
- */
-Transaction.prototype.runInChildSpan = function(options, fn) {
-  options = options || {};
-  var childContext = this.agent_.startSpan(options.name, {},
-    options.skipFrames ? options.skipFrames + 1 : 1);
-  return fn(new ChildSpan(this.agent_, childContext));
-};
-
-/**
  * PluginAPI constructor. Don't call directly - a plugin object will be passed to
  * plugin themselves
  * TODO(kjin): Should be called something else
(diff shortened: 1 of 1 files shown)


src/trace-plugin-interface.js
```diff
@@ -209,10 +223,32 @@ PluginAPI.prototype.runInRootSpan = function(options, fn) {
 
 /**
  * Convenience method which obtains a Transaction object with getTransaction()
+ * and creates a new child span nested within this transaction. If there is
+ * no current Transaction object, this function returns null.
+ * @param {object} options An object that specifies options for how the child
+ * span is created and propogated. @see Transaction.prototype.createChildSpan
+ * @returns A new ChildSpan object, or null if there is no active root span.
+ */
+PluginAPI.prototype.createChildSpan = function(options, fn) {
```
matthewloring: Do we want this to replace or operate alongside `runInChildSpan`?

kjin: Ideally it should reaplce runInChildSpan. I don't see any reason to keep it around otherwise. Except that it's "semver major" in a sense. Though no one is using it.

matthewloring: I think it should replace as well.

DominicKramer: I agree.  With both `createChildSpan` and `runInChildSpan` present, one might expect one to be synchronous while the other is asynchronous, but `runInChildSpan` is synchronous (provided the supplied function is synchronous).

kjin:
@GoogleCloudPlatform/node-team Removed `runInChildSpan`, please take another look. Thanks

src/trace-plugin-interface.js
```diff
@@ -209,26 +206,23 @@ PluginAPI.prototype.runInRootSpan = function(options, fn) {
...
  */
-PluginAPI.prototype.runInChildSpan = function(options, fn) {
+PluginAPI.prototype.createChildSpan = function(options, fn) {
   var transaction = this.getTransaction();
   if (transaction) {
     options = options || {};
     var childContext = transaction.agent_.startSpan(options.name, {},
       options.skipFrames ? options.skipFrames + 1 : 1);
-    return fn(new ChildSpan(transaction.agent_, childContext));
+    return new ChildSpan(transaction.agent_, childContext);
```
ofrobots: This seems to be duplicated code from `Transaction.prototype.createChildSpan`. Would this work:

```
if (transaction) {
  return transaction.createChildSpan(options)
}
```

src/trace-plugin-interface.js
```diff
@@ -209,26 +206,23 @@ PluginAPI.prototype.runInRootSpan = function(options, fn) {
...
- * once. @see Transaction.prototype.runInChildSpan
- * @returns The return value of calling fn.
+ * and creates a new child span nested within this transaction. If there is
+ * no current Transaction object, this function returns null.
+ * @param {object} options An object that specifies options for how the child
+ * span is created and propogated. @see Transaction.prototype.createChildSpan
+ * @returns A new ChildSpan object, or null if there is no active root span.
  */
-PluginAPI.prototype.runInChildSpan = function(options, fn) {
+PluginAPI.prototype.createChildSpan = function(options, fn) {
```
ofrobots: `fn` is no longer needed?

src/trace-plugin-interface.js
```diff
@@ -111,16 +111,13 @@ Transaction.prototype.getTraceContext = function() {
...
  * collecting call stack information for the child span, starting from the top;
  * this should be set to avoid including frames in the plugin. Defaults to 0.
- * @param {function(ChildSpan)} fn A function that will be called exactly
- * once, with a ChildSpan object exposing an interface operating on the child
- * span.
- * @returns The return value of calling fn.
+ * @returns A new ChildSpan object.
  */
-Transaction.prototype.runInChildSpan = function(options, fn) {
+Transaction.prototype.createChildSpan = function(options, fn) {
```
ofrobots: `fn` is no longer needed?

src/trace-plugin-interface.js
kjin: The duplication is intentional to avoid having to deal with an extra layer of `skipFrames`. In my view the duplication is small enough (4 lines) to not be extracted into a separate function.

kjin: But now, looking at it again, the duplication is still unnecessary because there is never any reason to call runInChildSpan on the `Transaction` object. I've removed it.

src/trace-plugin-interface.js
kjin: Good catch, removed.

src/trace-plugin-interface.js
kjin: Ditto

PR #374, opened 2017-02-08
Restify plugin

ofrobots:


src/plugins/plugin-restify.js
```diff
@@ -0,0 +1,84 @@
...
+      // we use the path part of the url as the span name and add the full url
+      // as a label later.
+      name: req.path,
+      url: req.url,
+      traceContext: req.header(api.constants.TRACE_CONTEXT_HEADER_NAME, null),
+      skipFrames: 3
+    };
+
+    // TODO(ofrobots): can this be moved to the inner scope near the req.end
+    // clobber?
```
matthewloring: I think that should be fine.

src/plugins/plugin-restify.js
```diff
@@ -0,0 +1,84 @@
...
+      var fullUrl = req.header('X-Forwarded-Proto', 'http') + '://' +
+                    req.header('host') + req.url;
+      transaction.addLabel(api.labels.HTTP_METHOD_LABEL_KEY, req.method);
+      transaction.addLabel(api.labels.HTTP_URL_LABEL_KEY, fullUrl);
+      transaction.addLabel(api.labels.HTTP_SOURCE_IP,
+                           req.connection.remoteAddress);
+
+      res.end = function(chunk, encoding) {
+        res.end = originalEnd;
+        var returned = res.end(chunk, encoding);
```
matthewloring: Can we apply `originalEnd` here instead of reassigning `res.end`?

ofrobots: I do think we need to reset `res.end` in case that's observed. Regardless, calling `originalEnd` would be a logical difference between the old hooks and this code, so I would like to keep it out of this PR.

matthewloring: I'm not sure what could observe the difference but I support keeping the logic the same as it was in the old hooks.

src/plugins/plugin-restify.js
ofrobots: Can be done in a follow-on change. For now, I would like this to be logically equivalent to the existing hook code.

src/plugins/plugin-restify.js
```diff
@@ -0,0 +1,84 @@
...
+
+var SUPPORTED_VERSIONS = '<=4.x';
+
+function patchRestify(restify, api) {
+  shimmer.wrap(restify, 'createServer', createServerWrap);
+
+  function createServerWrap(createServer) {
+    return function createServerTrace() {
+      var server = createServer.apply(this, arguments);
+      servers.use(middleware);
```
kjin: `servers` -> `server`

src/plugins/plugin-restify.js
```diff
@@ -0,0 +1,84 @@
...
+      url: req.url,
+      traceContext: req.header(api.constants.TRACE_CONTEXT_HEADER_NAME, null),
+      skipFrames: 3
+    };
+
+    // TODO(ofrobots): can this be moved to the inner scope near the req.end
+    // clobber?
+    var originalEnd = res.end;
+    api.runInRootSpan(options, function(transaction) {
+      if (transaction) {
```
kjin: Should be `if (!transaction)`.

src/plugins/plugin-restify.js
ofrobots: Done.

src/plugins/plugin-restify.js
ofrobots: Done.

src/plugins/plugin-restify.js
```diff
@@ -0,0 +1,90 @@
...
+      name: req.path(),
+      url: req.url,
+      traceContext: req.header(api.constants.TRACE_CONTEXT_HEADER_NAME, null),
+      skipFrames: 3
+    };
+
+    // TODO(ofrobots): can this be moved to the inner scope near the req.end
+    // clobber?
+    var originalEnd = res.end;
+    api.runInRootSpan(options, function(transaction) {
```
matthewloring: Prefer to avoid the word `transaction`. I think we settled on `rootSpan` to describe this entity.

ofrobots:
Variable changed. CI green. Landing.

PR #375, opened 2017-02-08
Make run-bump point to v0.6.x branch

kjin:


matthewloring:
I don't think we should hard code a branch other than master here. Maybe we can make the tool more general to accept the branch name to use as default?

PR #376, opened 2017-02-08
Support mysql tracing through the plugin API

DominicKramer:


DominicKramer:
PTAL

src/plugins/plugin-mysql.js
```diff
@@ -0,0 +1,89 @@
...
+    var labels = {};
+    if (api.enhancedDatabaseReportingEnabled()) {
+      if (err) {
+        labels.error = err;
+      }
+      if (res) {
+        labels.result = res;
+      }
+    }
+    span.endSpan(labels);
```
kjin: `endSpan` currently takes no arguments, so you'll have to use `addLabel`. Do you think having `endSpan` take in an object with key-value fields is useful? (Here I think it's actually slightly cleaner to use `addLabel` on lines 53 and 56.)

DominicKramer: Done

src/plugins/plugin-mysql.js
```diff
@@ -0,0 +1,94 @@
...
+
+var SUPPORTED_VERSIONS = '^2.9.x';
+
+function createQueryWrap(api, createQuery) {
+  return function createQuery_trace(sql, values, cb) {
+    var span = api.createChildSpan({
+      name: 'mysql-query'
+    });
+    var query = createQuery.apply(this, arguments);
+    if (span) {
```
matthewloring: Can we change this to an `if (!span) return query`?

PR #377, opened 2017-02-08
Support redis tracing through the plugin API

DominicKramer:


src/plugins/plugin-redis.js
```diff
@@ -0,0 +1,198 @@
...
+function setupSpan(api, cmd, args, skipped_frames) {
+  var labels = { command: cmd };
+  if (api.enhancedDatabaseReportingEnabled()) {
+    labels.arguments = JSON.stringify(args);
+  }
+  var span = api.createChildSpan({
+    name: 'redis-' + cmd,
+    skipFrames: skipped_frames + 1
+  });
+  Object.keys(labels).forEach(function(key) {
```
DominicKramer: Done

src/plugins/plugin-redis.js
```diff
@@ -0,0 +1,198 @@
...
+      cmd.callback = wrapCallback(api, span, cmd.callback);
+      return internal_send_command.call(this, cmd);
+    }
+    return startSpanFromArguments(api, cmd, args, cb, internal_send_command.bind(this));
+  };
+}
+
+function sendCommandWrap(api, send_command) {
+  return function send_command_trace(cmd, args, cb) {
+    var root = api.getTransaction();
```
DominicKramer: Done

src/plugins/plugin-redis.js
```diff
@@ -0,0 +1,198 @@
...
+    var labels = {};
+    if (api.enhancedDatabaseReportingEnabled()) {
+      if (err) {
+        labels.error = err;
+      }
+      if (res) {
+        labels.result = res;
+      }
+    }
+    Object.keys(labels).forEach(function(key) {
```
DominicKramer: Done

src/plugins/plugin-redis.js
```diff
@@ -0,0 +1,198 @@
...
+    });
+    span.endSpan();
+    if (done) {
+      done(err, res);
+    }
+  };
+  return api.wrap(fn);
+}
+
+////////////////////////////////////////////////////////////////////////////////
```
DominicKramer: I've refactored the code to make it easier to read.

src/plugins/plugin-redis.js
```diff
@@ -0,0 +1,198 @@
...
+function patchAll(redis, api) {
+  shimmer.wrap(redis, 'createClient', createClientWrap.bind(null, api));
+}
+
+module.exports = [
+  {
+    file: '',
+    versions: '<2.3.x',
+    patch: function(redis, api) {
+      patchBelowV2_6(redis, api);
```
DominicKramer: I've refactored the code to make it easier to read.

src/plugins/plugin-redis.js
```diff
@@ -0,0 +1,198 @@
...
+      patchAboveV2_3(redis, api);
+      patchBelowV2_6(redis, api);
+      patchAll(redis, api);
+    }
+  },
+  {
+    file: '',
+    versions: '2.6.x',
+    patch: function(redis, api) {
+      patchEqualV2_6(redis, api);
```
DominicKramer: I've refactored the code to make it easier to read.

src/plugins/plugin-redis.js
```diff
@@ -0,0 +1,210 @@
...
+      cb = args.pop();
+    }
+  }
+  var span = setupSpan(api, cmd, args, 1);
+  return send_command(cmd, args, wrapCallback(api, span, cb));
+}
+
+function internalSendCommandWrap(api, internal_send_command) {
+  return function internal_send_command_trace(cmd, args, cb) {
+    var root = api.getRootSpan();
```
matthewloring: This `root` check can be replaced by checks on the child spans when they are created. This may require some refactoring.

src/plugins/plugin-redis.js
```diff
@@ -0,0 +1,210 @@
...
+      cmd.callback = wrapCallback(api, span, cmd.callback);
+      return internal_send_command.call(this, cmd);
+    }
+    return startSpanFromArguments(api, cmd, args, cb, internal_send_command.bind(this));
+  };
+}
+
+function sendCommandWrap(api, send_command) {
+  return function send_command_trace(cmd, args, cb) {
+    var root = api.getRootSpan();
```
matthewloring: Same here.

src/plugins/plugin-redis.js
```diff
@@ -0,0 +1,210 @@
...
+    }
+    span.endSpan();
+    if (done) {
+      done(err, res);
+    }
+  };
+  return api.wrap(fn);
+}
+
+function wrapInternalSendCommand(redis, api) {
```
matthewloring: Personally, I don't find that pulling single statements into their own functions makes things more readable but it's just a matter of preference.

src/plugins/plugin-redis.js
```diff
@@ -0,0 +1,211 @@
...
+  var span = setupSpan(api, cmd, args, 1);
+  if (!span) {
+    return send_command(cmd, args, cb);
+  }
+  return send_command(cmd, args, wrapCallback(api, span, cb));
+}
+
+function internalSendCommandWrap(api, internal_send_command) {
+  return function internal_send_command_trace(cmd, args, cb) {
+    if (arguments.length === 1 && typeof cmd === 'object') {
```
kjin: Can you add a comment describing what is happening here? Just to save readers time in the future, you could say in brief that `internal_send_command` either takes three arguments `cmd, args, cb` or a single one `{command, args, callback}`.

kjin: In addition - is it possible to simplify this? It seems like the real reason for the `if` block is to correctly identify the command, arguments, and callback, if they are not provided directly as arguments. The remaining stuff in the `if` block can be done in `startSpanFromArguments` -- which also removed the need for a separate `setupSpan` object.

src/plugins/plugin-redis.js
```diff
@@ -0,0 +1,211 @@
...
+  }
+  span.addLabel('command', cmd);
+  if (api.enhancedDatabaseReportingEnabled()) {
+    span.addLabel('arguments', JSON.stringify(args));
+  }
+  return span;
+}
+
+function startSpanFromArguments(api, cmd, args, cb, send_command) {
+  if (!cmd || !args || typeof cmd !== 'string' || !Array.isArray(args) ||
```
kjin: Can you add a comment describing when any of these cases would apply here?

DominicKramer: Done

src/plugins/plugin-redis.js
DominicKramer: I've added a TODO for this.

src/plugins/plugin-redis.js
DominicKramer: Done

src/plugins/plugin-redis.js
DominicKramer: Done

PR #378, opened 2017-02-10
Remove agent.stop()

kjin:
This change removes agent.stop(), and updates tests to reflect this change.

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,6 +27,7 @@ var constants = require('./src/constants.js');
 var gcpMetadata = require('gcp-metadata');
 var semver = require('semver');
 var traceUtil = require('./src/util.js');
+var SpanData = require('./src/span-data.js');
 var util = require('util');
 
 var modulesLoadedBeforeTrace = [];
@@ -41,8 +42,33 @@ for (var i = 0; i < filesLoadedBeforeTrace.length; i++) {
 
 var onUncaughtExceptionValues = ['ignore', 'flush', 'flushAndExit'];
 
+/**
+ * Phantom implementation of the trace agent. This allows API users to decouple
+ * the enable/disable logic from the calls to the tracing API. The phantom API
+ * has a lower overhead than isEnabled checks inside the API functions.
+ * @private
+ */
+var phantomTraceAgent = {
+  startSpan: function() { return SpanData.nullSpan; },
+  endSpan: function(spanData) { spanData.close(); },
+  runInSpan: function(name, labels, fn) {
+    if (typeof(labels) === 'function') {
+      fn = labels;
+    }
+    fn(function() {});
+  },
+  runInRootSpan: function(name, labels, fn) {
+    if (typeof(labels) === 'function') {
+      fn = labels;
+    }
+    fn(function() {});
+  },
+  setTransactionName: function() {},
+  addTransactionLabel: function() {}
+};
+
 /** @private */
-var agent = null;
+var agent = phantomTraceAgent;
 
 var initConfig = function(projectConfig) {
   var config = {};
@@ -69,8 +95,7 @@ var initConfig = function(projectConfig) {
  */
 var publicAgent = {
   isActive: function() {
-    // TODO: remove
-    return !!agent;
+    return agent !== phantomTraceAgent;
   },
 
   startSpan: function(name, labels) {
--- a/index.js
+++ b/index.js
@@ -27,7 +27,6 @@ var constants = require('./src/constants.js');
 var gcpMetadata = require('gcp-metadata');
 var semver = require('semver');
 var traceUtil = require('./src/util.js');
-var SpanData = require('./src/span-data.js');
 var util = require('util');
 
 var modulesLoadedBeforeTrace = [];
@@ -42,33 +41,8 @@ for (var i = 0; i < filesLoadedBeforeTrace.length; i++) {
 
 var onUncaughtExceptionValues = ['ignore', 'flush', 'flushAndExit'];
 
-/**
- * Phantom implementation of the trace agent. This allows API users to decouple
- * the enable/disable logic from the calls to the tracing API. The phantom API
- * has a lower overhead than isEnabled checks inside the API functions.
- * @private
- */
-var phantomTraceAgent = {
-  startSpan: function() { return SpanData.nullSpan; },
-  endSpan: function(spanData) { spanData.close(); },
-  runInSpan: function(name, labels, fn) {
-    if (typeof(labels) === 'function') {
-      fn = labels;
-    }
-    fn(function() {});
-  },
-  runInRootSpan: function(name, labels, fn) {
-    if (typeof(labels) === 'function') {
-      fn = labels;
-    }
-    fn(function() {});
-  },
-  setTransactionName: function() {},
-  addTransactionLabel: function() {}
-};
-
 /** @private */
-var agent = phantomTraceAgent;
+var agent = null;
 
 var initConfig = function(projectConfig) {
   var config = {};
@@ -95,7 +69,8 @@ var initConfig = function(projectConfig) {
  */
 var publicAgent = {
   isActive: function() {
-    return agent !== phantomTraceAgent;
+    // TODO: remove
+    return !!agent;
   },
 
   startSpan: function(name, labels) {
@@ -186,12 +161,9 @@ var publicAgent = {
           }
         }
         if (err) {
-          // Fatal error. Disable the agent.
           logger.error('Unable to acquire the project number from metadata ' +
             'service. Please provide a valid project number as an env. ' +
-            'variable, or through config.projectId passed to start(). ' +
-            'Disabling trace agent. ' + err);
-          publicAgent.stop();
+            'variable, or through config.projectId passed to start(). ' + err);
           return;
         }
         config.projectId = projectId;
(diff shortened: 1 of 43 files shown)


index.js
```diff
@@ -186,12 +186,9 @@ var publicAgent = {
           }
         }
         if (err) {
-          // Fatal error. Disable the agent.
           logger.error('Unable to acquire the project number from metadata ' +
             'service. Please provide a valid project number as an env. ' +
-            'variable, or through config.projectId passed to start(). ' +
-            'Disabling trace agent. ' + err);
-          publicAgent.stop();
+            'variable, or through config.projectId passed to start(). ' + err);
```
matthewloring: I think that without calling stop here, the agent will continue attempting to publish to the service even though no publish will succeed. We may still need a way to turn off the trace writer even if we stop allowing developers from stopping an already started agent.

test/test-agent-metadata.js
```diff
@@ -42,32 +42,13 @@ describe('agent interaction with metadata service', function() {
...
-  });
-
-  it('should stop when the project number cannot be acquired', function(done) {
-    nock.disableNetConnect();
-    var scope = nock('http://metadata.google.internal')
-                .get('/computeMetadata/v1/project/project-id')
-                .times(2)
-                .reply(404, 'foo');
-
-    agent = trace.start({logLevel: 0});
```
matthewloring: I think we want some variant of this test to still exist.

src/trace-agent.js
```diff
@@ -304,7 +288,7 @@ TraceAgent.prototype.generateTraceContext = function(spanData, traced) {
 
 module.exports = {
   get: function(config, logger) {
-    if (traceAgent) {
+    if (traceAgent && !config.forceNewAgent_) {
```
matthewloring: Could this property be documented inside `config.js`.

src/trace-agent.js
```diff
@@ -61,22 +61,6 @@ function TraceAgent(config, logger) {
 }
 
 /**
- * Halts this agent and unpatches any patched modules.
- */
-TraceAgent.prototype.stop = function() {
```
DominicKramer: What do you think about having this functionality in a method in `test/hooks/common.js` since stopping is really only needed now in testing?  Then test code could use `common.js` to tear down the agent where the agent was being stopped.  The benefit would be that then (with some changes to the e2e tests) I think we could get rid of the `forceNewAgent_` config option.

kjin: Having `stop()` in `test/hooks/common.js` still requires support on the `src` side, such as exposing functionality to reset the reference to `agent` in `index.js`. No matter what there will have to be something exposed that is testing-only, and I went with `forceNewAgent_` because it (1) doesn't introduce extra methods and (2) is uniform with what we did in the Debug Agent.

index.js
kjin: I've moved the responsibility of handling metadata-based project ID errors to `trace-writer.js`:

https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/378/commits/b7822a1a820172328f3a4ee9bfdc77aa98405a25#diff-d2b33b8013e140d6fa2ffc2d03dcf77a

test/test-agent-metadata.js
kjin: Since I've moved project ID error handling in `trace-writer.js`, I've reinstated this test in `test-trace-writer.js`:

https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/378/commits/b7822a1a820172328f3a4ee9bfdc77aa98405a25#diff-fdad2e1d3cceacb6b3585907721d641f

src/trace-agent.js
kjin: Done

kjin:

src/trace-writer.js
```diff
@@ -145,6 +151,16 @@ TraceWriter.prototype.getProjectId = function(callback) {
     property: 'project-id',
     headers: headers
   }, function(err, response, projectId) {
+    if (response && response.statusCode !== 200) {
+      if (response.statusCode === 503) {
+        err = new Error('Metadata service responded with a 503 status ' +
```
matthewloring: I know this error message was copied but I don't think the `please try again` is relevant for users. They have not taken any manual action that they could try to repeat.

src/trace-writer.js
```diff
@@ -186,10 +202,15 @@ TraceWriter.prototype.writeSpan = function(spanData) {
 TraceWriter.prototype.queueTrace_ = function(trace) {
   var that = this;
 
+  if (!that.isActive) {
+    return;
```
matthewloring: This means that we will continue attempting to queue traces even when we have failed to initialize correctly. Is there a way that this bailing out could be done earlier in the process, maybe in the plugins, so that we avoid propagating context (which incurs most of the performance overhead) if we know the agent has turned off?

matthewloring: One option would be for the `bind` and `bindEmitter` operations to just no-op if the agent is disabled. This wouldn't be a nice as short-circuiting out of the plugin instrumentation but may avoid the majority of the performance overhead. It would be worth thinking about the design of this disabling mechanism without plugin unpatching in more detail to make sure we're getting it right.

kjin: I see - I think this is another compelling case for keeping "phantom" functionality in the sense that the agent object returned by `start()` and `get()` should opaquely auto-switch between no-op and full implementation depending on the status of `projectID`'s existence.

Another solution would be to somehow modify the trace policy to never allow traces, which for our plugins right now would effectively short-circuit. I'm not sure whether that's a reasonable long-term solution, but it does prevent context propagation as well.

In my mind, the exact behavior we want is: as soon as the trace agent knows that it's impossible to obtain the project ID, do not queue traces for publishing. When it comes to preventing as much unnecessary work as possible and ease of implementation, I think changing the tracing policy is the most effective way... what do you think?

matthewloring: As discussed in person, we will move back toward an unpatching/`stop` approach but `stop` will still be removed from the public interface.

kjin:

src/trace-agent.js
```diff
@@ -66,7 +66,8 @@ function TraceAgent(config, logger) {
 TraceAgent.prototype.stop = function() {
   hooks.deactivate();
   pluginLoader.deactivate();
-  cls.destroyNamespace();
+  this.config.enabled = false;
```
ofrobots: typo? `this.config_`

src/hooks/index.js
```diff
@@ -120,7 +121,10 @@ function checkLoadedModules(logger) {
 }
 
 function activate(agent) {
-
+  if (activated) {
```
matthewloring: In what case is this multiply activated?

src/trace-agent.js
```diff
@@ -66,7 +66,8 @@ function TraceAgent(config, logger) {
...
+  this.policy = tracingPolicy.createTracePolicy(this.config);
```
matthewloring: This is a roundabout way to get a policy that will never trace. Maybe we could add an explicit accessor for the `TraceNone` policy? Also, why is this policy change needed now that we've restored the unpatching of modules?

test/test-agent-metadata.js
```diff
@@ -61,13 +57,40 @@ describe('agent interaction with metadata service', function() {
     }, 500);
   });
 
+  it('should preserve public interface when stopped', function(done) {
+      nock.disableNetConnect();
```
matthewloring: It may be a better separation of concerns to break this into two tests (has same interface when disabled which manually calls `stop`, and another that ensures that the agent stops on configuration failure).

test/test-index.js
```diff
@@ -27,168 +27,22 @@ var cls = require('../src/cls.js');
...
-    agent.stop();
-    assert(!nodule[property].__unwrap,
-      property + ' should get unwrapped on stop');
-  }
-
-  it('should wrap/unwrap module._load on start/stop', function() {
-    wrapTest(agent, require('module'), '_load');
-  });
-
-  it('should not attach exception handler with ignore option', function() {
```
matthewloring: why was this test removed?

src/hooks/index.js
kjin: Oh - it was being activated multiple times in an intermediate form of a test. I've since refactored that test so this does not apply anymore.

src/trace-agent.js
kjin: Good catch, fixed

src/trace-agent.js
kjin: For plugins that don't unpatch anything, this would let them at the very least short-circuit (since spans will never be created).

kjin: I would be in favor of an explicit `TraceNone` policy.

test/test-agent-metadata.js
kjin: The test above this one already checks the latter. However, it uses the private `stop()` method which does not affect the phantomness of the public agent, so manually calling stop doesn't work here.

test/test-index.js
kjin: Oops, it became a victim of blanket delete. I've restored it.

src/trace-agent.js
matthewloring: sgtm

src/trace-agent.js
```diff
@@ -66,7 +66,7 @@ function TraceAgent(config, logger) {
 TraceAgent.prototype.stop = function() {
   hooks.deactivate();
   pluginLoader.deactivate();
-  cls.destroyNamespace();
+  this.policy = new tracingPolicy.TraceNonePolicy();
```
ofrobots: Add a comment here explaining why this is necessary – namely that unpatching may not necessarily completely remove trace generation from instantiated objects.

kjin: Done, thanks!

PR #379, opened 2017-02-10
Support hapi tracing through the plugin API

DominicKramer:


DominicKramer:
PTAL

src/plugins/plugin-hapi.js
```diff
@@ -0,0 +1,96 @@
...
+    // req.path would be more desirable but is not set at the time our middlewear runs.
+    transaction.addLabel(TraceLabels.HTTP_METHOD_LABEL_KEY, req.method);
+    transaction.addLabel(TraceLabels.HTTP_URL_LABEL_KEY, url);
+    transaction.addLabel(TraceLabels.HTTP_SOURCE_IP, req.connection.remoteAddress);
+
+    var context = transaction.getTraceContext();
+    res.setHeader(constants.TRACE_CONTEXT_HEADER_NAME, context);
+
+    // wrap end
+    res.end = function(chunk, encoding) {
```
AdriVanHoudt: In hapi there is a concept of `tails` where you can still do stuff in your handler after you replied.
In this case you end the span when you call `reply`. I have some routes where I do some queries after replying for performance reasons and it would be great if those queries would also end up in the trace of that request.
Would it be possible to integrate this? You have access to the full request object on that event and it is fired only once per request.
See https://github.com/hapijs/hapi/blob/master/API.md#server-events and https://github.com/hapijs/hapi/blob/master/API.md#requesttailname for more info.
And https://pbs.twimg.com/media/CL-bp3_W8AAimRi.png:large for a nice overview of how the hapi lifecycle works

src/plugins/plugin-hapi.js
```diff
@@ -0,0 +1,96 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+var shimmer = require('shimmer');
+var urlParse = require('url').parse;
+
+var constants = require('../constants.js');
+var TraceLabels = require('../trace-labels.js');
```
kjin: Can you use `api.constants` and `api.labels` instead?

src/plugins/plugin-hapi.js
```diff
@@ -0,0 +1,96 @@
...
+  var res = request.raw.res;
+  var originalEnd = res.end;
+  var options = {
+    name: urlParse(req.url).pathname,
+    traceContext: req.headers[constants.TRACE_CONTEXT_HEADER_NAME],
+    skipFrames: 3
+  };
+  api.runInRootSpan(options, function(transaction) {
+    if (!transaction) {
+      // TODO: Determine if this message is still needed
```
kjin: It's not needed.

src/plugins/plugin-hapi.js
DominicKramer: Thanks for the feedback.  Its greatly appreciated.  The change you are suggesting is out of the scope of this PR, but I've made an issue ([382](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/issues/382)) to track it.

src/plugins/plugin-hapi.js
DominicKramer: Done

src/plugins/plugin-hapi.js
DominicKramer: Done

PR #380, opened 2017-02-10
Support Koa tracing through the plugin API

DominicKramer:


DominicKramer:
PTAL

src/plugins/plugin-koa.js
```diff
@@ -0,0 +1,105 @@
...
+    root.addLabel(api.labels.HTTP_URL_LABEL_KEY, url);
+    root.addLabel(api.labels.HTTP_SOURCE_IP, req.connection.remoteAddress);
+
+    var context = root.getTraceContext();
+    if (context) {
+      res.setHeader(api.constants.TRACE_CONTEXT_HEADER_NAME, context);
+    }
+    //else {
+    //  TODO: Determine if this is still needed
+    //  agent.logger.warn('koa: Attempted to generate trace context for nullSpan');
```
matthewloring: This log is no longer needed since `runInRootSpan` returns `null` now if a span should not be created instead of a `nullSpan`. We should think about how we want to unify this method if indicating failure with the use of `nullSpan` on the existing instrumentation api. We originally used `nullSpan` so that people instrumenting their own code wouldn't need to wrap all operations on spans in a null check. @kjin just want to put this on your radar.

DominicKramer: Done

PR #381, opened 2017-02-10
Support Connect tracing through the plugin API

DominicKramer:


DominicKramer:
PTAL

src/plugins/plugin-connect.js
```diff
@@ -0,0 +1,90 @@
...
+}
+
+module.exports = [
+  {
+    file: '',
+    versions: SUPPORTED_VERSIONS,
+    intercept: function(connect, api) {
+      return function() {
+        var app = connect();
+        app.use(middleware.bind(null, api));
```
matthewloring: Prefer a global `api` object to a `bind` here. It will interfere with the value of `this` in the middleware (though this is not a concern in this case) and will have a high performance overhead on some node versions (@ofrobots may remember which ones).

src/plugins/plugin-connect.js
```diff
@@ -0,0 +1,90 @@
...
+function middleware(api, req, res, next) {
+  var originalEnd = res.end;
+  var options = {
+    name: urlParse(req.originalUrl).pathname,
+    traceContext: req.headers[api.constants.TRACE_CONTEXT_HEADER_NAME.toLowerCase()],
+    skipFrames: 3
+  };
+  api.runInRootSpan(options, function(root) {
+    if (!root) {
+      // TODO: Determine if this message is needed
```
matthewloring: This is not needed as the `runInRootSpan` call will generate it if needed.

src/plugins/plugin-connect.js
```diff
@@ -0,0 +1,90 @@
...
+ * limitations under the License.
+ */
+'use strict';
+
+var urlParse = require('url').parse;
+
+var SUPPORTED_VERSIONS = '3.x';
+
+function middleware(api, req, res, next) {
+  var originalEnd = res.end;
```
matthewloring: Was there a reason this was hoisted?

src/plugins/plugin-connect.js
```diff
@@ -0,0 +1,90 @@
...
+    root.addLabel(api.labels.HTTP_METHOD_LABEL_KEY, req.method);
+    root.addLabel(api.labels.HTTP_URL_LABEL_KEY, url);
+    root.addLabel(api.labels.HTTP_SOURCE_IP, req.connection.remoteAddress);
+
+    var context = root.getTraceContext();
+    if (context) {
+      res.setHeader(api.constants.TRACE_CONTEXT_HEADER_NAME, context);
+    } else {
+      // TODO: Determine if this message is still needed
+      // agent.logger.warn('Connect: Attempted to generate trace context for nullSpan');
```
matthewloring: This case should also no longer be necessary.

src/plugins/plugin-connect.js
DominicKramer: I don't think there is a reason for it to be hoisted.  I've moved it down to the redefinition of `res.end`.

src/plugins/plugin-connect.js
DominicKramer: Ok.  I've removed it.

src/plugins/plugin-connect.js
DominicKramer: I've removed it.

src/plugins/plugin-connect.js
matthewloring: We can revisit this decision in a follow on as well.

kjin: I would suggest putting `middleware` within the scope of `module.exports.intercept`. Not sure if that occurs any penalty versus having a file-global `api` object.

PR #385, opened 2017-02-14
Add docker start/stop script for DB unit tests

kjin:
This change adds a script to run docker containers to support local unit testing. It's not needed on any CI.

`./bin/docker-trace.sh start` starts docker containers listening locally for mongoDB, redis, and MySQL connections.

`./bin/docker-trace.sh stop` stops them (though, in a somewhat unsafe way).

### DIFF
--- a/bin/docker-trace.sh
+++ b/bin/docker-trace.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+COMMAND=$1
+if [ $COMMAND = 'start' ]; then
+  docker run -p 127.0.0.1:27017:27017 -d mongo
+  docker run -p 127.0.0.1:6379:6379 -d redis
+  docker run -p 127.0.0.1:3306:3306 -e MYSQL_ROOT_PASSWORD='Password12!' -e MYSQL_DATABASE=test -d mysql
+fi
+if [ $COMMAND = 'stop' ]; then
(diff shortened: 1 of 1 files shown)


bin/docker-trace.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+COMMAND=$1
+if [ $COMMAND = 'start' ]; then
+  docker run -p 127.0.0.1:27017:27017 -d mongo
+  docker run -p 127.0.0.1:6379:6379 -d redis
+  docker run -p 127.0.0.1:3306:3306 -e MYSQL_ROOT_PASSWORD='Password12!' -e MYSQL_DATABASE=test -d mysql
+fi
+if [ $COMMAND = 'stop' ]; then
+  docker stop `docker ps | sed -En 's/([a-f0-9]+)[[:space:]]+(redis|mongo|mysql).*/\1/p'`
```
DominicKramer: Another option would be to use `--name trace-test-mongo`, `--name trace-test-redis` etc. when using `docker run`.  Then the command 
```
docker stop `docker ps -q --filter name="trace-test-*"`
```
could be used to stop everything started for trace testing.  This would avoid stopping a `redis`, `mongo`, or `mysql` container that was started for a purpose other than trace testing.

ofrobots: +1

kjin: Done

PR #388, opened 2017-02-16
Fix for external unit test failures

kjin:
The combination of having a plugin for a certain module (for express, restify, presumably others) and changes in how the agent is started and stopped seem to produce failures in the unit tests for that module. This was observed specifically in express and restify (in the latter, the error code was not reported correctly). Note that these tests are only run on CI in Node v0.12, but the same issues arise in other Node versions.

This fix should resolve that, but in the meantime, I'm looking into why this is the case.

### DIFF
--- a/test/non-interference/start-agent.js
+++ b/test/non-interference/start-agent.js
@@ -15,6 +15,20 @@
  */
 'use strict';
 
+var shimmer = require('shimmer');
+var agent = require('../../src/trace-agent.js');
+// Stub getTraceContext so that it always returns the same thing.
+shimmer.wrap(agent, 'get', function(original) {
+  return function() {
+    var privateAgent = original.apply(this, arguments);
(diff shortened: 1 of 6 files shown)


kjin:

This became a number of grouped changes for allowing express tests to pass.

test/non-interference/start-agent.js
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+var shimmer = require('shimmer');
+var agent = require('../../src/trace-agent.js');
+// Stub getTraceContext so that it always returns the same thing.
```
matthewloring: Could you add a little more to this comment about why this is necessary? The bug you discovered is quite nuanced.

PR #389, opened 2017-02-17
Bugfix for creating root spans through plugin API

kjin:
`shouldTrace` isn't being called unless `options.url` is defined. This fixes that, so that `shouldTrace` is always being called.

In addition, `options.url` is used in place of `options.name` if it's not defined. This makes it so that if a trace's name is its URL, both fields don't have to be supplied.

### DIFF
--- a/src/trace-plugin-interface.js
+++ b/src/trace-plugin-interface.js
@@ -195,7 +195,7 @@ PluginAPI.prototype.createChildSpan = function(options) {
   var rootSpan = this.getRootSpan();
   if (rootSpan) {
     options = options || {};
-    var childContext = this.agent_.startSpan(options.name, {},
+    var childContext = this.agent_.startSpan(options.name || options.url, {},
       options.skipFrames ? options.skipFrames + 1 : 1);
     return new ChildSpan(this.agent_, childContext);
   } else {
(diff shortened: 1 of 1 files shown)


matthewloring:
We should have unit test for this in addition to the testing provided by the `test-trace-<framework>` files. It would be good to add a unit test that fails without this fix.

matthewloring:
We can add the test as a follow on since we do not currently have an infrastructure for testing the plugin interface. Once we replace the existing custom span api with the plugin interface then we can add tests like this to `test-index.js`.

src/trace-plugin-interface.js
```diff
@@ -195,7 +195,7 @@ PluginAPI.prototype.createChildSpan = function(options) {
   var rootSpan = this.getRootSpan();
   if (rootSpan) {
     options = options || {};
-    var childContext = this.agent_.startSpan(options.name, {},
+    var childContext = this.agent_.startSpan(options.name || options.url, {},
```
matthewloring: If `options.name` is not optional, I'm not sure why we need a fallback here.

kjin: Done

PR #390, opened 2017-02-17
Add gRPC Plugin

kjin:
This change adds gRPC tracing as a plugin. It's dependent on #389.

Side-by-side diff of `hook-grpc.js` and `plugin-grpc.js` can be found here:
https://github.com/kjin/cloud-trace-nodejs/pull/7/files (it may be helpful to add [`?w=1`](https://github.com/kjin/cloud-trace-nodejs/pull/7/files?w=1) to the end of that)


src/plugins/plugin-grpc.js
```diff
@@ -0,0 +1,433 @@
...
+'use strict';
+
+var shimmer = require('shimmer');
+var findIndex = require('lodash.findindex');
+
+var SKIP_FRAMES = 3;
+
+function patchClient(client, api) {
+  // Wraps a callback so that the current span for this trace is also ended when
+  // the callback (done) is invoked.
```
matthewloring: Could we re-add the jsdoc types here?

src/plugins/plugin-grpc.js
```diff
@@ -0,0 +1,433 @@
...
+          break;
+        case 'server_stream':
+          wrapServerStream(handlerSet, requestName);
+          break;
+        case 'client_stream':
+          wrapClientStream(handlerSet, requestName);
+          break;
+        case 'bidi':
+          wrapBidi(handlerSet, requestName);
+          break;
```
matthewloring: Can we preserve the `default` case here?

kjin: Reading through the gRPC server code - I noticed that there is no code path that provides `register` with a value other than the ones listed in cases, so the `default` case would be dead code. In addition - the application would crash if the `method_type` was an invalid value. Would the `default` case still be necessary here?

PR #392, opened 2017-02-18
make test-hooks-sample-warning functional again

ofrobots:


src/trace-plugin-interface.js
```diff
@@ -199,8 +203,9 @@ PluginAPI.prototype.createChildSpan = function(options) {
       options.skipFrames ? options.skipFrames + 1 : 1);
     return new ChildSpan(this.agent_, childContext);
   } else {
-    this.logger_.warn(options.name + ': Attempted to create child span ' +
-      'without root');
+    var pluginName = options.pluginName || 'unknown plugin';
```
kjin: Since this interface will be directly used as the custom tracing API - is there a more generalizable name than `pluginName`?

kjin: Either that, or use `'custom trace'` instead of `'unknown plugin'`? Though that still doesn't cover the case of plugins that don't provide this field.

ofrobots: open to suggestions!

kjin: Actually `pluginName` seems fine, and the default could just be 'trace agent' or nothing at all.

src/trace-plugin-interface.js
```diff
@@ -199,8 +203,9 @@ PluginAPI.prototype.createChildSpan = function(options) {
...
+    this.logger_.debug(options.pluginName +
```
matthewloring: Should this be `pluginName`?

src/trace-plugin-loader.js
```diff
@@ -95,6 +92,11 @@ function activate(agent) {
         }
         instrumentation.patches[moduleRoot] = patchSet;
       }
+
+      // Create a new object exposing functions to create trace spans and
+      // propagate context. This relies on functions currently exposed by the
+      // agent.
+      var api = new PluginAPI(agent, moduleRoot);
```
kjin: `moduleRoot` is not the correct value to use here. It should be the name of the module, but `moduleRoot` is the path to the module. In the case of `http` it is `null`.

Unfortunately, the name of the module is not accessible in this scope, so you'll have to add a `name: moduleName` field in `plugins[moduleName]` at the top of this function. Then you can use `instrumentation.moduleName`.

kjin: Once this change is made, `test-grpc-context.js` will fail because it's checking the second argument of `logger.debug`, which now doesn't contain a value.

kjin: LMK if you want me to upload a fix for this.

kjin: Try: https://github.com/kjin/cloud-trace-nodejs/commit/161e5cddec222297304fc8b1b185b215a1d53e6c

matthewloring:
These changes are now subsumed by https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/411.

PR #393, opened 2017-02-18
Refactor public api to match plugin api

### LINKED ISSUE #330: runInRootSpan always creates a context
`runInRootSpan` always creates a trace context. This means that incoming trace context cannot be propagated by users of the custom tracing api.

One way to fix this would be to accept optional arguments in `runInRootSpan` with the optional context if the caller has one already.

/cc @evanlucas.

matthewloring:
Fixes #330


matthewloring:
PTAL.

src/trace-plugin-interface.js
```diff
@@ -184,15 +157,14 @@ PluginAPI.prototype.runInRootSpan = function(options, fn) {
 };
 
 /**
- * Creates and returns a new ChildSpan object nested within the root span object
- * returned by getRootSpan. If there is no current RootSpan object, this
- * function returns null.
+ * Creates and returns a new ChildSpan object nested within the root span.If
```
kjin: Nit: Space after period

src/trace-plugin-interface.js
```diff
@@ -184,15 +157,14 @@ PluginAPI.prototype.runInRootSpan = function(options, fn) {
...
+ * there is no current RootSpan object, this function returns null.
  * @param {object} options An object that specifies options for how the child
  * span is created and propogated.
  * @returns A new ChildSpan object, or null if there is no active root span.
  */
 PluginAPI.prototype.createChildSpan = function(options) {
-  var rootSpan = this.getRootSpan();
+  var rootSpan = this.getRootSpan_();
```
kjin: Do you think it would be better to make this module-private (like `createRootSpan_`)?

index.js
```diff
@@ -43,28 +44,19 @@ for (var i = 0; i < filesLoadedBeforeTrace.length; i++) {
...
-  },
-  runInRootSpan: function(name, labels, fn) {
-    if (typeof(labels) === 'function') {
-      fn = labels;
-    }
-    fn(function() {});
-  },
-  setTransactionName: function() {},
-  addTransactionLabel: function() {}
+  enhancedDatabaseReportingEnabled: function() { return false; },
```
kjin: I think there are now functions within the private agent (in `trace-agent.js`) that are completely dead now. Maybe that could go in a separate cleanup PR, but they should be removed at some point.

matthewloring: Hmm, I thought I cleaned up the now dead functions. Which ones were you thinking of?

src/trace-plugin-interface.js
matthewloring: Updated.

index.js
matthewloring: I think I cleaned up the dead code. Which operations did you have in mind?

kjin: Oops, I missed that part. Never mind!

index.js
```diff
@@ -43,28 +44,19 @@ for (var i = 0; i < filesLoadedBeforeTrace.length; i++) {
...
+  runInRootSpan: function(opts, fn) { return fn(PluginApi.nullSpan); },
+  createChildSpan: function(opts) { return PluginApi.nullSpan; },
+  wrap: function(fn) { return fn; },
+  wrapEmitter: function(ee) { return ee; },
+  constants: constants,
+  labels: TraceLabels
 };
 
 /** @private */
```
ofrobots: Add JSDoc that this would either be a PhantomAgent or a PluginAPI object. Also, the name `agent` seems wrong when this can be an object of the type `PluginAPI`. I am guessing this is going to be cleaned up in a follow-on?

matthewloring: Yup, we will handle the renaming away from `agent` in a follow on.

DominicKramer:
LGTM

PR #394, opened 2017-02-18
Add stack trace to service network errors

### LINKED ISSUE #349: Add stack traces to logged errors
Errors logged by the agent coming from the trace service do not include the actual error object. This prevents them from being detected by tools that scan stdout/stderr such as `pm2`. We should include the full error object if it is available.

Original discussion here: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/issues/342.

/cc @AdriVanHoudt

matthewloring:
Fixes #349

### DIFF
--- a/src/trace-writer.js
+++ b/src/trace-writer.js
@@ -256,7 +256,7 @@ TraceWriter.prototype.publish_ = function(projectId, json) {
   }, function(err, response, body) {
     if (err) {
       that.logger_.error('TraceWriter: error: ',
-        (response && response.statusCode) || '', err);
+        ((response && response.statusCode) || '') + '\n' +  err.message + '\n' + err.stack);
     } else {
       that.logger_.info('TraceWriter: published. statusCode: ' + response.statusCode);
     }


matthewloring:
PTAL.

src/trace-writer.js
```diff
@@ -256,7 +256,7 @@ TraceWriter.prototype.publish_ = function(projectId, json) {
   }, function(err, response, body) {
     if (err) {
       that.logger_.error('TraceWriter: error: ',
-        (response && response.statusCode) || '', err);
+        (response && response.statusCode) || '' + '\n' +  err.message + '\n' + err.stack);
```
kjin: I think you need parentheses around `(response && response.statusCode) || ''` if a stack trace should be reported regardless of whether there was a response or not.

kjin: I think you need parentheses around `(response && response.statusCode) || ''` if a stack trace should be reported regardless of whether there was a response or not.

matthewloring: https://jsconsole.com/?var%20o%20%3D%20undefined%3B%20(o%20%26%26%20o.statusCode)%20%7C%7C%20%27%27%20%2B%20%27hi%27%20%2B%20%27there%27%3B

matthewloring: Sorry, good catch, fixed.

src/trace-writer.js
```diff
@@ -256,7 +256,7 @@ TraceWriter.prototype.publish_ = function(projectId, json) {
   }, function(err, response, body) {
     if (err) {
       that.logger_.error('TraceWriter: error: ',
-        (response && response.statusCode) || '', err);
+        ((response && response.statusCode) || '') + '\n' +  err.message + '\n' + err.stack);
```
ofrobots: Doesn't `err.stack` include `err.message`?

$ node -p "(new Error('foo')).stack"
Error: foo
    at [eval]:1:2
    at ContextifyScript.Script.runInThisContext (vm.js:26:33)
    at Object.exports.runInThisContext (vm.js:79:17)
    at Object.<anonymous> ([eval]-wrapper:6:22)
    at Module._compile (module.js:571:32)
    at Immediate.<anonymous> (bootstrap_node.js:383:29)
    at runCallback (timers.js:651:20)
    at tryOnImmediate (timers.js:624:5)
    at processImmediate [as _immediateCallback] (timers.js:596:5)

matthewloring: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/412

PR #395, opened 2017-02-18
cleanup todos

matthewloring:


matthewloring:
PTAL.

src/trace-plugin-interface.js
```diff
@@ -102,7 +102,6 @@ RootSpan.prototype.getTraceContext = function() {
 /**
  * PluginAPI constructor. Don't call directly - a plugin object will be passed to
  * plugin themselves
- * TODO(kjin): Should be called something else
```
kjin: Though it's been renamed since I added this TODO (from `Plugin` to `PluginAPI`), I still think this should be renamed, maybe to `TraceInterface` or something?

src/plugins/plugin-redis.js
```diff
@@ -94,12 +94,13 @@ function startSpanFromArguments(api, cmd, args, cb, send_command) {
...
   return function internalSendCommandWrap(internal_send_command) {
-    // TODO: Document and simplify this code.
     return function internal_send_command_trace(cmd, args, cb) {
       if (arguments.length === 1 && typeof cmd === 'object') {
+        // New versions of redis (2.4+) use a single options object instead
+        // of separate named arguments.
         var span = setupSpan(api, cmd.command, cmd.args, 0);
         if (!span) {
-          return internal_send_command.call(this, cmd, args, cb);
+          return internal_send_command.call(this, cmd);
```
kjin: I don't want to block this PR, but I think there is even more that can be done to simplify this code.

kjin: Namely, the helper `setupSpan` can be completely collapsed into `startSpanFromArguments`. Do you want me to open a follow-up PR for that?

test/test-trace-agent.js
```diff
@@ -34,17 +34,13 @@ var assert = require('assert');
 var config = require('../config.js');
 var file = require('../src/trace-agent.js');
 var SpanData = require('../src/span-data.js');
-// TODO: This line will silently fail if a logger is not supplied as
-//       the second argument.
```
kjin: I think this TODO refers to `TraceAgent` not accounting for the second argument not being present. I know `TraceAgent` isn't exposed publicly but it would still be worth addressing this so future tests don't also cause the agent to silently fail.

matthewloring: The second argument not being present is a legitimate failure condition for the agent. Before the TODO was introduced the tests did not pass a logger but it worked out due to the limited use of the agent in the tests. Adding the mock logger addresses this issue.

kjin: OK, makes sense.

src/trace-plugin-interface.js
matthewloring: I'm going to hold off landing this until the `PluginAPI` refactoring is done. Then I'll rename to `TraceAPI` as discussed in person.

matthewloring: This renaming will occur in https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/411. No longer blocked.

PR #396, opened 2017-02-18
Correct broken test

matthewloring:


matthewloring:
PTAL

test/test-agent-stopped.js
```diff
@@ -19,16 +19,20 @@
...
-
 describe('test-agent-stopped', function() {
   var agent;
   before(function() {
+    delete process.env.GCLOUD_PROJECT;
     agent = require('..').start();
   });
 
+  after(function() {
+    process.env.GCLOUD_PROJECT = 0;
```
ofrobots: Not a comment for this PR, but in general, we should move away from depending on this env. in the tests as much as possible. Only the tests that are checking picking up the project number from the environment should use this. All else should use the config mechanism to pass in a project id.

matthewloring: Agreed, that would clean things up.

PR #397, opened 2017-02-18
Instrument multiple versions of the same module

matthewloring:


matthewloring:
PTAL

test/hooks/test-hooks-multi-version.js
```diff
@@ -0,0 +1,68 @@
...
+    });
+  });
+
+  it('should record spans', function(done) {
+    common.runInTransaction(agent, function(endTransaction) {
+      clientv0.get('v0', function(err, n) {
+        clientv2.get('v2', function(err, n) {
+          endTransaction();
+          var spans = common.getMatchingSpans(agent, redisPredicate.bind(null, 'redis-get'));
+          assert.equal(spans.length, 2);
```
kjin: Optional suggestion: check `labels.arguments` to make sure the arguments to these calls are `v0` and `v2` respectively. Would require `enhancedDatabaseReporting` to be set.

matthewloring: Done.

PR #398, opened 2017-02-19
Dead code elimination :(

matthewloring:


matthewloring:
PTAL

matthewloring:
Use of this was lost during the hook -> plugin transition. We should fix this in the plugins as opposed to removing it.

PR #399, opened 2017-02-19
Remove unused config option

matthewloring:


matthewloring:
Enforcement of this config option was lost during the hook -> plugin transition. We should fix this in the plugins as opposed to removing the option.

PR #400, opened 2017-02-19
Add database result summarization to plugins

matthewloring:


matthewloring:
I'm not convinced we want this operation on the plugin api but this PR will restore the functionality we had in the old hooks.

matthewloring:
PTAL

PR #401, opened 2017-02-19
simplify interop-mongo-express

matthewloring:


matthewloring:
This can be simplified now that we run all test files in their own process.

matthewloring:
PTAL

PR #402, opened 2017-02-19
Experimental postgres plugin

matthewloring:


src/plugins/plugin-pg.js
```diff
@@ -0,0 +1,78 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+var shimmer = require('shimmer');
+
+var SUPPORTED_VERSIONS = '^6.x';
+
+function makeQueryWrap(api) {
```
kjin: Suggestion: can we capture `api` in a closure here and in `wrapCallback`?

src/plugins/plugin-pg.js
```diff
@@ -0,0 +1,78 @@
...
+}
+
+function wrapCallback(api, span, done) {
+  var fn = function(err, res) {
+    if (api.enhancedDatabaseReportingEnabled()) {
+      if (err) {
+        span.addLabel('error', err);
+      }
+      if (res) {
+        // TODO(mattloring): summarize once it is available on the public api
```
kjin: I think with recent changes this TODO can go away, right?

kjin:
If it helps, you can apply this patch on top of this PR once you rebase: https://github.com/kjin/cloud-trace-nodejs/commit/dd50a45f44e07ea819bf15000aa884d9a0a95a2d

src/plugins/plugin-pg.js
matthewloring: Yup. removed.

src/plugins/plugin-pg.js
matthewloring: Simplified and moved into the closure.

matthewloring:
Thank you very much for putting together the patch you linked! On cancelation, if the request has been buffered but not sent, then no async work has actually been done and there is no span to report. If the request has been sent and is canceled in flight then I think it should resolve normally with information in the response indicating that it was canceled (we should verify this though).

PR #404, opened 2017-02-20
Context propagation for google-gax

matthewloring:
This should allow tracing of pubsub, vision, monitoring, spanner,
speech, and nlp.

### DIFF
--- a/src/plugins/plugin-google-gax.js
+++ b/src/plugins/plugin-google-gax.js
@@ -1,0 +1,50 @@
+/**
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+var shimmer = require('shimmer');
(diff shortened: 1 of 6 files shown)


matthewloring:
@kjin Did you get a chance to look through the gax implementation while reviewing this? There are usually many possible patches that could be applied to preserve context. I want to make sure this approach makes the most sense in addition to solving the problem.

jmuk:
I am not familiar with how the wrapping works, but patching like this would be great. GAX also has a logic of retrying or bundling inside of the `funcWithAuth` promise result, which could destroy the callstack unfortunately.

I am also wondering how refactoring gax could save this situation (in other words, what types of changes on google-gax side can simplify the patch like this). Let me know if you have any proposals.

src/plugins/plugin-google-gax.js
```diff
@@ -0,0 +1,48 @@
...
+            var result = funcWithAuthThen.call(this, cb);
+            var resultThen = result.then;
+            result.then = function(cb) {
+              return resultThen.call(this, api.wrap(cb));
+            };
+            return result;
+          };
+          var apiCallInner = createApiCall.call(this, funcWithAuth, settings, optDescriptor);
+          return function apiCallInner_trace(request, callOptions, callback) {
+            return apiCallInner.call(this, request, callOptions, api.wrap(callback));
```
kjin: Looking at this in more detail, I think the `api.wrap(callback)` is not needed since context can be automatically propagated to `callback` from here. In other words, `createApiCall_trace` can directly return `createApiCall.call(this, funcWithAuth, settings, optDescriptor)`.

kjin: Talked in person, and it seems like this is still necessary for propagating context when control is yielded back to the user.

matthewloring: I've added a comment to make this explicit.

matthewloring:
@jmuk I believe there may be opportunities to refactor gax and simplify things here but I'll have to dig a little further.

PR #406, opened 2017-02-21
Consolidate usage of `agent.private_()`

DominicKramer:
With this PR, `agent.private_()` is only used directly in `common.js`.  Thus, tests only use functionality provided through `agent.private_()` through functions exposed through `common.js`.

### DIFF
--- a/test/hooks/common.js
+++ b/test/hooks/common.js
@@ -208,6 +208,24 @@ function createChildSpan(agent, cb, duration) {
   };
 }
 
+function createRootSpanData(agent, name, traceId, parentId, skipFrames,
+                            spanKind) {
+  return agent.private_().createRootSpanData(name, traceId, parentId,
+                                             skipFrames, spanKind);
+}
+
+function getTraceWriter(agent) {
+  return agent.private_().traceWriter;
+}
+
+function getConfig(agent) {
+  return agent.private_().config();
+}
+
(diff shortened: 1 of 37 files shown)


test/hooks/common.js
```diff
@@ -21,7 +21,7 @@ if (!process.env.GCLOUD_PROJECT) {
 }
 
 // We want to disable publishing to avoid conflicts with production.
-require('../../src/trace-writer').publish_ = function() {};
+// require('../../src/trace-writer').publish_ = function() {};
```
matthewloring: Why was this commented out?

test/test-trace-header-context.js
```diff
@@ -25,14 +25,14 @@ describe('test-trace-header-context', function() {
...
   var express;
   before(function() {
-    agent = require('..').start({ samplingRate: 0 }).private_();
+    agent = require('..').start({ samplingRate: 0 });
     express = require('./hooks/fixtures/express4');
   });
 
   afterEach(function() {
-    // TODO: Investigate why this is needed
+    // TODO: Investigate why this is needed on Node v0.12
```
matthewloring: Maybe we could expose a `common.resetNamespace` which does a `destroy` -> `create` on the agent namespace to avoid this access to `private_`?

PR #410, opened 2017-02-22
Remove old hooks mechanism

matthewloring:


test/test-agent-stopped.js
```diff
@@ -21,9 +21,14 @@ var http = require('http');
 
 describe('test-agent-stopped', function() {
   var agent;
-  before(function() {
+  before(function(done) {
     delete process.env.GCLOUD_PROJECT;
     agent = require('..').start();
+    // Wait 300ms for agent to fail getting remote project id.
+    setTimeout(function() {
+      assert.ok(!agent.isActive());
```
kjin: I think the other `assert.ok(!agent.isActive());`s at the top of each test can be removed.

kjin:
Can we move `test-trace-*.js` from `hooks` to `plugins`, either here or as another PR?

test/test-agent-stopped.js
kjin: Talked in person; we should leave this here in case anything changes between unit tests.

matthewloring:
Updated the naming of test files/directories to say `plugins`.

config.js
```diff
@@ -53,10 +53,6 @@ module.exports = {
       'restify': path.join(__dirname, 'src/plugins/plugin-restify.js')
     },
 
-    // Valid entries are:
-    // 'express', 'hapi', 'http', 'restify'
-    excludedHooks: [],
```
ofrobots: This is an API change. How do people exclude hooks going forward?

test/test-agent-stopped.js
```diff
@@ -21,9 +21,14 @@ var http = require('http');
 
 describe('test-agent-stopped', function() {
   var agent;
-  before(function() {
+  before(function(done) {
     delete process.env.GCLOUD_PROJECT;
     agent = require('..').start();
+    // Wait 300ms for agent to fail getting remote project id.
```
ofrobots: Where does the 300ms number come from? I think it would be better to nock the metadata API to reduce timer dependence here.

config.js
ofrobots: I guess you could argue that the API change preceded this commit, but the question still stands.

matthewloring: Plugins are now specified through the config. Users can specify in their config precisely which plugins they want.

test/test-agent-stopped.js
matthewloring: I am not aware of a way to have a callback fire after the agent has resolved the metadata whether or not it resolves using `nock`.

matthewloring: Using `nock` should allow us to reduce the timeout without increasing the flakiness but it would be great to have a solution that didn't rely on timing.

ofrobots: In the debug agent we expose an event emitter on the `Debuglet` to expose lifecycle events to the tests. The debuglet concept is a private concept, so it is easy to implement test-only interface on it.

matthewloring: That would simplify things here. I'd like to keep this PR as close to a mechanical refactoring as possible. I can add the event lifecycle as a follow on.

ofrobots: sgtm. It would be good to add a nock and get rid of the hard-coded time limit however.

matthewloring: I've added the nock but we cannot avoid the hard-coded time limit since nock still resolves asynchronously.

matthewloring: This will be cleaned up once we add events to the agent.

config.js
kjin: On that note -- how would users exclude plugins? Right now you'd have to obtain the value of `require('@google-cloud/trace/config.js').trace.plugins` and filter out the fields you don't want. Maybe we could expose this object somewhere else.

matthewloring: Good point. After an in person discussion, I'll put together a fix here: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/414.

PR #411, opened 2017-02-23
Make trace interface uniform between module exports and plugins

kjin:
This change:

- Makes `PluginAPI` a per-plugin object that can accept a plugin name
- Overlays a "public API" object over `PluginAPI` that can switch between it and a no-op implementation


index.js
```diff
@@ -83,160 +63,115 @@ var initConfig = function(projectConfig) {
   return config;
 };
 
+var agent = pluginInterface.create('custom');
```
matthewloring: Maybe this could be more descriptive? `'Custom Span API'`?

test/test-agent-stopped.js
```diff
@@ -21,9 +21,13 @@ var http = require('http');
 
 describe('test-agent-stopped', function() {
   var agent;
-  before(function() {
+  before(function(done) {
```
matthewloring: These changes will be obsolete after https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/410

src/trace-plugin-interface.js
```diff
@@ -219,9 +220,80 @@ PluginAPI.prototype.constants = constants;
...
+      wrapEmitter: function(ee) {
+        return api.wrapEmitter(ee);
+      },
+      constants: api.constants,
+      labels: api.labels,
+      isActive: function() {
+        return api.isActive();
+      },
+      initialize_: function(agent) {
+        api = new PluginAPI(agent, pluginName);
```
matthewloring: Do we want to add an assert to ensure this isn't called multiple times?

src/trace-plugin-interface.js
```diff
@@ -219,9 +220,80 @@ PluginAPI.prototype.constants = constants;
...
+      },
+      disable_: function() {
```
matthewloring: And here?

src/trace-plugin-interface.js
```diff
@@ -111,9 +111,10 @@ var nullSpan = {
  * plugin themselves
  * TODO(kjin): Should be called something else
  */
-function PluginAPI(agent) {
+function PluginAPI(agent, pluginName) {
   this.agent_ = agent;
   this.logger_ = agent.logger;
+  this.pluginName_ = pluginName;
```
matthewloring: Is this used?

matthewloring:
It would also be good to add a few tests to enforce the lifecycle of the agent (cannot be re-initialized after disabling).

index.js
kjin: That sounds like a good name, I'll change it to that

test/test-agent-stopped.js
kjin: OK, will address during rebase

src/trace-plugin-interface.js
kjin: It isn't, but I wanted to include it anyway to make the follow-up change that adds it to log messages more straightforward. In fact, I think most of that is small enough that I could do it in this PR.

matthewloring: Up to you. If you include the logic here, make sure the test changes from https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/392/ are also reflected.

src/trace-plugin-interface.js
kjin: I added error logging, though through the publicly exposed interface, neither of these functions will be called multiple times.

src/trace-plugin-interface.js
kjin: Same as ^

src/trace-plugin-interface.js
kjin: I'll just do it now, then.

kjin:
Since `fix for v0.12`:

* Rebased on master
* Addressed comments
* Added plugin name to logs emitted from within `trace-interface.js`
* Moved tests related to the custom span API to `test-trace-interface.js` (they also apply to the plugin interface as well)
* Added new tests for the custom span API to `test-trace-interface.js` as well
* Added fix for a test that was also addressed in #392

@matthewloring I've decided to keep the facade to two states without enforcing anything about state changing, since throwing an error complicates forcing a new agent between tests.

src/trace-api.js
```diff
@@ -174,15 +173,15 @@ PluginAPI.prototype.runInRootSpan = function(options, fn) {
...
   var rootSpan = getRootSpan_(this);
   if (rootSpan) {
     options = options || {};
     var childContext = this.agent_.startSpan(options.name, {},
-      options.skipFrames ? options.skipFrames + 1 : 1);
+      options.skipFrames ? options.skipFrames + 2 : 2);
     return new ChildSpan(this.agent_, childContext);
   } else {
-    this.logger_.warn(options.name + ': Attempted to create child span ' +
+    this.logger_.debug(this.pluginName_ + ': Attempted to create child span ' +
```
matthewloring: Could we add a `TODO` on this to fix once we can disambiguate context loss from sampling?

src/trace-api.js
```diff
@@ -208,20 +208,91 @@ PluginAPI.prototype.wrap = function(fn) {
...
+ * Finally, when the trace agent is stopped, this object enters the "disabled"
+ * state, and its underlying implementation is switched back to no-op.
+ * Currently, this only happens when the application's GCP project ID could
+ * not be determined from the GCP metadata service.
+ * This object's state changes strictly from uninitialized to initialized,
+ * and from initialized to disabled.
+ */
+module.exports = function TraceApi(pluginName) {
+  var impl = phantomApiImpl;
+  Object.assign(this, {
```
matthewloring: `Object.assign` is not available in 0.12.

src/trace-api.js
```diff
@@ -194,9 +193,10 @@ PluginAPI.prototype.createChildSpan = function(options) {
  * that are called asynchronously (for example, in a network response handler).
  * @param {function} fn A function to which to bind the trace context.
  */
-PluginAPI.prototype.wrap = function(fn) {
+TraceApiImplementation.prototype.wrap = function(fn) {
   if (!this.agent_.namespace) {
-    this.logger_.warn('Trace agent: No CLS namespace to bind function to');
+    this.logger_.warn(this.pluginName_ + ': No CLS namespace to bind ' +
```
matthewloring: I know this is a copy but could we fix the grammar. Maybe "No CLS namespace available to bind function"?

src/trace-plugin-loader.js
```diff
@@ -57,14 +57,18 @@ function activate(agent) {
...
-  var api = new PluginAPI(agent);
   var pluginConfig = agent.config().plugins;
   for (var moduleName in pluginConfig) {
+    // Create a new object exposing functions to create trace spans and
+    // propagate context. This relies on functions currently exposed by the
+    // agent.
+    var api = new TraceApi(moduleName);
+    api.enable_(agent);
     plugins[moduleName] = {
+      name: moduleName,
```
matthewloring: Where is this field used?

test/test-grpc-context.js
```diff
@@ -60,7 +61,7 @@ Object.keys(versions).forEach(function(version) {
 
       common.replaceDebugLogger(agent, function(error, uri) {
         if (error.indexOf('http') !== -1) {
-          assert.notStrictEqual(uri.indexOf('localhost'), -1);
+          httpDebugPrinted = true;
```
matthewloring: This doesn't check the same property as before. Why was this change needed?

src/trace-plugin-loader.js
kjin: Oops, it's an artifact of an earlier revision. Removing this.

test/test-grpc-context.js
kjin: The test was broken before (in a similar manner to `test-plugins-sample-warning.js`) since the condition `error.indexOf('http') !== -1` was never true. The old assert checks the format of the message being printed in the `http` hook; the format has since changed. Asserting in `afterEach` rather than here just makes sure we don't (erroneously) fail if additional debug-level messages are printed.

src/trace-api.js
kjin: Forgot about this -- rebasing now to use new `extend` depedency.

src/trace-api.js
kjin: Done

src/trace-api.js
kjin: Done

src/trace-api.js
```diff
@@ -207,20 +212,91 @@ PluginAPI.prototype.wrap = function(fn) {
...
 
-PluginAPI.prototype.constants = constants;
+TraceApiImplementation.prototype.constants = constants;
 
-PluginAPI.prototype.labels = TraceLabels;
+TraceApiImplementation.prototype.labels = TraceLabels;
 
-PluginAPI.nullSpan = nullSpan;
+TraceApiImplementation.prototype.isActive = function() {
+  return true;
```
matthewloring: Could we replace this with a check for the phantom api in the public interface?

test/test-trace-api.js
```diff
@@ -0,0 +1,183 @@
...
+
+'use strict';
+
+var assert = require('assert');
+var EventEmitter = require('events');
+var TraceAPI = require('../src/trace-api.js');
+var common = require('./plugins/common.js');
+var TracingPolicy = require('../src/tracing-policy.js');
+
+var config = Object.assign({}, require('../config.js').trace,
```
matthewloring: Also unsafe on 0.12.

src/trace-api.js
kjin: Done

test/test-trace-api.js
kjin: Done

kjin:
TBR: @ofrobots

src/trace-api.js
```diff
@@ -207,20 +212,86 @@ PluginAPI.prototype.wrap = function(fn) {
...
+  createChildSpan: function(opts) { return nullSpan; },
+  wrap: function(fn) { return fn; },
+  wrapEmitter: function(ee) {},
+  constants: constants,
+  labels: TraceLabels
+};
 
-module.exports = PluginAPI;
+/**
+ * Creates an object that provides an interface to the trace agent
```
ofrobots: nit: fix word wrapping for this comment.

test/plugins/common.js
```diff
@@ -235,7 +236,11 @@ function avoidTraceWriterAuth(agent) {
 }
 
 function stopAgent(agent) {
-  agent.private_().stop();
+  if (agent.isActive()) {
```
ofrobots: Ideally, once the comments I expressed in `index.js` are addressed, this function be removed.

index.js
```diff
@@ -215,16 +66,100 @@ var publicAgent = {
...
+      }
+      config.projectId = projectId;
+    });
+  } else if (typeof config.projectId !== 'string') {
+    logger.error('config.projectId, if provided, must be a string. ' +
+      'Disabling trace agent.');
+    return traceApi;
+  }
+
+  agent = require('./src/trace-agent.js').get(config, logger);
```
ofrobots: Any reason why this require is not done at the top?

Secondly, we have three variables here that have a strong dependence on each other. `agent`, `traceApi`, and `pluginLoader`. 

* First of all can we move `pluginLoader` into the agent? it should be a concept owned by the agent. Since there is a strong dependence on `pluginLoader.activate` and `deactivate`, when `agent.start` and `stop` are called respectively, it would better if `agent` was *responsible* for calling these? I see that you moved this from the agent to here. Is there a reason for doing this?
* Similarly, can we make the `agent` responsible for calling `traceApi.enable` and `disable` from `start` and `stop` respectively?
* Nit: we have three names for similar operations: `{start, enable, activate}` and `{stop, disable, deactivate}`. Do we really need all three?

src/trace-api.js
kjin: That puts it over 80 characters... I'll reword it though.

index.js
kjin: I agree that the function calls that occur when one starts or stops the agent should be coalesced in one place, and I think it should be in `index.js` because my thoughts were that `TraceApiImplementation` and `TraceAgent` would eventually be folded into a single object (because one is really just a thin layer over the other, and this extra layer is unnecessary). I pulled `pluginLoader` out to take a step in that direction, because `TraceApiImplementation` in my opinion should not be responsible for it.

For the sake of readability I'd be OK with moving it back (and also reducing the number of synonyms on `start` and `stop`). But hopefully that gives a better understanding of the rationale behind this move.

PR #414, opened 2017-02-23
Fix merging of nest configuration objects

matthewloring:


package.json
```diff
@@ -39,6 +39,7 @@
   "dependencies": {
     "@google-cloud/common": "^0.12.0",
     "continuation-local-storage": "^3.1.4",
+    "extend": "^3.0.0",
```
AdriVanHoudt: Why not use https://github.com/sindresorhus/object-assign, it will use the native Object.assign when present (afaik you only need the polyfill for 0.12)

matthewloring: We need to use the deep copying capability for nested objects in the config: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Deep_Clone.

AdriVanHoudt: ah, I see 💡 
I looked at https://github.com/sindresorhus/deep-assign thinking it would use Object.assing underneath but does not appear so, `extend` is fine then (although `deep-assign` seems less bloated)

test/test-config-plugins.js
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  it('should overwrite builtin plugins correctly', function() {
+    var agent = trace.start({forceNewAgent_: true, plugins: {
+      express: 'foo'
+    }});
+    var plugins = common.getConfig(agent).plugins;
+    assert.strictEqual(JSON.stringify(Object.keys(plugins)),
+      JSON.stringify(instrumentedModules));
+    var modulesWithoutExpress = instrumentedModules.filter(function(e) {
```
ofrobots: nit: more functionally..

```js
assert.ok(instrumentedModules.filter(...).every(...));

File: test/test-config-plugins.js [hunk 3]
```diff
@@ -0,0 +1,66 @@
...
+var instrumentedModules = ['connect', 'express', 'grpc', 'hapi', 'http', 'koa',
+  'mongodb-core', 'mysql', 'redis', 'restify'];
+
+describe('plugin configuration', function() {
+  it('should have correct defaults', function() {
+    var agent = trace.start({forceNewAgent_: true});
+    var plugins = common.getConfig(agent).plugins;
+    assert.strictEqual(JSON.stringify(Object.keys(plugins)),
+      JSON.stringify(instrumentedModules));
+    for (var i = 0; i < instrumentedModules.length; i++) {
```
ofrobots: likewise.. use Array.prototype.every

test/test-config-plugins.js
```diff
@@ -0,0 +1,66 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
```
matthewloring: Fixed.

test/test-config-plugins.js
matthewloring: Nice, much cleaner!

test/test-config-plugins.js
matthewloring: done.

PR #415, opened 2017-02-23
New configuration for globally controlled label value sizes

matthewloring:


src/span-data.js
```diff
@@ -71,7 +78,7 @@ function SpanData(agent, trace, name, parentSpanId, isRoot, skipFrames) {
         callSite.getFileName(), callSite.getLineNumber(),
         callSite.getColumnNumber()));
     });
-    this.span.setLabel(TraceLabels.STACK_TRACE_DETAILS_KEY,
+    this.addLabel(TraceLabels.STACK_TRACE_DETAILS_KEY,
```
kjin: Isn't this still being truncated? I thought we didn't want it to be.

matthewloring: Yes, fixed.

matthewloring:
Done.

config.js
```diff
@@ -29,9 +29,9 @@ module.exports = {
     // attached to spans representating database operations.
     enhancedDatabaseReporting: false,
 
-    // The maximum result size in characters to report on database spans if
-    // `enhancedDatabaseReporting` is enabled.
-    databaseResultReportingSize: 127,
+    // The maximum number of characters reported on a label value. This
+    // cannot exceed 16383, the maximum value accepted by the service.
+    maximumLabelValueSize: 512,
```
ofrobots: The previous value was 127. Should this be 511? Does it matter?

src/span-data.js
```diff
@@ -92,7 +97,10 @@ SpanData.prototype.createChildSpanData = function(name, skipFrames) {
 };
 
 SpanData.prototype.addLabel = function(key, value) {
-  this.span.setLabel(key, value);
+  key = traceUtil.truncate(key, constants.TRACE_SERVICE_LABEL_KEY_LIMIT);
+  var val = typeof value === 'object' ? util.inspect(value) : '' + value;
```
ofrobots: nit: This seems assymetric. You assign to the argument `key` but keep create an additional names for `value`. Keep local variables for both or neither.

src/span-data.js
```diff
@@ -72,7 +76,8 @@ function SpanData(agent, trace, name, parentSpanId, isRoot, skipFrames) {
         callSite.getColumnNumber()));
     });
     this.span.setLabel(TraceLabels.STACK_TRACE_DETAILS_KEY,
-      JSON.stringify({stack_frame: stackFrames}));
+      traceUtil.truncate(JSON.stringify({stack_frame: stackFrames}),
```
ofrobots: nit: add a comment that we are intentionally not respective `config.maxLabelValueSize` here.

src/span-data.js
ofrobots: If `value` is a Symbol, string concatenation will throw. Can you add tests for all non-object types?

test/test-util.js
```diff
@@ -21,29 +21,21 @@ var Module = require('module');
...
   it('should truncate objects larger than size', function() {
-    assert.equal(util.stringifyPrefix(o, 15),
-      '{a:5,b:hi,c:...');
+    assert.equal(util.truncate('abcdefghijklmno', 5), 'ab...');
   });
 
   it('should not truncate objects smaller than size', function() {
-    assert.equal(util.stringifyPrefix(o, 150),
-      '{a:5,b:hi,c:[Function],d:{e:{f:null,g:undefined,h:{0:1,1:2}}}}');
+    assert.equal(util.truncate('abcdefghijklmno', 50), 'abcdefghijklmno');
```
ofrobots: `strictEqual`.

config.js
matthewloring: Arbitrary, should not matter.

src/span-data.js
matthewloring: Done.

test/test-util.js
matthewloring: Done.

src/span-data.js
matthewloring: Done.

PR #417, opened 2017-02-24
Make TraceWriter a Service object

ofrobots:


test/nocks.js
```diff
@@ -0,0 +1,60 @@
...
+// In the future _=>true.
+function accept() {
+  return true;
+}
+
+function nockOAuth2(validator) {
+  validator = validator || accept;
+  return nock('https://accounts.google.com')
+      .post('/o/oauth2/token', validator)
+      .once()
```
matthewloring: Isn't `once` the default behavior on nocks?

test/nocks.js
```diff
@@ -0,0 +1,60 @@
...
+}
+
+function nockProjectId(reply) {
+  return nock('http://metadata.google.internal')
+    .get('/computeMetadata/v1/project/project-id')
+    .once()
+    .reply(reply);
+}
+
+function nockPatchTraces(project, validator, reply, withError) {
```
matthewloring: Are all arguments after `project` optional here?

test/nocks.js
```diff
@@ -0,0 +1,60 @@
...
+    scope = scope.reply(reply || 200);
+  }
+  return scope;
+}
+
+module.exports = {
+  oauth2: nockOAuth2,
+  patchTraces: nockPatchTraces,
+  projectId: nockProjectId,
+};
```
matthewloring: '\n'

test/test-trace-writer.js
```diff
@@ -18,94 +18,133 @@
...
-      samplingRate: 0,
-      forceNewAgent_: true
-    });
-    common.avoidTraceWriterAuth(agent);
-    cls.getNamespace().run(function() {
-      queueSpans(2, agent);
-      buf = common.getTraces(agent);
-      setTimeout(function() {
-        scope.done();
+function createFakeSpan(name) {
```
matthewloring: It would be nice for something like this to live in `common`.

test/nocks.js
ofrobots: Being explicit is better though.

test/nocks.js
ofrobots: Yes.

test/test-trace-writer.js
ofrobots: Maybe. Depends on whether the other tests want the mock to be more functional. It is a tradeoff between what this set of unit tests needs vs reimplementing all the functionality of `SpanData`. `SpanData` has dependencies on `Agent` and `Logger` which need to be injected, perhaps it is better to inject these dependencies and use the real `SpanData` class?

src/trace-writer.js
```diff
@@ -244,16 +249,19 @@ TraceWriter.prototype.flushBuffer_ = function(projectId) {
...
 
-  this.request_({
+  var options = {
     method: 'PATCH',
     uri: uri,
     body: json,
     headers: headers
-  }, function(err, response, body) {
+  };
+  that.request(options, function(err, body, response) {
```
kjin: Should this be `function(err, response, body)`? Both `this.request_` and `that.request` in the old and new revisions respectively are both assigned with a call to `util.makeAuthenticatedRequestFactory` (the latter in [`service.js`](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/common/src/service.js)).

test/nocks.js
ofrobots: Done.

bin/run-test.sh
```diff
@@ -39,7 +39,7 @@ function run {
 }
 
 # Run test/coverage
-for test in test/*.js test/plugins/*.js ;
```
matthewloring: We could also move the non-test config file into `test/fixtures`.

test/test-trace-writer.js
matthewloring: This is something we can aim for long term. Nothing to change in this PR.

src/trace-writer.js
ofrobots: I have been scratching my head about how this used to work before, but this does need to be `(err, body, response)`.

kjin: 🙃

PR #418, opened 2017-02-24
Plugins with falsey paths don't get loaded

kjin:
This change seeks to introduce not loading a plugin if its path is falsey.

This would make it possible to stop built-in plugins from loading. While we have previously discussed that we don't necessarily know what use cases we are supporting with this, I can name one now: I want to be able to run per-plugin benchmarks, and it would be really neat to isolate a single plugin at a time.

I think falsey is the best approach here because we know for sure that it's not a valid path.

### DIFF
--- a/config.js
+++ b/config.js
@@ -33,12 +33,16 @@ module.exports = {
     // `enhancedDatabaseReporting` is enabled.
     databaseResultReportingSize: 127,
 
-    // An object describing which modules to trace. To enable tracing for a
-    // module, add its name as a key under this object, as well as the
-    // require-friendly module path of the plugin that implements tracing for
-    // that module as the corresponding value. Relative paths are not accepted.
-    // An empty object means that no modules will be automatically traced at
-    // all.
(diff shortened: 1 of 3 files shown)


matthewloring:
I'm fine with this as long as there is documentation and testing of this behavior.

kjin:
@matthewloring I've added both.

config.js
```diff
@@ -33,12 +33,16 @@ module.exports = {
...
+    // name of the module to trace, and its value is the require-friendly path
+    // to the plugin.
+    // By default, all of the following plugins are loaded.
+    // Specifying a different object for this field in the configuration passed
+    // to the method that starts the trace agent will cause that object to be
+    // merged with this one.
+    // To disable a plugin in this list, you may override its path with a falsey
+    // value. Disabling any of the default plugins may cause unwanted behavior,
+    // so use caution.
     // This field is experimental.
```
matthewloring: Let's remove this.

kjin: Done

PR #423, opened 2017-02-27
Remove version lower bound check

matthewloring:


matthewloring:
The engines field in `package.json` serves this purpose now.

PR #426, opened 2017-02-27
Update default logLevel to warn

matthewloring:


matthewloring:
We're going to leave this as is for now. If any users observe problems, they can increase the log level to get information related to missed spans.

PR #427, opened 2017-02-28
Trace API and Plugin Developer Docs

kjin:
Follow-up to #362.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,15 +126,44 @@ The trace agent can be configured by passing a configurations object to the agen
 
 One configuration option of note is `enhancedDatabaseReporting`. Setting this option to `true` will cause database operations for redis and MongoDB to record query summaries and results as labels on reported trace spans.
 
-## Disabling the trace agent
+### Disabling the trace agent
 
 The trace agent can be turned off by either setting the `GCLOUD_TRACE_DISABLE` environment variable or specifying `enabled: false` in your configuration file.
 
-## Trace batching and sampling
+### Trace batching and sampling
 
 The aggregation of trace spans before publishing can be configured using the `flushDelaySeconds` and `bufferSize` [options](config.js). The spans recorded for each incoming requests are placed in a buffer after the request has completed. Spans will be published to the UI in batch when the spans from `bufferSize` requests have been queued in the buffer or after `flushDelaySeconds` have passed since the last publish, whichever comes first.
 
 The trace configuration additionally exposes the `samplingRate` option which sets an upper bound on the number of traced requests captured per second. Some Google Cloud environments may override this sampling policy.
+
+### Additional Plugins
+
+TODO
+
+## Custom Tracing API
+
+The custom tracing API can be used to add custom spans to trace. A *span* is a particular unit of work within a trace, such as an RPC request. Spans may be nested; the outermost span is called a *root span*, even if there are no nested child spans. Root spans typically correspond to incoming requests, while *child spans* typically correspond to outgoing requests, or other work that is triggered in response to incoming requests.
+
+For any of the web frameworks for which we provide [built-in plugins](#what-gets-traced), a root span is automatically started whenever an incoming request is received (in other words, all middleware already runs within a root span). If you wish to record a span outside of any of these frameworks, any traced code must run within a root span that you create yourself.
+
+### Accessing the API
+
+Calling the `start` function returns an instance of `TraceApi`, which provides an interface for tracing:
+
+```javascript
+  var traceApi = require('@google/cloud-trace').start();
+```
+
+It can also be retrieved by subsequent calls to `get` elsewhere:
+
+```javascript
+  // after start() is called
+  var traceApi = require('@google/cloud-trace').get();
+```
+
+The object returned by both of these calls is guaranteed to have the surface described below, even if the agent is disabled.
+
+A fully detailed overview of the `TraceApi` object is available [here](./trace-api.md).
 
 ## Contributing changes
 
--- a/README.md
+++ b/README.md
@@ -136,9 +136,23 @@ The aggregation of trace spans before publishing can be configured using the `fl
 
 The trace configuration additionally exposes the `samplingRate` option which sets an upper bound on the number of traced requests captured per second. Some Google Cloud environments may override this sampling policy.
 
-### Additional Plugins
+### Tracing Additional Modules
 
-TODO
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -126,16 +126,45 @@ The trace agent can be configured by passing a configurations object to the agen
...
-## Trace batching and sampling
+### Trace batching and sampling
 
 The aggregation of trace spans before publishing can be configured using the `flushDelaySeconds` and `bufferSize` [options](config.js). The spans recorded for each incoming requests are placed in a buffer after the request has completed. Spans will be published to the UI in batch when the spans from `bufferSize` requests have been queued in the buffer or after `flushDelaySeconds` have passed since the last publish, whichever comes first.
 
 The trace configuration additionally exposes the `samplingRate` option which sets an upper bound on the number of traced requests captured per second. Some Google Cloud environments may override this sampling policy.
 
+### Additional Plugins
+
+TODO
```
matthewloring: Let's not land this with TODOs in the readme despite the LGTM

kjin: Sorry, was not done writing stuff. Updated

README.md
```diff
@@ -126,16 +126,59 @@ The trace agent can be configured by passing a configurations object to the agen
...
+```
+
+It can also be retrieved by subsequent calls to `get` elsewhere:
+
+```javascript
+  // after start() is called
+  var traceApi = require('@google/cloud-trace').get();
+```
+
+The object returned by both of these calls is guaranteed to have the surface described below, even if the agent is disabled.
```
matthewloring: "described below"

README.md
```diff
@@ -126,16 +126,59 @@ The trace agent can be configured by passing a configurations object to the agen
 
 One configuration option of note is `enhancedDatabaseReporting`. Setting this option to `true` will cause database operations for redis and MongoDB to record query summaries and results as labels on reported trace spans.
 
-## Disabling the trace agent
+### Disabling the trace agent
 
 The trace agent can be turned off by either setting the `GCLOUD_TRACE_DISABLE` environment variable or specifying `enabled: false` in your configuration file.
```
ofrobots: While you're here.. `GCLOUD_TRACE_DISABLE` is no more, can you remove it? Better.. get rid of this whole section.

README.md
```diff
@@ -126,16 +126,59 @@ The trace agent can be configured by passing a configurations object to the agen
...
+### Tracing Additional Modules
+
+In addition to the modules listed [above](#what-gets-traced), the trace agent can be configured to trace additional modules through the use of *plugins*. To load an additional plugin, specify it in the agent's configuration:
+
+```javascript
+  require('@google/cloud-trace').start({
+    plugins: {
+      // You may use a package name or absolute path to the file.
+      'my-module': '@google/cloud-trace-plugin-my-module',
+      'another-module': path.join(__dirname, 'path/to/another/module/plugin.js')
```
ofrobots: nit: the example suggest strong coupling between `another-module` and the where the plugin file comes from. That is, makes me think that `plugin.js` is part of `another-module` that I must dig out. I suggest:

`path.join(__dirname, 'path/to/my-custom-plugins/plugin-another-module.js')`

README.md
```diff
@@ -126,16 +126,59 @@ The trace agent can be configured by passing a configurations object to the agen
...
+    }
+  });
+```
+
+This list of plugins will be merged with the list of built-in plugins, which will be loaded by the plugin loader. Each plugin is only loaded when the module that it patches is loaded; in other words, there is no computational overhead for loading plugins for unused modules.
+
```
ofrobots: nit: `s/for loading/for listing/`

doc/plugin-guide.md
```diff
@@ -0,0 +1,54 @@
...
+* `patch`: A function describing how the module exports for a given file should be patched. It will be passed two arguments: the object exported from the file, and an instance of [`TraceApi`](./trace-api.md).
+* `intercept`: A function describing how the module exports for a file should be replaced with a new object. It accepts the same arguments as `patch`, but unlike `patch`, it should return the object that will be treated as the replacement value for `module.exports` (hence the name `intercept`).
+* `unpatch`: A function describing how the module exports for a given file should be unpatched. This should generally mirror the logic in `patch`; for example, if `patch` wraps a method, `unpatch` should unwrap it.
+
+If `patch` is supplied, then `unpatch` will be called if the agent must be disabled for any reason. This does not hold true for `intercept`: instead, the module exports for the original file will automatically be set to its original, unintercepted value.
```
ofrobots: Does this mean that `unpatch` also does `unintercept`?

doc/plugin-guide.md
```diff
@@ -0,0 +1,54 @@
...
+
+In addition, `patch` and `intercept` have overlapping functionality.
```
ofrobots: Do we throw/error when both are provided? Which one gets used if both are provided?

doc/trace-api.md
```diff
@@ -32,11 +9,11 @@ A `TraceApi` instance provides functions that facilitate the following:
 
 In addition to the above, `TraceApi` also provides a number of well-known label keys and constants through its `labels` and `constants` fields respectively.
 
-#### Trace Spans
+## Trace Spans
 
 These functions provide the capability to create trace spans, add labels to them, and close them.
 
-* `TraceApi#api.runInRootSpan(options, fn)`
+* `TraceApi#runInRootSpan(options, fn)`
```
DominicKramer: We should probably mention here that the `endSpan` method needs to be manually called.  Otherwise, the trace won't get published.

doc/plugin-guide.md
kjin: We don't have anything to check this. Right now, both `patch` and `intercept` are called, in that order -- which effectively means only `intercept` gets used.

I suppose it's a pretty serious error when both are provided -- so what do you think about throwing an error when this is this case?

doc/plugin-guide.md
kjin: `unintercept` actually happens naturally because the list of intercepts is emptied out upon `deactivate`. I'll make it more clear.

README.md
kjin: Done

doc/trace-api.md
kjin: Done

README.md
kjin: The latter half (`enabled: false`) is still applicable, do we want to leave that in?

PR #428, opened 2017-02-28
Plugin loader throws for plugins with glaring issues

kjin:
Throw for plugin patches
- with both `patch` and `intercept`
- with neither

Warn for plugin patches
- with `unpatch` and `intercept` (should this throw as well?)

Open to any thoughts regarding severity level.

### DIFF
--- a/src/trace-plugin-loader.js
+++ b/src/trace-plugin-loader.js
@@ -51,14 +51,17 @@ function checkLoadedModules() {
 
 function checkPatch(patch) {
   if (!patch.patch && !patch.intercept) {
-    throw new Error('Plugin patch for ' + patch + ' doesn\'t patch ' +
+    throw new Error('Plugin for ' + patch.file + ' doesn\'t patch ' +
       'anything.');
   } else if (patch.patch && patch.intercept) {
-    throw new Error('Plugin patch for ' + patch.file + ' has ' +
+    throw new Error('Plugin for ' + patch.file + ' has ' +
       'both intercept and patch functions.');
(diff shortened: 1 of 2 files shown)


src/trace-plugin-loader.js
```diff
@@ -49,6 +49,19 @@ function checkLoadedModules() {
   }
 }
 
+function checkPatch(patch) {
+  if (!patch.patch && !patch.intercept) {
+    throw new Error('Plugin patch for ' + patch + ' doesn\'t patch ' +
```
matthewloring: This message could be ambiguous. Maybe "Plugin for ' + patch + ' doesn\'t contain any patches."?

src/trace-plugin-loader.js
```diff
@@ -49,6 +49,19 @@ function checkLoadedModules() {
...
+      'anything.');
+  } else if (patch.patch && patch.intercept) {
+    throw new Error('Plugin patch for ' + patch.file + ' has ' +
+      'both intercept and patch functions.');
+  } else if (patch.unpatch && patch.intercept) {
```
matthewloring: I think having a warnings for patch without unpatch and unpatch without patch might be good. I think with these combinations a separate warning for unpatch with intercept is obsolete.

kjin: Added for `patch.patch && !patch.unpatch` (the latter case is already covered).

src/trace-plugin-loader.js
kjin: Done (also corrected `patch` to `patch.file`)

PR #431, opened 2017-03-01
Update module name in docs

matthewloring:


ofrobots:
Also src/trace-plugin-loader.js

matthewloring:
Done.

PR #435, opened 2017-03-08
Support tracing of code that uses generic-pool

DominicKramer:


test/plugins/test-trace-generic-pool.js
```diff
@@ -0,0 +1,84 @@
...
+
+    var pool = genericPool.createPool(factory, opts);
+
+    var promise;
+    agent.runInRootSpan({ name: rootSpanName }, function(span) {
+      promise = pool.acquire().then(function(fn) {
+        fn('SomeInput');
+        span.endSpan();
+      }).then(function() {
+        var spans = common.getTraces(agent)[0].spans;
```
matthewloring: Do we have context at this program point?

src/plugins/plugin-generic-pool.js
```diff
@@ -0,0 +1,42 @@
...
+ */
+'use strict';
+
+var shimmer = require('shimmer');
+
+function patchModuleRoot(genericPool, api) {
+  shimmer.wrap(genericPool.Pool.prototype, 'acquire', function(original) {
+    return function() {
+      var result = original.apply(this, arguments);
+      shimmer.wrap(result, 'then', function(original) {
```
matthewloring: Does the module guarantee a fresh promise is returned from every invocation of `acquire`? We want to make sure we don't end up wrapping multiple times.

config.js
```diff
@@ -54,7 +54,8 @@ module.exports = {
       'mysql': path.join(__dirname, 'src/plugins/plugin-mysql.js'),
       'pg': path.join(__dirname, 'src/plugins/plugin-pg.js'),
       'redis': path.join(__dirname, 'src/plugins/plugin-redis.js'),
-      'restify': path.join(__dirname, 'src/plugins/plugin-restify.js')
+      'restify': path.join(__dirname, 'src/plugins/plugin-restify.js'),
+      'generic-pool': path.join(__dirname, 'src/plugins/plugin-generic-pool.js')
```
matthewloring: This should go in alphabetical order.

package.json
```diff
@@ -28,6 +28,7 @@
     "glob": "^7.0.3",
     "istanbul": "^0.4.2",
     "jshint": "^2.9.1",
+    "lodash.isfunction": "^3.0.8",
```
matthewloring: We should use `is#function` which is already a dependency of this module.

test/plugins/fixtures/generic-pool-2/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "generic-pool-2",
```
matthewloring: Prefer to follow the naming style used by other fixtures: `"generic-pool2"`.

test/plugins/test-trace-generic-pool.js
```diff
@@ -0,0 +1,154 @@
...
+
+    var pool;
+    api.runInRootSpan({ name: ROOT_SPAN }, function(span) {
+      pool = new genericPool.Pool(config);
+      span.endSpan();
+    });
+
+    pool.acquire(function(err, fn) {
+      assert.ifError(err);
+      var childSpan = api.createChildSpan({ name: CHILD_SPAN_1 });
```
matthewloring: I don't think we want to be able to create a child of `ROOT_SPAN` here since we are calling `acquire` from outside the context in which `ROOT_SPAN` is active. The general correctness condition here is that if you call `pool.acquire` from inside a `runInRootSpan` callback, then the same context is available when `pool.acquire` is called and inside the passed to `pool.acquire`. This is analogous to what you did below for generic-pool@3.

test/plugins/test-trace-generic-pool.js
```diff
@@ -0,0 +1,154 @@
...
+        assert.ok(childSpan);
+        fn('SomeInput');
+        childSpan.endSpan();
+        span.endSpan();
+      }).then(function() {
+        // With the current implementation, context propogation is lost
+        // at this point and the commented out assert that verifies
+        // that the child span is not null will fail.
+        // It looks like a generalized Promise context propogation solution
+        // is needed to support this invocation of then().
```
matthewloring: Let's hold off on landing this part of the test until we understand how https://github.com/othiym23/async-listener/pull/105/files interacts with it.

DominicKramer:
PTAL

package.json
DominicKramer: I've removed `lodash.isfunction`.

src/plugins/plugin-generic-pool.js
```diff
@@ -0,0 +1,42 @@
...
+        return function(onFulfilled, onRejected) {
+          return original.apply(this,
```
DominicKramer: Done.

config.js
DominicKramer: Done.

test/plugins/fixtures/generic-pool-2/package.json
DominicKramer: Done

test/plugins/test-trace-generic-pool.js
DominicKramer: I've left the tests, but removed a lot of the plugin code since it is no longer needed.

src/plugins/plugin-generic-pool.js
```diff
@@ -0,0 +1,35 @@
...
+      return original.call(this, callback ? api.wrap(callback) : callback,
+                                 priority);
+    };
+  });
+}
+
+module.exports = [
+  {
+    file: '',
+    versions: '2.x.x',
```
matthewloring: Prefer '2'

src/plugins/plugin-generic-pool.js
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+'use strict';
+
+var shimmer = require('shimmer');
+
+function wrapModuleRoot2(Pool, api) {
+  shimmer.wrap(Pool.Pool.prototype, 'acquire', function(original) {
+    return function(callback, priority) {
+      return original.call(this, callback ? api.wrap(callback) : callback,
```
matthewloring: In what case is this callback not defined?

test/plugins/test-trace-generic-pool.js
```diff
@@ -0,0 +1,147 @@
...
+
+describe('generic-pool2', function() {
+  var ROOT_SPAN = 'root-span';
+  var CHILD_SPAN_1 = 'child-span-1';
+  var CHILD_SPAN_2 = 'child-span-2';
+
+  var api;
+  var genericPool;
+  before(function() {
+    api = require('../..').start({ samplingRate: 0, stackTraceLimit: 0 });
```
matthewloring: Why are stack traces disabled here?

test/plugins/test-trace-generic-pool.js
```diff
@@ -0,0 +1,147 @@
...
+    var pool = genericPool.createPool(factory, opts);
+
+    var promise;
+    agent.runInRootSpan({ name: ROOT_SPAN }, function(span) {
+      promise = pool.acquire().then(function(fn) {
+        var childSpan = agent.createChildSpan({ name: CHILD_SPAN_1 });
+        assert.ok(childSpan);
+        fn('SomeInput');
+        childSpan.endSpan();
+        span.endSpan();
```
matthewloring: This won't affect test behavior but for readability could this root span be closed after both of its children have been closed?

src/plugins/plugin-generic-pool.js
DominicKramer: The API for generic-pool@2 specifies that the priority is optional but the callback is not.  However, if the callback is falsey, I thought it would be best to pass it directly to the original function call so that generic-pool can, itself, handle the fact that it is falsey.

src/plugins/plugin-generic-pool.js
DominicKramer: Done

test/plugins/test-trace-generic-pool.js
DominicKramer: This was unintentional.  I have removed `stackTraceLimit: 0` from both the generic-pool 2 and 3 tests.

test/plugins/test-trace-generic-pool.js
DominicKramer: Fixed

DominicKramer:
PTAL

Issue #434, opened 2017-03-07
Can't get gRPC Context Client to Server to match


ethanrubio:
I'm trying to use this library to trace gRPC and Hapi requests throughout a small microservice environment.

There are three services (using a combo of gRPC and HapI) that I'm trying to log correctly in StackDriver Trace. I started all three services under the same project, but on StackDriver, I'm unable to link up the gRPC Client requests with the gRPC Server handling. However, a service with gRPC making HTTP requests to a Hapi server are correctly matched up on the web interface.

Any insights would be greatly appreciated! Thanks!

matthewloring:
We currently propagate distributed context using an http header meaning that distributed tracing will only work over http for now. We are designing an alternate mechanism for grpc (likely using http://www.grpc.io/docs/guides/concepts.html#metadata) but need to coordinate between different language trace agents (we want to ensure a request sent from a Node.js grpc client to a go grpc server will handle context correctly).

kjin:
I've opened #436 to address this by writing trace context to gRPC metadata. While this field isn't standardized for other Cloud Trace clients, it should provide distributed tracing for Node.js microservices.

kjin:
Hi @ethanrubio, we just release version 1.0.1 which includes the PR mentioned above. Please give it a try when you have the time and feel free to re-open this issue if there are any problems!

ethanrubio:
@kjin Thanks again for adding this! Is it possible to search for `x-cloud-trace-context` id in the web UI?

matthewloring:
The value of `x-cloud-trace-context` has the form <trace_id>/<span_id>;o=<options>. It is possible to look up a trace from a trace id by navigating to `https://console.cloud.google.com/traces/traces?project=<your_project_name>&tid=<your_trace_id>`.

PR #436, opened 2017-03-08
Distributed tracing support in gRPC

kjin:
Fix for #434. This changes the gRPC plugin to add the `'x-cloud-trace-context'` metadata field to outgoing gRPC requests, and read it from incoming gRPC requests.

### DIFF
--- a/src/plugins/plugin-grpc.js
+++ b/src/plugins/plugin-grpc.js
@@ -96,6 +96,15 @@ function patchClient(client, api) {
         var metadata = new Metadata();
         if (!method.requestStream) {
           // unary or server stream
+          if (args.length === 0) {
+            // No argument (for the gRPC call) was provided, so we will have to
+            // provide one, since metadata cannot be the first argument.
+            // The internal representation of argument defaults to undefined
+            // in its non-presence.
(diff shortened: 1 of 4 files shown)


src/plugins/plugin-grpc.js
```diff
@@ -442,6 +470,12 @@ module.exports = [
     unpatch: unpatchClient
   },
   {
+    file: 'src/node/src/metadata.js',
+    versions: SUPPORTED_VERSIONS,
+    patch: patchMetadata,
```
matthewloring: Could you also add a small comment here (similar to above) stating why a no-op unpatching is safe (nothing is modified when we grab the constructor)?

kjin:
@matthewloring Once tests pass could you take a look at 80888a6? It addresses the previously failing tests.

kjin:
CI failure is due to our MySQL tracing tests failing on Node v7.7.2. I can reproduce this locally with v7.7.2 but not v7.7.1, indicating that it arose as a result of the new version of Node being released. We should fix this in a separate PR.

matthewloring:
LGTM given failures unrelated.

PR #438, opened 2017-03-13
Ensure http response streams are paused

### LINKED ISSUE #437: Tracing destroys proxied payload
We've got an app that uses hapi with [h2o2](https://github.com/hapijs/h2o2) for proxying requests to a different service. When the trace agent is enabled, the responses seem to be empty. The response headers are included as expected, but there is no body.

I've made a reproducible test case here: https://github.com/rexxars/gtrace-proxy-issue

If I start the agent and disable *either* the `http` or the `hapi` instrumentation (or both), everything seems to work as expected.

matthewloring:
Responses returned by http#request are yielded in paused mode. By
attaching a 'data' listener to the request the http instrumentation was
switching the stream to flowing mode which could cause the request to
drain before the calling framework had a chance to attach their own
listeners.

Fixes #437

### DIFF
--- a/src/plugins/plugin-http.js
+++ b/src/plugins/plugin-http.js
@@ -76,8 +76,24 @@ function patchRequest (http, api) {
       var req = request.call(this, options, function(res) {
         api.wrapEmitter(res);
         var numBytes = 0;
-        res.on('data', function (chunk) {
-          numBytes += chunk.length;
+        var listenerAttached = false;
+        // Responses returned by http#request are yielded in paused mode. Attaching
+        // a 'data' listener to the request will switch the stream to flowing mode
+        // which could cause the request to drain before the calling framework has
+        // a chance to attach their own listeners. To avoid this, we attach our listener
+        // lazily.
+        // This approach to tracking data size will not observe data read by
+        // explicitly calling `read` on the request. We expect this to be very
+        // uncommon as it is not mentioned in any of the official documentation.
+        shimmer.wrap(res, 'on', function onWrap(on) {
+          return function on_trace(eventName, cb) {
+            if (eventName === 'data' && !listenerAttached) {
+              on.call(this, 'data', function(chunk) {
+                numBytes += chunk.length;
+              });
+            }
+            on.apply(this, arguments);
+          };
         });
         res.on('end', function () {
           requestLifecycleSpan
--- a/test/plugins/test-trace-http.js
+++ b/test/plugins/test-trace-http.js
@@ -17,6 +17,7 @@
 
 var common = require('./common.js');
 var constants = require('../../src/constants.js');
+var stream = require('stream');
 var TraceLabels = require('../../src/trace-labels.js');
 
 var agent = require('../..').start({ samplingRate: 0 });
@@ -60,19 +61,6 @@ describe('test-trace-http', function() {
     );
   });
 
-  it('should accurately measure get time, no callback', function(done) {
-    server.listen(common.serverPort, common.runInTransaction.bind(null, agent,
-      function(endTransaction) {
-        http.get({port: common.serverPort, headers: {}});
-        setTimeout(function() {
-          endTransaction();
-          common.assertDurationCorrect(agent, common.serverWait);
-          done();
-        }, common.serverWait * 1.5);
-      })
-    );
-  });
-
   it('should not trace api requests', function(done) {
     server.listen(common.serverPort, common.runInTransaction.bind(null, agent,
       function(endTransaction) {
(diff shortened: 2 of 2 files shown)


ofrobots:
As discussed, it would be good to explore alternative ways to gather the request size rather than depending on internal state machine in `ReadableStream` – or to get reasonable confidence that this indeed is the best way to do this.

matthewloring:
It looks like there are facilities for tracking the bytes read from a connection (https://nodejs.org/api/net.html#net_socket_bytesread) but connections can be reused and we don't observe all requests so we can't determine per-request size based on the bytes observed by the socket.

There is also the content-length header (https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html) but servers are not required to supply this to my knowledge. 

As far as managing the stream state machine, I tried using `https://nodejs.org/api/stream.html#stream_readable_pause` but this doesn't restore the stream to its uninitialized state (https://nodejs.org/api/stream.html#stream_three_states). As a result the stream can buffer internally where it otherwise could not and with sufficient buffering may un-pause.

I can still investigate patches to `on`.

matthewloring:
Here is a patch that will do the lazy initialization of our `data` listener. It's more wrapping/indirection but at least it doesn't dig into stream internals (even if they are documented officially). WDYT?

```
diff --git a/src/plugins/plugin-http.js b/src/plugins/plugin-http.js
index 355bcae..b8d5ea1 100644
--- a/src/plugins/plugin-http.js
+++ b/src/plugins/plugin-http.js
@@ -76,8 +76,22 @@ function patchRequest (http, api) {
       var req = request.call(this, options, function(res) {
         api.wrapEmitter(res);
         var numBytes = 0;
-        res.on('data', function (chunk) {
-          numBytes += chunk.length;
+        var listenerAttached = false;
+        // Responses returned by http#request are yielded in paused mode. By
+        // attaching a 'data' listener to the request the http instrumentation was
+        // switching the stream to flowing mode which could cause the request to
+        // drain before the calling framework had a chance to attach their own
+        // listeners. We now attach this listener lazily.
+        shimmer.wrap(res, 'on', function onWrap(on) {
+          return function on_trace() {
+            var eventName = arguments[0];
+            if (eventName === 'data' && !listenerAttached) {
+              on.call(this, 'data', function(chunk) {
+                numBytes += chunk.length;
+              });
+            }
+            on.apply(this, arguments);
+          };
         });
         res.on('end', function () {
           requestLifecycleSpan
```

matthewloring:
As per offline discussion, we will go with the lazy initialization of our `on('data')` listener.

src/plugins/plugin-http.js
```diff
@@ -76,8 +76,22 @@ function patchRequest (http, api) {
       var req = request.call(this, options, function(res) {
         api.wrapEmitter(res);
         var numBytes = 0;
-        res.on('data', function (chunk) {
-          numBytes += chunk.length;
+        var listenerAttached = false;
+        // Responses returned by http#request are yielded in paused mode. By
+        // attaching a 'data' listener to the request the http instrumentation was
+        // switching the stream to flowing mode which could cause the request to
```
ofrobots: nit: Phrase the comment in present tense. A future reader of the comment may not have to context about what used to happen. The comment should instead explain what is the intent of code as present.

src/plugins/plugin-http.js
```diff
@@ -76,8 +76,22 @@ function patchRequest (http, api) {
       var req = request.call(this, options, function(res) {
         api.wrapEmitter(res);
         var numBytes = 0;
-        res.on('data', function (chunk) {
-          numBytes += chunk.length;
+        var listenerAttached = false;
```
ofrobots: Would this mechanism still work if the use calls `res.read()` rather than `res.on('data')` to consume the data?

matthewloring: Nope, added a comment to that effect.

src/plugins/plugin-http.js
```diff
@@ -76,8 +76,22 @@ function patchRequest (http, api) {
...
+        // Responses returned by http#request are yielded in paused mode. By
+        // attaching a 'data' listener to the request the http instrumentation was
+        // switching the stream to flowing mode which could cause the request to
+        // drain before the calling framework had a chance to attach their own
+        // listeners. We now attach this listener lazily.
+        shimmer.wrap(res, 'on', function onWrap(on) {
+          return function on_trace() {
+            var eventName = arguments[0];
```
matthewloring: Done.

src/plugins/plugin-http.js
matthewloring: done.

ofrobots:
Lgtm

PR #439, opened 2017-03-14
Add Support for redis 2.7

kjin:
The only change needed for this is expanding the semver condition range for the 2.6 patch to include 2.7 as well.

### DIFF
--- a/test/plugins/fixtures/redis2.x/package.json
+++ b/test/plugins/fixtures/redis2.x/package.json
@@ -3,6 +3,6 @@
   "version": "1.0.0",
   "main": "index.js",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


test/plugins/fixtures/redis2.x/package.json
```diff
@@ -3,6 +3,6 @@
   "version": "1.0.0",
   "main": "index.js",
   "dependencies": {
-    "redis": "^2.2.5"
+    "redis": "^2.6.x"
```
matthewloring: Could we change this to `"^2.x"` for readability?

kjin: Done

PR #442, opened 2017-03-19
Koa 2.x support

### LINKED ISSUE #433: Add instrumentation for Koa2

tinovyatkin:
Fixes #433

### DIFF
--- a/src/plugins/plugin-koa.js
+++ b/src/plugins/plugin-koa.js
@@ -18,8 +18,6 @@
 const shimmer = require('shimmer');
 var urlParse = require('url').parse;
 
-const SUPPORTED_VERSIONS = '1.x';
-
 function createUseWrap(api) {
   return function useWrap(use) {
     return function useTrace() {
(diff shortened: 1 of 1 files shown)


matthewloring:
Thank you for putting together this PR! A few high level thoughts:

1. The 2x middleware and the original middleware are structurally similar (with some differences in how middleware arguments are unpacked). Could the code be refactored to reduce duplication?

1. The 2x middleware currently uses some ES6 features that do not work on Node 4 (like spreads). While koa 2 does not support Node 4, it will require broader changes to our test harness to allow these features in the instrumentation. Are they necessary or could they be replaced with Node 4 compatable alternatives?

1. This patching will need testing akin to what we have for koa 1. We are happy to help write the tests but if there are any specific tests you'd like to include that would be a good starting point.

ageblade:
@tinovyatkin Is there any progress on this one ? I will be happy to help somehow

ofrobots:
Superceded by #464.

Issue #446, opened 2017-03-22
ignoreUrls issue


ethanrubio:
I've added various strings and regex to the `config.ignoreUrls`, but it seems to not be actually be ignoring the appropriate urls, because the url passed into FilterPolicy.prototype.matches is an empty string.

I've traced this issue all the way to `runInRootSpan` where the options passed in happens to be an example like this one `{ name: '/metrics', traceContext: undefined, skipFrames: 3 }`. And then when `createRootSpan_` is called `options.url` is always `undefined`, so it'll pass an empty string. 

Have I misconfigured something in my configuration where the url passed in becomes name, so it isn't recognized?

Thanks!

Update:
I've done more investigating, and I see that the hapi middleware that is created only passes this to `runInRootSpan`:

```js
    var options = {
      name: urlParse(req.url).pathname,
      traceContext: req.headers[api.constants.TRACE_CONTEXT_HEADER_NAME],
      skipFrames: 3
    };
```
with no url field, which causes the ignoreUrls to not work with hapi. Is this intentional?

oshalygin:
Here is my configuration and it's working like charm

```javascript

require('@google-cloud/trace-agent').start({
  ignoreUrls: [
    '/_ah/health',
    '/healthz'
  ]
});
```

kjin:
This is definitely a bug, thanks for bringing our attention to it.

Just noticed that you opened a PR for this. I'll take a look!

kjin:
Hi Ethan, we've released this under Node 1.0.3. Please give it a try, and re-open this if there are any outstanding issues!

PR #447, opened 2017-03-22
Add url field to hapi middleware options

ethanrubio:
Currently, the config ignoreUrls won't work when working with a Hapi server, since the url field does not exist when `runInRootSpan` is called.

See #446


ethanrubio:
I signed it!

kjin:
Thanks for contributing this! Usually we would like a unit test to go a long with it, but since we have to fix this for other modules, we'll add a follow-up to add unit tests for each of the web framework modules.

This looks good to me, so I'll merge it as soon as the CIs pass.

kjin:
v0.12 failures unrelated. Merging.

PR #449, opened 2017-03-22
Track download counts with non-namespaced dependency

kjin:
See also https://github.com/kjin/google-cloud-counters

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
   },
   "dependencies": {
     "@google-cloud/common": "^0.12.0",
+    "cloud-trace-counter": "^3.0.0",
     "continuation-local-storage": "^3.1.4",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -38,6 +38,7 @@
   },
   "dependencies": {
     "@google-cloud/common": "^0.12.0",
+    "cloud-trace-counter": "^3.0.0",
```
ofrobots: would be better if the dependency name could be a function of the tracked package name. I.e. possible to write a function that would convert `@google-cloud/trace-agent` into the name of the package counter module. Something like counter_google-cloud_trace-agent?

kjin: Ok, I can see why that would be nice. Yes, I'll do that

kjin: Done

PR #450, opened 2017-03-23
Stop running google-gax unit test on v0.12

kjin:
The `google-gax` plugin unit test doesn't work on Node v0.12 + gRPC v1.2, so this change stops it from running. I've also added a disclaimer in the README stating that in general, we won't officially provide support for modules that depend on gRPC (or gRPC itself) if the application is running on a version of Node that that version of gRPC does not support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ The agent will also automatic trace of the following kinds of RPCs:
 * [Redis](https://www.npmjs.com/package/redis) (versions 0.12 - 2)
 * [MySQL](https://www.npmjs.com/package/mysql) (version ^2.9)
 
-*Note: The latest versions of gRPC (versions 1.1 and up) have dropped support for Node.js <4.0. We do not officially support tracing gRPC on unsupported versions of Node.js.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -103,19 +103,21 @@ This is the trace list that shows a sampling of the incoming requests your appli
...
 The agent will also automatic trace of the following kinds of RPCs:
 * Outbound HTTP requests through the `http` and `https` core modules
-* [gRPC](https://www.npmjs.com/package/grpc) client (version 1)
+* [gRPC](https://www.npmjs.com/package/grpc)* client (version 1)
 * [MongoDB-core](https://www.npmjs.com/package/mongodb-core) (version 1)
 * [Mongoose](https://www.npmjs.com/package/mongoose) (version 4)
 * [Redis](https://www.npmjs.com/package/redis) (versions 0.12 - 2)
 * [MySQL](https://www.npmjs.com/package/mysql) (version ^2.9)
 
+*Note: The latest versions of gRPC (versions 1.1 and up) have dropped support for Node.js <4.0. We do not officially support tracing gRPC on unsupported versions of Node.js.
```
matthewloring: "We do not support tracing gRPC..."

PR #457, opened 2017-03-31
Add serviceContext option

kjin:
This change adds a serviceContext field in the Trace Agent's configuration, in the same manner of the Debug Agent.

### DIFF
--- a/config.js
+++ b/config.js
@@ -126,8 +126,10 @@ module.exports = {
     // Used by unit tests to force loading of a new agent if one exists already.
     forceNewAgent_: false,
 
-    // Specifies the current service context. These fields may be overridden by
-    // environmental variables set from within GAE or GCE.
+    // Specifies the service context with which traces from this application
+    // will be associated. This may be useful in filtering traces originating
+    // from a specific service within a project. These fields will automatically
+    // be set through environment variables on Google App Engine.
     serviceContext: {
       service: null,
       version: null,
--- a/config.js
+++ b/config.js
@@ -124,5 +124,13 @@ module.exports = {
 
     // For testing purposes only.
     // Used by unit tests to force loading of a new agent if one exists already.
-    forceNewAgent_: false
+    forceNewAgent_: false,
+
+    // Specifies the current service context. These fields may be overridden by
+    // environmental variables set from within GAE or GCE.
+    serviceContext: {
+      service: null,
+      version: null,
+      minorVersion: null
+    }
 };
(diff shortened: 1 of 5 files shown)


oshalygin:
OH god yes!  Thanks!!

index.js
```diff
@@ -45,7 +45,12 @@ var onUncaughtExceptionValues = ['ignore', 'flush', 'flushAndExit'];
 var initConfig = function(projectConfig) {
   var envConfig = {
     logLevel: process.env.GCLOUD_TRACE_LOGLEVEL,
-    projectId: process.env.GCLOUD_PROJECT
+    projectId: process.env.GCLOUD_PROJECT,
+    serviceContext: {
+      service: process.env.GAE_SERVICE || process.env.GAE_MODULE_NAME,
+      version: process.env.GAE_VERSION || process.env.GAE_MODULE_VERSION,
+      minorVersion: process.env.GAE_MINOR_VERSION
```
matthewloring: @ofrobots Is minorVersion needed for the trace agent? What does it represent in the context of kubernetes?

kjin: I think it's not necessary. I wanted to get thoughts from @ofrobots on this since he wrote a note about it on the debug agent.

config.js
```diff
@@ -124,5 +124,13 @@ module.exports = {
 
     // For testing purposes only.
     // Used by unit tests to force loading of a new agent if one exists already.
-    forceNewAgent_: false
+    forceNewAgent_: false,
+
+    // Specifies the current service context. These fields may be overridden by
+    // environmental variables set from within GAE or GCE.
```
ofrobots: nit: better comment:

```js
// Identifies the context of the running service - [ServiceContext](https://cloud.google.com/error-reporting/reference/rest/v1beta1/ServiceContext
// This information is utilized by the UI to identify the 'service' to attribute your traces to. This is
// discovered automatically when your application is running on AppEngine. In other environments
// you may want to provide a service and version manually.
```

index.js
ofrobots: Given that trace-writer.js uses the minor version already, we should have a way to accept it as part of the service context.

AdriVanHoudt:
If I set this inside of a GCE I will need to filter with the urls specified in the linked issue and I cannot filter using the dropdowns?
Would you advise to set these vars in the hopes the dropdowns will allow me to do this in the future?

matthewloring:
Currently the dropdowns cannot be made to work with GCE but there is a solution in the works. Once it is available, there will be documentation available on how to use it. Setting these vars in the mean time could help your filtering but will not help with the dropdowns.

AdriVanHoudt:
@matthewloring thanks for clarifying, I'm looking forward to the dropdowns :P

PR #459, opened 2017-04-06
Allow Manual Test Fixture Management

DominicKramer:
This change adds the `bin/install-test-fixtures.sh` and `clean-test-fixtures.sh` files.

In particular, `bin/run-test.sh` calls `bin/install-test-fixtures.sh` which only does an `npm install` on fixture directories that do not have a `node_modules` directory.

This should not alter how tests are run in the CI, but should make local testing faster.


DominicKramer:
PTAL

PR #463, opened 2017-04-10
Set response trace context based on that of incoming requests

kjin:
Previously, plugins that patched web frameworks would add in response headers the trace context corresponding to the locally-created root span. This should actually be set to the trace context of the incoming request (modified based on whether this request was traced).


src/plugins/plugin-express.js
```diff
@@ -40,15 +40,18 @@ function patchModuleRoot(express, api) {
       skipFrames: 3
     };
     api.runInRootSpan(options, function(rootSpan) {
+      // Set response trace context.
+      var outgoingTraceContext =
```
ofrobots: `responseTraceContext` here and elsewhere.

src/plugins/plugin-hapi.js
```diff
@@ -54,14 +61,11 @@ function createMiddleware(api) {
     
       // we use the path part of the url as the span name and add the full
       // url as a label
-      // req.path would be more desirable but is not set at the time our middlewear runs.
+      // req.path would be more desirable but is not set at the time our middleware runs.
```
ofrobots: lol

src/trace-api.js
```diff
@@ -297,4 +338,4 @@ function createRootSpan_(api, options, skipFrames) {
     incomingTraceContext.spanId,
     skipFrames + 1);
   return new RootSpan(api.agent_, rootContext);
-}
+}
```
ofrobots: Add back the newline.

src/util.js
```diff
@@ -70,10 +70,12 @@ function parseContextFromHeader(str) {
       (matches[2] && isNaN(matches[2]))) {
     return null;
   }
+  // We assume that if no value is set for options, we treat the incoming
+  // context as if it had options set to 1.
```
ofrobots: Doesn't match my [expectations](https://docs.google.com/document/d/1A4yPsG__X1T4JbZZrT0GcOkCkqRIi48-eY9ZZJUdF-Y/edit).

kjin:
@ofrobots I haven't done any test refactoring in this PR, what do you mean by that the "two commits aren't individually landable"?

kjin:
Oh I see - I was planning to squash these commits.

kjin:
@GoogleCloudPlatform/node-team CI passed, PTAL

src/trace-api.js
```diff
@@ -179,6 +191,31 @@ TraceApiImplementation.prototype.createChildSpan = function(options) {
...
+ * header, the string to be set as this header's value. Otherwise, an empty
+ * string.
+ */
+TraceApiImplementation.prototype.getResponseTraceContext = function(
+    incomingTraceContext, isTraced) {
+  var traceContext = this.agent_.parseContextFromHeader(incomingTraceContext);
+  if (!traceContext) {
+    return '';
+  }
+  traceContext.options = traceContext.options && isTraced ? 1 : 0;
```
matthewloring: Not sure which is more readable but `traceContext.options & isTraced` should also work here.

src/trace-api.js
```diff
@@ -221,6 +258,7 @@ var phantomApiImpl = {
   enhancedDatabaseReportingEnabled: function() { return false; },
   runInRootSpan: function(opts, fn) { return fn(null); },
   createChildSpan: function(opts) { return null; },
+  getResponseTraceContext: function(context, traced) { return null; },
```
matthewloring: Should this return `''` to keep things stringy?

src/util.js
```diff
@@ -73,7 +73,7 @@ function parseContextFromHeader(str) {
   return {
     traceId: matches[1],
     spanId: matches[2],
-    options: Number(matches[3])
+    options: typeof(matches[3]) === 'undefined' ? 0 : Number(matches[3])
```
matthewloring: This is a really sneaky bug!! Could you add a unit test for this as well?

matthewloring: Also, maybe `Number(matches[3] || 0)` is more readable? In either case, could you add a comment as well?

kjin: Done

PR #464, opened 2017-04-11
Koa 2.x support

matthewloring:


ofrobots:
Note parallel dependency on #463. Whichever lands last needs to some more work?

src/plugins/plugin-koa.js
```diff
@@ -18,83 +18,111 @@
...
+    return next();
   };
 }
 
 module.exports = [
   {
     file: '',
-    versions: SUPPORTED_VERSIONS,
+    versions: '1.x',
+    patch: function(koa, api) {
```
ofrobots: nit: Dedupe this function.

matthewloring: done.

matthewloring:
CLA should be fine given that the included commit was already opened in https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/442 which received CLA approval.

PR #465, opened 2017-04-11
Increase timeout to reduce flakiness

matthewloring:


test/test-agent-stopped.js
```diff
@@ -43,7 +43,7 @@ describe('test-agent-stopped', function() {
       assert.ok(!agent.isActive());
       scope.done();
       done();
-    }, 20);
+    }, 200);
```
kjin: Nit: There's a comment above this that says `Wait 20ms` and it should say `200ms` instead.

PR #466, opened 2017-04-11
Add tests that context is propagated by child frameworks

matthewloring:


test/plugins/test-trace-grpc.js
```diff
@@ -315,6 +315,28 @@ Object.keys(versions).forEach(function(version) {
       });
     });
 
+    it('should propagate context', function(done) {
+      var start = Date.now();
+      common.runInTransaction(agent, function(endTransaction) {
+        callUnary(client, grpc, {}, function() {
+          endTransaction();
```
kjin: I may be misunderstanding this, but it doesn't seem like we're checking `hasContext` in this test.

matthewloring: Yikes, good catch!

PR #468, opened 2017-04-14
Support Knex

DominicKramer:


DominicKramer:
PTAL

test/plugins/test-trace-knex.js
```diff
@@ -0,0 +1,209 @@
...
+  var assert;
+  before(function() {
+    agent = require('../..').start({
+      logLevel: 2,
+      flushDelaySeconds: 1,
+      enhancedDatabaseReporting: true,
+      databaseResultReportingSize: RESULT_SIZE
+    });
+    assert = require('assert');
+    knex = require('./fixtures/knex0.12')({
```
matthewloring: Is there any way knex could be added as a "version" of mysql so that we can reuse some of the existing mysql test logic instead of copying it?

src/plugins/plugin-bluebird.js
```diff
@@ -0,0 +1,41 @@
...
+
+var shimmer = require('shimmer');
+var is = require('is');
+
+function wrapFunction(api, fn) {
+  return fn && is.fn(fn) ? api.wrap(fn) : fn;
+}
+
+function patchPromise(Promise, api) {
+  shimmer.wrap(Promise.prototype, '_addCallbacks', function(original) {
```
matthewloring: If this is all that is required for knex we don't necessarily need to add more to this PR but if we end up patching bluebird long term we may want to patch it more exhaustively. It would be good to give bluebird the same context propagation semantics as [async-listener gives to the global promise](https://github.com/othiym23/async-listener/blob/master/index.js#L337).

src/plugins/plugin-knex.js
```diff
@@ -0,0 +1,65 @@
...
+var shimmer = require('shimmer');
+var is = require('is');
+
+var VERSIONS = '0.12.x';
+
+function patchClient(Client, api) {
+  shimmer.wrap(Client.prototype, 'runner', function(original) {
+    return function() {
+      var runner = original.apply(this, arguments);
+      runner.query = api.wrap(runner.query);
```
matthewloring: Is `runner` invoked once per outgoing database call?

src/plugins/plugin-knex.js
```diff
@@ -0,0 +1,65 @@
...
+      return runner;
+    };
+  });
+}
+
+function interceptKnex(Knex, api) {
+  return function() {
+    var result = Knex.apply(this, arguments);
+    var proto = Object.getPrototypeOf(result.client);
```
matthewloring: Is this different from `result.client.prototype`?

src/plugins/plugin-knex.js
```diff
@@ -0,0 +1,65 @@
...
+    shimmer.wrap(proto, 'transaction', function(original) {
+      return function() {
+        var args = Array.prototype.slice.call(arguments).map(function(item) {
```
matthewloring: Is this only to wrap a callback in the final argument position or are there multiple possible function arguments?

src/plugins/plugin-knex.js
```diff
@@ -0,0 +1,62 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+var shimmer = require('shimmer');
+var util = require('util');
+var is = require('is');
+
+var VERSIONS = '0.12.x';
```
matthewloring: It looks like the current version of `knex` is 0.13.0. It would be good to extend this patching as far forward and back in version support as it can go without logic modification.

src/plugins/plugin-knex.js
```diff
@@ -0,0 +1,62 @@
...
+  });
+}
+
+function wrapIfFn(fn, api) {
+  return fn && is.fn(fn) ? api.wrap(fn) : fn;
+}
+
+function interceptTransaction(Transaction, api) {
+  function WrappedTransaction(client, container, config, outerTx) {
+    Transaction.call(this, client, wrapIfFn(container, api), config, outerTx);
```
matthewloring: Under what conditions would this not be a function? It is generally good to fully bail out or warn if the world isn't in the form we expect it to be.

test/plugins/test-trace-knex.js
DominicKramer: There is significant differences between the knex and mysql tests that prevents this.

src/plugins/plugin-bluebird.js
DominicKramer: I've updated the PR to patch knex itself without patching bluebird.

src/plugins/plugin-knex.js
DominicKramer: I looks like that is the case.

src/plugins/plugin-knex.js
DominicKramer: I've updated the code to patch a different method and thus this code is no longer used.

src/plugins/plugin-knex.js
DominicKramer: I've updated the code to patch a different method and thus this code is no longer used.

src/plugins/plugin-knex.js
DominicKramer: I've updated the tests and plugin to for knex `>=0.10 <=0.13.0`.

src/plugins/plugin-knex.js
DominicKramer: I've updated the code to address this issue.  In particular, the `container` argument should always be a function.

src/plugins/plugin-knex.js
matthewloring: For future readers, it sounds as though there could be many invocations of `runner` per transaction but `runner`s are not reused between transactions which could have cause context to leak.

test/plugins/test-trace-knex.js
```diff
@@ -0,0 +1,267 @@
...
+  var agent;
+  var assert;
+  before(function() {
+    agent = require('../..').start({
+      logLevel: 2,
+      flushDelaySeconds: 1,
+      enhancedDatabaseReporting: true,
+      databaseResultReportingSize: RESULT_SIZE
+    });
+    assert = require('assert');
```
matthewloring: Can this be moved to the top level. Assert does not hold any state that could interfere with the trace agent's patching.

DominicKramer: Fixed.

PR #471, opened 2017-04-26
Add `runPromiseInSpan` to the trace API

DominicKramer:
The `runPromiseInSpan` method runs a given `Promise` in a span
and ensures the span is ended either when the promise is resolved
or rejected.

### DIFF
--- a/src/trace-api.js
+++ b/src/trace-api.js
@@ -190,6 +190,21 @@ TraceApiImplementation.prototype.createChildSpan = function(options) {
   }
 };
 
+TraceApiImplementation.prototype.runPromiseInSpan = function(options, promise) {
+  var span = this.createChildSpan(options);
+  if (!span) {
+    return promise;
+  }
+
+  return promise.then(function(result) {
+    span.endSpan();
+    return result;
+  }).catch(function(err) {
+    span.endSpan();
+    throw err;
+  });
+};
+
 /**
  * Generates a stringified trace context that should be set as the trace context
  * header in a response to an incoming web request. This value is based on
@@ -258,6 +273,7 @@ var phantomApiImpl = {
   enhancedDatabaseReportingEnabled: function() { return false; },
   runInRootSpan: function(opts, fn) { return fn(null); },
   createChildSpan: function(opts) { return null; },
+  runPromiseInSpan: function(opts, promise) {},
   getResponseTraceContext: function(context, traced) { return ''; },
   wrap: function(fn) { return fn; },
   wrapEmitter: function(ee) {},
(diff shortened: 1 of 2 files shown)


src/trace-api.js
```diff
@@ -258,6 +273,7 @@ var phantomApiImpl = {
   enhancedDatabaseReportingEnabled: function() { return false; },
   runInRootSpan: function(opts, fn) { return fn(null); },
   createChildSpan: function(opts) { return null; },
+  runPromiseInSpan: function(opts, promise) {},
```
matthewloring: This should probably have an explicit return just for readability.

test/test-trace-api.js
```diff
@@ -123,6 +123,88 @@ describe('Trace Interface', function() {
...
+    it('should propagate rejections from a promise within a root span', function() {
+      return traceAPI.runInRootSpan({name: 'root'}, function(root) {
+        var e = new Error('Custom Message');
+        return traceAPI.runPromiseInSpan({name: 'sub'}, Promise.reject(e))
+          .then(function() {
+            assert.fail('then', 'catch', 'Should enter catch() instead of then()');
+          })
+          .catch(function(err) {
+            assert.ok(err);
+            assert.equal(err, e);
```
matthewloring: It would be good to run this test asynchronously and finish in here. We want to make sure this block gets run.

src/trace-api.js
```diff
@@ -190,6 +190,21 @@ TraceApiImplementation.prototype.createChildSpan = function(options) {
   }
 };
 
+TraceApiImplementation.prototype.runPromiseInSpan = function(options, promise) {
```
kjin: Would it be worthwhile/feasible to have an equivalent function for root spans?

ofrobots:
I am somewhat confused by the naming here. I am not sure if the concept of 'running' a promise is well founded. For example, if the promise is already settled, then you are not running anything. Even in the case of unsettled promises, the executor is already synchronously executed before users can call `runInPromiseInSpan`. The child span wouldn't include this time making this less useful.

matthewloring:
What if the API accepted an executor and started timing before constructing the promise. The function could then return the newly constructed promise capturing both sync + async time.

ofrobots:
Secondly, AFAIK there is no callback counterpart of this function at this point.. i.e. `runInSpan` doesn't exist. We only have `createChilSpan`. It seems odd to be adding a utility function but only for the promise case.

matthewloring:
Yup, that part makes perfect sense. I just wanted to mention how sync + async tracking could be done. It's unclear the value of syntactic sugar outweighs the confusion of understanding exactly which promises in a long chain would be traced. I'm fine abandoning this as well.

matthewloring:
As discussed, we are not planning to make this API change.

PR #472, opened 2017-05-01
drop dependency on dummy counter module

ofrobots:
Now that npm has download stats for scoped modules, we no longer need this.


ofrobots:
Build failures are unrelated. See #473.

PR #478, opened 2017-05-08
Update deps

matthewloring:
This change updates @google-cloud/common to a version which is no longer
compatable with node v0.12. To upgrade this dependency, we must drop
support for 0.12 as well. Additionally, we will stop running some of the
non-interference tests which only ran on 0.12. These tests will be
upgraded to work with newer versions of node in follow on work.

### DIFF
--- a/bin/run-test.sh
+++ b/bin/run-test.sh
@@ -50,5 +50,3 @@ fi
 
 # Run non-interference tests
 node test/non-interference/http-e2e.js || exit 1
-node test/non-interference/express-e2e.js || exit 1
-node test/non-interference/restify-e2e.js || exit 1
--- a/test/fixtures/preloaded-agent.js
+++ b/test/fixtures/preloaded-agent.js
@@ -21,3 +21,6 @@ var agent = require('../../');
 assert(agent.get().isActive());
 
 console.log('Preload test passed.');
+// Exit explicitly since the process is being held open by retries to the
+// metadata service.
+process.exit(0);
--- a/test/fixtures/start-agent.js
+++ b/test/fixtures/start-agent.js
@@ -18,3 +18,6 @@
 process.env.GCLOUD_TRACE_LOGLEVEL = 4;
 require('glob'); // Load a module before agent
 require('../..').start();
+// Exit explicitly since the process is being held open by retries to the
+// metadata service.
+process.exit(0);
--- a/test/non-interference/http-e2e.js
+++ b/test/non-interference/http-e2e.js
@@ -49,7 +49,16 @@ var test_glob = semver.satisfies(process.version, '0.12.x') ?
 
 // Run tests
 console.log('Running tests');
-var gcloud_require = 'require(\'' + path.join(__dirname, '..', '..') +
+var gcloud_require =
+    'var proxyquire = require(\'' +
+    path.join(__dirname, '../../node_modules/proxyquire') +
+    '\');' +
+    'proxyquire(\'' +
+    path.join(__dirname, '../../node_modules/gcp-metadata') +
+    '\', { \'retry-request\': require(\'' +
+    path.join(__dirname, '../../node_modules/request') +
+    '\')});' +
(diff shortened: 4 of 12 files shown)


matthewloring:
Blocked by: https://github.com/stephenplusplus/retry-request/issues/9

matthewloring:
Ready for a look

matthewloring:
Tests are passing, this is ready for a look.

test/non-interference/http-e2e.js
```diff
@@ -49,7 +49,16 @@ var test_glob = semver.satisfies(process.version, '0.12.x') ?
 
 // Run tests
 console.log('Running tests');
-var gcloud_require = 'require(\'' + path.join(__dirname, '..', '..') +
+var gcloud_require =
+    'var proxyquire = require(\'' +
+    path.join(__dirname, '..', '..', 'node_modules', 'proxyquire') +
```
ofrobots: I am pretty sure`path.join(__dirname, '../../node_modules/proxyquire')` would work portably.

bin/run-test.sh
```diff
@@ -50,5 +50,3 @@ fi
 
 # Run non-interference tests
 node test/non-interference/http-e2e.js || exit 1
-node test/non-interference/express-e2e.js || exit 1
```
ofrobots: What's the rationale for dropping these?

test/non-interference/http-e2e.js
matthewloring: Is that because we don't run this test on windows? I would guess that `require` does not handle mixed delimiter path names.

test/fixtures/preloaded-agent.js
```diff
@@ -21,3 +21,4 @@ var agent = require('../../');
 assert(agent.get().isActive());
 
 console.log('Preload test passed.');
+process.exit(0);
```
matthewloring: Done.

test/fixtures/start-agent.js
```diff
@@ -18,3 +18,4 @@
 process.env.GCLOUD_TRACE_LOGLEVEL = 4;
 require('glob'); // Load a module before agent
 require('../..').start();
+process.exit(0);
```
matthewloring: Done.

bin/run-test.sh
matthewloring: These never ran on anything other than 0.12. I thought it would be more misleading to leave them if they can never run. At some point we should write updated versions of these.

test/non-interference/http-e2e.js
ofrobots: path.join would normalize the path automatically.

bin/run-test.sh
ofrobots: How do we make sure these get added back in the future?

More generally, I would like to see rationale like this documented in the issue comment and also in the commit message. When we look back at the history we would be wondering what happened and would have no documentation of the rationale.

test/non-interference/http-e2e.js
matthewloring: I believe it just connects the components with the correct platform dependent separator.
```
> path.join('c:\foo\bar', 'a/b/c');
'c:\foo\bar/a/b/c'
```

bin/run-test.sh
matthewloring: We currently have a bug internally tracking the upgrade of these tests. We could also hold off on dropping support/upgrading the dependencies until these tests can be upgraded. I can update the commit message.

test/non-interference/http-e2e.js
ofrobots: Did you run the above no windows?

ofrobots: Are you saying that require wants unix paths? In that case why are we using join at all?

matthewloring: Sorry, I just didn't fully understand how `path.join` works. Fixed.

Issue #477, opened 2017-05-05
Missing trace calls where the socket closed before a response was made


kdeclerck:
We are using Hapi and we have some calls which are missing in traces.
This seems to be for requests where the socket is closed before we send a response.
Is it possible that traces aborts or something on a socket close event?

matthewloring:
This should be fixed by #479 which will be available after the next release.

PR #479, opened 2017-05-09
hapi/koa: End span when request is aborted

kjin:
This is an attempt to address #477.

Currently checking if the same problem is present in other web frameworks (it's not in express as far as I can tell).

### DIFF
--- a/test/plugins/test-trace-connect.js
+++ b/test/plugins/test-trace-connect.js
@@ -190,6 +190,35 @@ describe('test-trace-connect', function() {
       });
     });
   });
+
+  it('should end spans even if client aborts', function(done) {
+    var app = connect();
+    app.use('/', function (req, res) {
+      setTimeout(function() {
+        res.end(common.serverRes);
+        setImmediate(function() {
+          var traces = common.getTraces(agent);
+          assert.strictEqual(traces.length, 1);
+          assert.strictEqual(traces[0].spans.length, 1);
+          var span = traces[0].spans[0];
+          common.assertSpanDurationCorrect(span, common.serverWait);
+          done();
+        });
+      }, common.serverWait);
+    });
+    server = app.listen(common.serverPort, function() {
+      var req = http.get({port: common.serverPort, path: '/'},
+        function(res) {
+          assert.fail();
+        });
+      // Need error handler to catch socket hangup error
+      req.on('error', function() {});
+      // Give enough time for server to receive request
+      setTimeout(function() {
+        req.abort();
+      }, common.serverWait / 2);
+    });
+  });
 });
 
 function connectPredicate(span) {
--- a/test/plugins/test-trace-hapi.js
+++ b/test/plugins/test-trace-hapi.js
@@ -43,7 +43,8 @@ describe('hapi', function() {
   before(function() {
     agent = require('../..').start({
       ignoreUrls: ['/ignore'],
-      samplingRate: 0
+      samplingRate: 0,
+      logLevel: 4
     });
   });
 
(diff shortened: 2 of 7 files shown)


AdriVanHoudt:
On the same team as @kdeclerck and I think this happening due to a request taking too long to respond aka call reply after >2min where then node would close the socket causing a response close event (If i remember correctly)

kjin:
@AdriVanHoudt Ah, I see -- good to know. I will look into it further.

kjin:
@AdriVanHoudt It seems like the `'abort'` event is still emitted as a result of what you just described (server doesn't respond within 2 min), so this covers that case... unless you are talking about something else, in which case, would you be able to provide an example?

AdriVanHoudt:
If that is the case then 👌

kjin:
Updated to also include Koa. Our trace agent seems to end spans for other frameworks.

test/plugins/test-trace-hapi.js
```diff
@@ -308,6 +308,41 @@ describe('hapi', function() {
           });
         });
       });
+
+      it('should end spans when client aborts request', function(done) {
```
matthewloring: Would it be possible to include comparable tests for the other frameworks even if no logical changes are required in the hooks? How does an aborted request surface for those frameworks?

kjin: The middleware runs to completion as far as I can tell. Yes, I can (and should) add those tests.

test/plugins/test-trace-connect.js
```diff
@@ -190,6 +190,35 @@ describe('test-trace-connect', function() {
...
+    var app = connect();
+    app.use('/', function (req, res) {
+      setTimeout(function() {
+        res.end(common.serverRes);
+        setImmediate(function() {
+          var traces = common.getTraces(agent);
+          assert.strictEqual(traces.length, 1);
+          assert.strictEqual(traces[0].spans.length, 1);
+          var span = traces[0].spans[0];
+          common.assertSpanDurationCorrect(span, common.serverWait);
```
matthewloring: Wouldn't we expect the span to have duration `common.serverWait / 2` since that is the time between start and abort?

kjin: For connect/express/restify I think the code in which spans are ended (`res.end` or equivalent) are run regardless of whether the request is aborted, suggesting that the server does the full amount of work without realizing the request was aborted.

Actually, reading more on this, I realize I should try this with multiple middleware.

test/plugins/test-trace-hapi.js
ofrobots: This is not adding up for me. In the hapi plugin we are adding events on the raw request object (aka. http.IncomingMessage). If this request object is not getting the `end` event fired for hapi, then I am not sure how restify or express could work.

One possible scenario is that for restify and express the 'req' object passed to our middleware isn't the actual raw request object but some other that is guaranteed to receive the `end` event. This doesn't seem plausible to me.

kjin: So I've found that what makes hapi and koa different from express + others is when `res.end` is called. From what I've seen digging into each of these codebases, hapi and koa call `res.end` after all middleware is done, but the function from which this is called short-circuits if the socket is already closed. On the other hand, express calls `res.end` in a call to `res.send`, a method that applications are meant to call within their middleware. This call to `res.end`, as far as I can tell, is guaranteed.

test/plugins/test-trace-connect.js
kjin: Things still work for multiple middleware. Explanation in other comment chain.

matthewloring: Could you add a comment describing this behavior?

kjin: Done.

test/plugins/test-trace-hapi.js
AdriVanHoudt: First off,  💪 on the work, really appreciate it!
Second should `res.end` not always be called or am I getting this wrong? (https://nodejs.org/api/http.html#http_response_end_data_encoding_callback)

kjin: I think if you're calling `res.end` on your own, you shouldn't change whether you call it or not based on client behavior.

hapi (I'm looking [here](https://github.com/hapijs/hapi/blob/09c2e3374cf0ad63144f90268fb0b0d0e8a2ceba/lib/transmit.js#L249)) seems to not call it if the request was aborted, suggesting that it might be a no-op.

PR #480, opened 2017-05-09
patching: fix module-internal requires not being intercepted

kjin:
Intercepts are only being done for top-level modules -- in other words, files being required internally within modules won't be intercepted.

In this fix, we cache modules that were intercepted in a separate object so that when loaded, we'll return the replaced value instead.

### DIFF
--- a/src/trace-plugin-loader.js
+++ b/src/trace-plugin-loader.js
@@ -164,7 +164,7 @@ function activate(agent) {
           return patchedRoot;
         }
       } else {
-        var modulePath = Module._resolveFilename(request, parent);
+        var modulePath = Module._resolveFilename(request, parent).replace('/', path.sep);
         if (intercepts[modulePath]) {
           return intercepts[modulePath].interceptedValue;
         }
--- a/src/trace-plugin-loader.js
+++ b/src/trace-plugin-loader.js
@@ -151,8 +151,8 @@ function activate(agent) {
     // Future requires get patched as they get loaded.
     return function Module_load(request, parent, isMain) {
       var instrumentation = plugins[request];
-      var moduleRoot = util.findModulePath(request, parent);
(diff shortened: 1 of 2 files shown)


src/trace-plugin-loader.js
```diff
@@ -121,15 +122,18 @@ function activate(agent) {
...
-          var loadPath = moduleRoot ? path.join(moduleRoot, patch.file) : patch.file;
           patch.module = originalModuleLoad(loadPath, module, false);
         }
         if (patch.patch) {
           patch.patch(patch.module, instrumentation.api);
         }
         if (patch.intercept) {
           patch.module = patch.intercept(patch.module, instrumentation.api);
+          intercepts[loadPath] = {
+            interceptedValue: patch.module
```
matthewloring: Is there a reason not to store this value directly on `intercepts[loadPath]`?

kjin: If the intercepted value is falsey then it becomes more complicated to check whether the value was replaced or not. Ex. returning `undefined` from the intercept function (though, not sure why one would do that).

src/trace-plugin-loader.js
```diff
@@ -159,6 +163,11 @@ function activate(agent) {
         if (patchedRoot !== null) {
           return patchedRoot;
         }
+      } else {
+        var modulePath = Module._resolveFilename(request, parent).replace('/', path.sep);
```
matthewloring: Why is this logic not needed in the patch case?

kjin: The issue is with the persistence of the patch/intercept. Patches persist because the object that's being mutated is being cached, so the original module loader should be able to retrieve it. On the other hand, the only indication that a module was intercepted is within the scope of this file.

Issue #481, opened 2017-05-12
(http-hook) bug on wrapping event-emitter on http response


vmarchaud:
When wrapping the event emitter of the http response, someone forgot to return `on.apply(this, arguments);` to allow chaining, little example : 

```javascript
const http = require('http');
const express = require('express')
const app = express()

const options = {
  hostname: 'encrypted.google.com',
  path: '/',
  method: 'GET'
};

app.get('/', function (req, res) {
  var _req = http.request(options, (_res) => {
    let responseData = '';
    _res.on('data', function (data) {
      responseData += data;
    }).on('end', () => res.send(responseData));
  });
  _req.on('error', console.error);
  _req.end();
});

app.listen(3000, function () {
  console.log('Example app listening on port 3000!');
});
```

As you can see, i'm chaining multiple`.on` the `_res`. If you run this with the latest version and enabling the tracing, you'll have an error : 
```javascript
TypeError: Cannot read property 'on' of undefined
      at http.request (/tmp/yolo.js:17:7)
      at ClientRequest.<anonymous> (/home/vmarchaud/xxxxx/yyyyy/src/plugins/plugin-http.js:105:18)
```

To fix it, we simply need to add a `return` [here](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/src/plugins/plugin-http.js#L95) : `return on.apply(this, arguments);`

I'm making a PR right now to fix it :)

matthewloring:
Thank you for raising this and submitting the fix. Fix is now landed.

PR #482, opened 2017-05-12
http: return response object for chaining #481

vmarchaud:
Fix for https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/issues/481


matthewloring:
Thank you for putting this together!

PR #485, opened 2017-05-16
Add test notifications to travis

matthewloring:


ofrobots:
Actually, please double check that external users are able to send emails to this address.

PR #486, opened 2017-05-16
trace policy: only call match if candidate isn't string

kjin:
This change makes it so that `String.prototype.match` is only called in `FilterPolicy` for `RegExp` objects. It shouldn't be called for strings -- consider if we wanted to ignore `'/'` but not anything else.


ofrobots:
Hi @kjin. What's the status of this? If this isn't ready to land, let's close this pull-request.

PR #487, opened 2017-05-17
Add a system-test

ofrobots:
The encrypted key files for travis aren't setup yet (hence this will fail on travis), but opening early so that I can get a review on the JS bits.


system-test/trace-express.js
```diff
@@ -0,0 +1,129 @@
...
+const assert = require('assert');
+const googleAuth = require('google-auto-auth');
+const got = require('got');
+const queryString = require('querystring');
+const uuid = require('uuid');
+
+const WRITE_CONSISTENCY_DELAY_MS = 20 * 1000;
+const projectId = process.env.GCLOUD_PROJECT;
+
+function makeAuthorizedGot(auth) {
```
matthewloring: I agree that this promise code is messy. Could you instead do:

```javascript
auth.authorizeRequest(options, (err, opts) => {
  if (err) ...
  got(url, opts).then(...);
});

File: system-test/trace-express.js [hunk 2]
```diff
@@ -0,0 +1,129 @@
...
+
+  return (url, options) => {
+    return authorize(options)
+      .then((authorizedOptions) => {
+        return got(url, authorizedOptions);
+      });
+  };
+}
+
+function listTraces(testPath) {
```
matthewloring: Does it make sense to pull this into a separate API client as is done with the debugger?

system-test/trace-express.js
```diff
@@ -0,0 +1,129 @@
...
+  const query = queryString.stringify({
+    filter: `span:${testPath}`
+  });
+  const uri = `${BASE_URI}/traces?${query}`;
+
+  return agot(uri, { json: true }).then((response) => {
+    const body = response.body;
+    const promises = body.traces.map((trace) => {
+      const uri = `${BASE_URI}/traces/${trace.traceId}`;
+      return agot(uri, {
```
matthewloring: What information is available in this response that is not available from the original call? I thought they both returned full "trace" objects.

system-test/trace-express.js
```diff
@@ -0,0 +1,129 @@
...
+        listTraces(testPath).then(verifyTraces);
+      }, WRITE_CONSISTENCY_DELAY_MS);
+    });
+
+    function verifyTraces(traces) {
+      assert.equal(traces.length, 1, 'there should be exactly one trace');
+
+      const trace = traces[0];
+      console.log(trace);
+      assert.equal(trace.spans.length, 2, 'should be 2 spans: parent, child');
```
matthewloring: There is a chance that there will be a third span for auth.

package.json
```diff
@@ -22,10 +23,14 @@
     "node": ">=0.12"
   },
   "devDependencies": {
+    "@google-cloud/datastore": "^1.0.2",
```
ofrobots: What was the rationale for making them only show up in subdirectories? Going forward, I would like this dependency to be kept up to date so that we are integrating with the latest datastore library. Making it a fixtures makes it easy to miss the update.

In fact, I think we should change the semver range here to "*".

system-test/trace-express.js
ofrobots: I am not sure I follow how would that work? The authorized got function needs to return a promise. Are you suggesting getting rid of the got factory? That would incur multiple nesting levels in the code where I need to use `got`.

The code is dense, but it is readable if you understand promises.

Eventually this code can go into a separate module if we need to use this in other places.

system-test/trace-express.js
ofrobots: Yep, but that is outside the scope of this PR.

An API client would be nice, but it require quite a lot more work. BTW, there is also this: https://github.com/google/google-api-nodejs-client/blob/master/apis/tracing/v2.ts.

system-test/trace-express.js
ofrobots: Nope they don't.

system-test/trace-express.js
ofrobots: Hmm.. that's problematic. How much of a chance?

system-test/trace-express.js
matthewloring: Discussed offline, this is more general than necessary here but will hopefully be more broadly used in the future. Maybe we could add a TODO to move this to a more widely accessible utility?

system-test/trace-express.js
matthewloring: Agreed.

system-test/trace-express.js
matthewloring: This contradicts the public docs: https://cloud.google.com/trace/docs/reference/v1/rest/v1/projects.traces#TraceSpan. We should follow up with the trace service team.

system-test/trace-express.js
matthewloring: Should be fully deterministic. Not sure why you wouldn't but this was more just a heads up in case you observed more than 2 spans.

PR #489, opened 2017-05-18
Add a system-test

ofrobots:
Second attempt at #487, but travis on-merge tests should pass now.

### DIFF
--- a/bin/run-test.sh
+++ b/bin/run-test.sh
@@ -50,3 +50,15 @@ fi
 
 # Run non-interference tests
 node test/non-interference/http-e2e.js || exit 1
+
+
+# If on travis, and this is not a pull request build, run the system-tests
+# as well.
+if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then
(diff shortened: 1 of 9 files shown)


bin/run-test.sh
```diff
@@ -50,3 +50,15 @@ fi
 
 # Run non-interference tests
 node test/non-interference/http-e2e.js || exit 1
+
+
+# If on travis, and this is not a pull request build, run the system-tests
+# as well.
+if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then
```
matthewloring: This will probably fail locally as the travis env variable isn't set but the key + iv variables aren't either.

ofrobots: Yep, but we also don't want it running on appveyor and circle. I guess I have to go find the laundry list of environment variables to add here. :/

ofrobots: Will do this in a follow-on.

matthewloring: SGTM

PR #494, opened 2017-05-23
system-tests to use custom env. vars.

ofrobots:
This way we can avoid globally setting GCLOUD_PROJECT and
GOOGLE_APPLICATION_CREDENTIALS in travis.yml. Having those env. vars.
set can cause the code to use different paths.

### DIFF
--- a/system-test/trace-express.js
+++ b/system-test/trace-express.js
@@ -16,24 +16,29 @@
 
 'use strict';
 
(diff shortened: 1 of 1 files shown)


ofrobots:
Yep, I have already added the new env. vars. Once this lands, we can remove the old vars.

PR #497, opened 2017-05-27
fix credentials on system-test

ofrobots:
We no longer have GOOGLE_APPLICATION_CREDENTIALS defined on travis and this test was sensitive to picking up the credentials from there.


ofrobots:
TBR=@matthewloring

PR #499, opened 2017-05-30
start testing against Node 8

ofrobots:
Node 8 is being tested now. Because of changes in Node core, we were no longer able to trace http.get calls. This PR includes a fix for that too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 npm-debug.log
 coverage
 .DS_Store
+package-lock.json
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: node_js
 node_js:
   - "4"
   - "6"
-  - "7"
+  - "8"
 
 services:
   - mongodb
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,9 +2,9 @@
 environment:
   matrix:
   # node.js
-  # Testing against LTS/Current, 0.12 intentionally omitted to reduce maintenance burden
   - nodejs_version: "4"
   - nodejs_version: "6"
+  - nodejs_version: "8"
 
 services:
   - mongodb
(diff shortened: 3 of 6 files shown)


matthewloring:
Will circle pick it up automatically or do we need to do something there too?

ofrobots:
Circle tests a single version. I think it should continue to test 6.

ofrobots:
This need another look. Fixed the test failures due to our inability to trace `http.get` on Node 8.

.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules
 npm-debug.log
 coverage
 .DS_Store
+package-lock.json
```
AdriVanHoudt: why not commit this?

ofrobots: Haven't wrapped my head around whether we want package-lock.json or yarn.lock to be checked into in the repo. I need to understand the difference in the tree and then figure out which is a better solution for the development team here.

AdriVanHoudt: This might be a relevant twitter thread https://twitter.com/maybekatz/status/869819920611016705

matthewloring: Npm definitely wants this file to be checked in:
```
> npm install
npm notice created a lockfile as package-lock.json. You should commit this file.
```
Would it be a big deal if the package-lock.json and yarn.lock were not precisely the same (given they were semver compatible)?

matthewloring:
Maybe we could land just the travis portion of this + fixes for node 8? I don't think we need to be blocked by the grpc compilation errors on windows.

.gitignore
ofrobots: Some more context:  https://yarnpkg.com/blog/2017/05/31/determinism/

I think it should be okay to check both in. We can do this in a follow-on.

matthewloring:
I was hoping not to check this in while it was breaking appveyor. Do we have an eta on when the grpc fix will be ready?

ofrobots:
Oh I misunderstood. I think the gRPC compile fix on Windows will be available 'shortly'.

PR #501, opened 2017-06-07
docs: Elaboration on GKE Scopes + Syntax Highlighting

kjin:
Added an example CLI for running under GKE.


kjin:
Appveyor failure unrelated (docs only change)

PR #504, opened 2017-06-08
Roll back to old version of got

matthewloring:
New versions of got do not play well with async listener. The bug
tracking this issue is here:
https://github.com/othiym23/async-listener/issues/112.

Got is only a dev dependency used by the system tests so relying on an
older version should be fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "express": "^4.15.2",
     "glob": "^7.0.3",
     "google-auto-auth": "^0.7.0",
-    "got": "^7.0.0",
+    "got": "^5.7.1",
     "istanbul": "^0.4.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -30,7 +30,7 @@
     "express": "^4.15.2",
     "glob": "^7.0.3",
     "google-auto-auth": "^0.7.0",
-    "got": "^7.0.0",
+    "got": "5.7.1",
```
ofrobots: Any reason for locking down this rigidly? Wouldn't 5.x work better?

matthewloring: Done.

matthewloring:
Windows x node 8 failure unrelated.

PR #505, opened 2017-06-09
Stop running google-cloud test on windows

matthewloring:
There is no precompiled binary available for grpc 1.3.x on windows with
node version 8.

### DIFF
--- a/test/test-trace-gcloud.js
+++ b/test/test-trace-gcloud.js
@@ -19,6 +19,13 @@ var common = require('./plugins/common.js');
 var nock = require('nock');
 var assert = require('assert');
 var path = require('path');
+var semver = require('semver');
+
+if (process.platform === 'win32' && semver.satisfies(process.version, '8.x')) {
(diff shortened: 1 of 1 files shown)


ofrobots:
Can you clarify if this is a temporary solution or a permanent one?

/cc @mlumish

murgatroid99:
The windows precompiled binaries for Node 8 are available for grpc 1.3.8.

matthewloring:
Oh, based on the error message, it looks like it is trying to install `grpc@1.3.7`. It looks like grpc isn't depended on directly (https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/google-cloud/package.json#L108). I'll try to figure out why it isn't picking up 1.3.8.

matthewloring:
More complete solution here: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/509. Also, it looks like the weird version was coming from a stale npm cache.

PR #506, opened 2017-06-09
fix patching of res.end

ofrobots:
Use apply rather than named parameters – we were missing the callback argument.


ofrobots:
windows x node 8 failures are not related.

PR #510, opened 2017-06-13
Support for gRPC 1.4.0

matthewloring:
Steps taken to verify:

1) test/plugins/test-trace-grpc.js -> success
2) npm install@1.4.0-pre1 in test/plugins/fixtures/grpc1
3) test/plugins/test-trace-grpc.js -> failure
4) update version range in src/plugins/plugin-grpc.js to 0.13 -
1.4.0-pre1
5) test/plugins/test-trace-grpc.js -> success
6) update version range in src/plugins/plugin-grpc.js to 0.13 - 1.4.0

### DIFF
--- a/src/plugins/plugin-grpc.js
+++ b/src/plugins/plugin-grpc.js
@@ -469,7 +469,7 @@ function unpatchServer(server) {
 
 // # Exports
 
-var SUPPORTED_VERSIONS = '0.13 - 1';
+var SUPPORTED_VERSIONS = '0.13 - 1.4.0';
 
 module.exports = [
   {


src/plugins/plugin-grpc.js
```diff
@@ -469,7 +469,7 @@ function unpatchServer(server) {
 
 // # Exports
 
-var SUPPORTED_VERSIONS = '0.13 - 1';
+var SUPPORTED_VERSIONS = '0.13 - 1.4.0';
```
ofrobots: You're restricting the semver range here.. was that your intention? You expect that we won't be able ot support 1.4.1?

PR #511, opened 2017-06-14
Deduplicate internal code

kjin:
* `ChildSpan`/`RootSpan` are a thin layer on top of `SpanData`, so they will be removed and `SpanData` will be changed to match the interface of `RootSpan`.
* `TraceWriter` becomes a singleton (#517), so it’s accessible anywhere (reference doesn’t have to be passed from `TraceAgent` to new `SpanData` instances) and initialized in the public `start()` function.
* `TraceAgent` becomes a per-plugin object, can be disabled, and gets the interface that `TraceAPI` had before. One instance (for custom spans) is exposed publicly, and can be retrieved with the public `get()` function (as before).
* GCP Metadata queries are done all in one place.
* Unit tests will be updated to reflect above changes (partially done in #511).


test/test-unpatch.js
```diff
@@ -16,22 +16,37 @@
...
 var trace = require('..');
 
-var common = require('./plugins/common.js');
-
 describe('index.js', function() {
   var agent;
+  var scope;
   var checkUnpatches = [];
+  before(function() {
+    delete process.env.GCLOUD_PROJECT;
```
ofrobots: Please restore the environment variable after the tests.

test/test-unpatch.js
```diff
@@ -16,22 +16,37 @@
...
+    nock.disableNetConnect();
```
ofrobots: Can this be done at the top-level? That has a better chance of catching any early automatic requests as well.

test/test-trace-writer.js
```diff
@@ -64,26 +64,28 @@ describe('TraceWriter', function() {
...
+  //         serviceContext: {},
+  //         onUncaughtException: 'ignore',
+  //         forceNewAgent_: true
+  //       });
+  //       setTimeout(function() {
+  //         assert.ok(scope.isDone());
+  //         done();
+  //       }, DEFAULT_DELAY);
+  //     });
+  // });
```
ofrobots: Commented out code.

index.js
```diff
@@ -77,91 +173,57 @@ var agent;
...
-    logger.error('The value of onUncaughtException should be one of ',
-      onUncaughtExceptionValues);
-    throw new Error('Invalid value for onUncaughtException configuration.');
-  }
-
-  var headers = {};
-  headers[constants.TRACE_AGENT_REQUEST_HEADER] = 1;
+  // CLS namespace for context propagation
+  cls.createNamespace();
+  TraceWriter.create(logger, config);
```
ofrobots: Could we get the metadata and `then` create the TraceWriter, passing in all necessary information into the constructor and getting rid of `setMetadata`? Either way, it seems like TraceWriter is a singleton, and arbitrary calls to `TraceWriter.get()` can return either an initialized or uninitialized (`setMetadata` hasn't been called yet) TraceWriter. It would be better to return null, or a fully initialized writer.

I think it would be better for `TraceWriter.get

index.js
```diff
@@ -53,14 +62,101 @@ var initConfig = function(projectConfig) {
...
 
-var traceApi = new TraceApi('Custom Span API');
-var agent;
+/**
+ * Retrieves various fields from the GCP metadata service, and calls
+ * setMetadata on the TraceWriter singleton with these values.
+ * @param {common.logger} logger A logger object.
+ * @param {Boolean} getProjectId Whether the project ID should be retrieved
+ * from the metadata service.
+ */
```
ofrobots: add `return` annotation.

index.js
```diff
@@ -53,14 +62,101 @@ var initConfig = function(projectConfig) {
...
+  var headers = {};
+  headers[constants.TRACE_AGENT_REQUEST_HEADER] = 1;
+
+  var metadata = {};
+  var getHostnameAndInstanceId = function() {
+    gcpMetadata.instance({
+      property: 'hostname',
+      headers: headers
+    }, function(err, response, hostname) {
+      if (err && err.code !== 'ENOTFOUND') {
```
ofrobots: You didn't modify this, but I find it disturbing that we lack symmetry in how we deal with `err` between here and how get the projectId.

Given that these pieces of code are together now, it would be good to make error handling consistent. If you don't want to make these changes as part of this PR, please open an issue for this.

src/metadata.js
```diff
@@ -0,0 +1,55 @@
+var gcpMetadata = require('gcp-metadata');
```
ofrobots: missing copyright, use strict.

src/metadata.js
```diff
@@ -0,0 +1,55 @@
...
+      var property = fields[i].property;
+      var onError = fields[i].onError;
+
+      var prevNext = next;
+      next = function() {
+        gcpMetadata[endpoint]({
+          property: property,
+          headers: headers
+        }, function(err, response, body) {
+          if (response && response.statusCode !== 200) {
```
ofrobots: wouldn't it make sense to check for err first? Otherwise we end up clobbering the error message.

src/metadata.js
```diff
@@ -0,0 +1,55 @@
...
+            }
+          }
+          if (!result[endpoint]) {
+            result[endpoint] = {};
+          }
+          result[endpoint][property] = body;
+          prevNext();
+        })
+      }
+    })()
```
ofrobots: missing semicolons

src/metadata.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+  next()
```
ofrobots: missing semicolon

src/metadata.js
```diff
@@ -0,0 +1,55 @@
...
+}
```
ofrobots: missing semi-colon and newline.

src/metadata.js
```diff
@@ -0,0 +1,55 @@
+var gcpMetadata = require('gcp-metadata');
+var constants = require('./constants.js');
+
+// prevent self tracing
+var headers = {};
+headers[constants.TRACE_AGENT_REQUEST_HEADER] = 1;
+
+module.exports = function getMetadata(logger, fields, cb) {
```
ofrobots: I really don't like the string based enum ('continue', 'break') that adds strong coupling between tracewriter and this file.

A few suggestions:
1. Make this file return a promise for a single metadata property.
2. Keep the then functions in trace writer.
3. Use try/catch.

src/metadata.js
kjin: I believe the function follows the Node callback convention (only `err` or `response` is non-null).

src/span-data.js
```diff
@@ -35,15 +35,19 @@ var uid = 1;
...
   var spanId = uid++;
-  this.agent = agent;
   var spanName = traceUtil.truncate(name, constants.TRACE_SERVICE_SPAN_NAME_LIMIT);
   this.span = new TraceSpan(spanName, spanId, parentSpanId);
   this.trace = trace;
   this.isRoot = isRoot;
+  this.serializedTraceContext = traceUtil.generateTraceContext({
+    traceId: this.trace.traceId,
+    spanId: this.span.spanId,
+    options: 1 // always traced; TODO
```
kjin: Removed TODO -- it was left there earlier to remind me to verify that options should always be set to 1.

src/metadata.js
kjin: Done

src/trace-writer.js
```diff
@@ -62,12 +64,51 @@ function TraceWriter(logger, options) {
...
+util.inherits(TraceWriter, common.Service);
+
+TraceWriter.prototype.stop = function() {
+  this.isActive = false;
+};
+
+TraceWriter.prototype.initialize = function(cb) {
+  var that = this;
+  // Ensure that cb is called only once.
+  var pendingOperations = 2;
```
ofrobots: prefer to use https://www.npmjs.com/package/once

index.js
```diff
@@ -77,91 +103,59 @@ var agent;
...
-        }
-        return;
-      }
-      config.projectId = projectId;
-    });
-  } else if (typeof config.projectId !== 'string') {
+  traceAgent = new TraceAgent('Custom Span API', logger, config);
+  pluginLoader.activate(logger, config);
+
+  // Get metadata.
```
ofrobots: this comment is not relevant anymore?

index.js
```diff
@@ -77,91 +103,59 @@ var agent;
...
-  headers[constants.TRACE_AGENT_REQUEST_HEADER] = 1;
-
   if (modulesLoadedBeforeTrace.length > 0) {
     logger.error('Tracing might not work as the following modules ' +
       'were loaded before the trace agent was initialized: ' +
       JSON.stringify(modulesLoadedBeforeTrace));
   }
+  // CLS namespace for context propagation
+  cls.createNamespace();
+  TraceWriter.create(logger, config);
```
ofrobots: Can `create` accept a callback, and call `initialize` internally?

index.js
```diff
@@ -77,91 +103,59 @@ var agent;
...
-        return;
-      }
-      config.projectId = projectId;
-    });
-  } else if (typeof config.projectId !== 'string') {
+  traceAgent = new TraceAgent('Custom Span API', logger, config);
+  pluginLoader.activate(logger, config);
+
+  // Get metadata.
+  if (typeof config.projectId !== 'string' && typeof config.projectId !== 'undefined') {
```
ofrobots: this check can be done immediately after `initConfig` as it is checking the type of a property on a read only object. You won't even need to call `stop` anymore.

src/trace-writer.js
kjin: `once` doesn't cover the exact desired behavior here, as we want the callback to be fired when the latter operation completes, not the former.

src/plugins/plugin-http.js
```diff
@@ -66,8 +66,9 @@ function patchHttp(http, api) {
   }
 
   function isTraceAgentRequest (options) {
-    return options && options.headers &&
+    var result = options && options.headers &&
```
kjin: Oops, debugging change, thanks for catching

index.js
kjin: Done

index.js
kjin: Removed

index.js
kjin: Discussed in person, moving it might result in a semver-minor change, so would do in follow-up PR (this one is meant to be patch-only).

src/span-data.js
```diff
@@ -35,15 +35,19 @@ var uid = 1;
...
   this.trace = trace;
   this.isRoot = isRoot;
+  this.serializedTraceContext = traceUtil.generateTraceContext({
+    traceId: this.trace.traceId,
+    spanId: this.span.spanId,
+    options: 1 // always traced
+  });
   trace.spans.push(this.span);
-  if (agent.config().stackTraceLimit > 0) {
+  if (TraceWriter.get().config_.stackTraceLimit > 0) {
```
kjin: Done

src/trace-writer.js
```diff
@@ -16,42 +16,46 @@
...
-  common.Service.call(this, serviceOptions, options);
+  common.Service.call(this, serviceOptions, config);
 
   /** @private */
   this.logger_ = logger;
 
   /** @private */
-  this.config_ = options;
+  this.config_ = config;
+  this.projectId_ = config.projectId;
```
kjin: Done, reverted back

src/metadata.js
```diff
@@ -0,0 +1,49 @@
...
+        }
+      }
+      if (err) {
+        reject(err);
+      } else {
+        resolve(body);
+      }
+    });
+  });
+};
```
kjin: Entire file was removed, so no newline added.

PR #514, opened 2017-06-14
Fix document source link

oshalygin:
- The link in the TraceApi documentation was referencing
  a local directory path but instead needed to be out
  one directory.
- The line number referencing runInRootSpan is now on L42
  ipo L35.  That said, in the future it may make sense to
  remove the line number because keeping all of this in sync
  is borderline maintenance pain.


kjin:
@oshalygin Thanks for the fix! You're right, in the future we might fix to a specific commit.

PR #517, opened 2017-06-15
make TraceWriter a singleton

kjin:
This change makes TraceWriter a singleton.

### DIFF
--- a/test/plugins/common.js
+++ b/test/plugins/common.js
@@ -25,6 +25,7 @@ proxyquire('gcp-metadata', {
 require('../../src/trace-writer').publish_ = function() {};
 
 var cls = require('../../src/cls.js');
+var TraceWriter = require('../../src/trace-writer.js');
 var pluginLoader = require('../../src/trace-plugin-loader.js');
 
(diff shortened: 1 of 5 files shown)


test/plugins/common.js
```diff
@@ -74,8 +75,10 @@ function cleanTraces(agent) {
   }
 
   var privateAgent = agent.private_();
-  privateAgent.traceWriter.buffer_ = [];
-  privateAgent._shouldTraceArgs = [];
+  if (privateAgent) {
+    privateAgent._shouldTraceArgs = [];
+  }
+  TraceWriter.get().buffer_ = [];
```
ofrobots: A type checker would tell you that TraceWriter.get() can potentially return undefined. Now you haven't changed this code, but I think your implicit assumption is that TraceWrite.create() must have already been called?

If that's the case, shouldn't there be an assertion in `get()` to make sure it doesn't return undefined?

kjin: Yes - it should be an error to call `get` before `create`. I'll make an update.

PR #518, opened 2017-06-15
test: omit agent argument from all test-common helper functions

kjin:
Dependent on #517

Most changes are mechanical, the large change in particular is in [`test-trace-api.js`](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/518/files#diff-6672fc4a2f77ed575dedbaced33e52cf). Because the `agent` argument is removed, and `test-trace-api.js` is the only file that effectively used this argument, some functionality that was previously in `common.js` had to be duplicated here.

### DIFF
--- a/test/test-trace-api.js
+++ b/test/test-trace-api.js
@@ -17,16 +17,23 @@
 'use strict';
 
 var assert = require('assert');
+var cls = require('../src/cls.js');
 var common = require('./plugins/common.js');
 var EventEmitter = require('events');
 var extend = require('extend');
+var request = require('request');
 var TraceAPI = require('../src/trace-api.js');
 var TracingPolicy = require('../src/tracing-policy.js');
+var TraceWriter = require('../src/trace-writer.js');
 
 var config = extend({}, require('../config.js'),
   { samplingRate: 0, projectId: '0' });
 var logger = require('@google-cloud/common').logger();
 var agent = require('../src/trace-agent.js').get(config, logger);
+
+function replaceTracingPolicy(traceApi, fn) {
+  return common.replaceFunction(traceApi.private_(), 'policy', fn);
+}
 
 function assertAPISurface(traceAPI) {
   assert.strictEqual(typeof traceAPI.enhancedDatabaseReportingEnabled(), 'boolean');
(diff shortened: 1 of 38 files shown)


test/test-trace-api.js
```diff
@@ -94,13 +101,14 @@ describe('Trace Interface', function() {
     
     before(function() {
       traceAPI.enable_(agent);
-      common.init(traceAPI);
-      common.avoidTraceWriterAuth(traceAPI);
+      traceAPI.private_().traceWriter.request = request;
```
DominicKramer: Even though the functionality is only used in this one file, I had code for doing the above in the `common.js` file so that the only place that uses the `private_()` method directly is the `common.js` file.  I did this because I thought it could make it easier to ensure that the use of `private_()` isn't causing issues with the tests.

matthewloring: Yup, I think it would be good to expose this functionality from `common.js` even if it is only used here. Making sure `private_()` is only accessed through `common.js` is important to simplify future refactors that touch agent internals.

test/test-trace-api.js
```diff
@@ -94,13 +101,14 @@ describe('Trace Interface', function() {
...
+      common.avoidTraceWriterAuth();
     });
 
     afterEach(function() {
-      common.cleanTraces(traceAPI);
-      common.clearNamespace(traceAPI);
+      TraceWriter.get().buffer_ = []; // clear traces
```
matthewloring: Why couldn't this be expressed using `cleanTraces`/`clearNamespace`?

test/test-trace-api.js
kjin: `private_` is meant to disappear in a forthcoming PR, since the several layers of indirection are being merged.

In the particular line being highlighted, however, the `private_` access isn't needed since the TraceWriter can be accessed as a singleton.

test/test-trace-api.js
kjin: Clearing the buffer can be expressed with `cleanTraces`; I've made small changes to make this possible.

Since I removed the ability to specify `traceAPI` for `clearNamespace`, there isn't a way for `clearNamespace` to set the `namespace` property on `traceAPI`.

matthewloring: Ok, what do you think of making `clearNamespace` be the only operation in common that still accepts a `traceAPI`? I would be fine with this or the existing approach.

test/test-trace-api.js
```diff
@@ -94,13 +101,14 @@ describe('Trace Interface', function() {
...
-      common.init(traceAPI);
-      common.avoidTraceWriterAuth(traceAPI);
+      TraceWriter.get().request = request;
+      common.avoidTraceWriterAuth();
     });
 
     afterEach(function() {
-      common.cleanTraces(traceAPI);
-      common.clearNamespace(traceAPI);
+      common.clearTraces();
```
matthewloring: `cleanTraces`

PR #520, opened 2017-06-16
warn when creating a child of a closed span

ofrobots:
In cases where context is misattributed, bugs manifest with spans being attributed to an already closed span. It would be good to have a warning for these as a mechanism to debug.

### DIFF
--- a/src/span-data.js
+++ b/src/span-data.js
@@ -94,6 +94,10 @@ function SpanData(agent, trace, name, parentSpanId, isRoot, skipFrames) {
  * @returns {SpanData} The new child trace span data.
  */
 SpanData.prototype.createChildSpanData = function(name, skipFrames) {
+  if (this.span.isClosed()) {
+    this.agent.logger.warn('creating child for an already closed span',
+        name, this.span.name);
+  }
   return new SpanData(this.agent, this.trace, name, this.span.spanId, false,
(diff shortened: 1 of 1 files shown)


src/span-data.js
```diff
@@ -94,6 +94,10 @@ function SpanData(agent, trace, name, parentSpanId, isRoot, skipFrames) {
  * @returns {SpanData} The new child trace span data.
  */
 SpanData.prototype.createChildSpanData = function(name, skipFrames) {
+  if (this.span.isClosed()) {
```
matthewloring: We observed this happening during normal execution for applications that finalize database updates after responding to user requests. I think having this log message available will be good but maybe it makes more sense at `INFO`? `WARN` would probably also be fine but maybe we should change the message to indicate that it may not be a problem?

ofrobots: I would prefer to keep it at `WARN`. It is not just informational in many cases. `WARN` is not the default log level, so the times when a user is going to use this is when they/we are debugging context loss. The problem with updating the message to put caveats is that it increases the amount of spam when you are actually debugging.

matthewloring: Ah sorry, I misremembered and believed `WARN` to log by default. This is definitely fine if off by default.

PR #521, opened 2017-06-20
Explain context propagation

ofrobots:
It would be good to document how tracing works and what the limitations
are.


README.md
```diff
@@ -198,6 +198,15 @@ A `TraceApi` object is guaranteed to be returned by both of these calls, even if
 
 A fully detailed overview of the `TraceApi` object is available [here](doc/trace-api.md).
 
+## How does automatic tracing work?
+
+Node.js provides concurrency through as asynchronous execution model. To be able to keep track of the current request across asynchronous execution boundaries we use the [async-listener][] module. It modifies the core async APIs in Node.js, and Promises, to save and restore the 'execution context' across the async bounaries. In addition, we instrument frameworks (like express, hapi, etc. if loaded) to create request contexts for the execution of the incoming requests. Similarly, we instrument RPC libraries (like mysql, redis, outbound http, etc.) to keep track of the latency of these operations and to correlate them to the incoming request.
```
kjin: A few fixes/proposals:
- First sentence: as -> an
- Last two sentences should indicate that we utilize the saved/restored execution context to instrument frameworks and RPC libraries.
- Nit: Last sentence should also end with "incoming requests" for the sake of uniformity

README.md
```diff
@@ -198,6 +198,15 @@ A `TraceApi` object is guaranteed to be returned by both of these calls, even if
...
+
+* It is possible to use JavaScript code that does its own queuing of callback functions – simulating async hops. For example, one may write a http request buffering library that queues requests and then performs them in a batch in one shot. In such a case, when all the callbacks fire, they seem to be executing on behalf of the wrong incoming context. This problem is called the pooling problem or the [user-space queuing problem][queueing-problem], and is a fundamental limitation of JavaScript. If your application uses such code, you will notice that certain portions of your outbound RPCs do not get traced. In such cases we try to work around the problem through monkey patching, or work with the library authors to fix the code to properly propagate context. However, finding problematic code is not always trivial.
+* Presently, it is not possible for `async-listener` to keep track of async transitions in ES7 async/await functions that are available with Node 7+. If your application uses async functions, we will not be properly track RPCs. We do expect to be able to track async/await functions in the future once we switch to the new [async-hooks][] API in Node 8+.
```
kjin: Should we add a brief note earlier in the README to indicate that tracing requests across async/await functions won't work in Node 7?

README.md
ofrobots: PTAL.

README.md
ofrobots: Added.

README.md
```diff
@@ -18,6 +18,7 @@ This module provides Stackdriver Trace support for Node.js applications. [Stackd
 1. Your application will need to be using Node.js version 4.0 or greater.
 1. You will need a project in the [Google Developers Console][cloud-console]. Your application can run anywhere, but the trace data is associated with a particular project.
 1. [Enable the Trace API](https://console.cloud.google.com/flows/enableapi?apiid=cloudtrace) for your project.
+1. At this point, code using async/await is not supported by this module. See [this section](#how-does-automatic-tracing-work).
```
matthewloring: Do we want to mention it is only untranspiled async/await that is problematic? Not sure if that is obvious or not.

README.md
```diff
@@ -198,6 +199,15 @@ A `TraceApi` object is guaranteed to be returned by both of these calls, even if
 
 A fully detailed overview of the `TraceApi` object is available [here](doc/trace-api.md).
 
+## How does automatic tracing work?
+
+Node.js provides concurrency through an asynchronous execution model. To be able to keep track of the current request across asynchronous execution boundaries we use the [async-listener][] module. It modifies the core async APIs in Node.js, and Promises, to save and restore the 'execution context' across the async bounaries. In addition, we instrument frameworks (like express, hapi, etc. if loaded) to create request contexts for the execution of the incoming requests. Similarly, we instrument RPC libraries (like mysql, redis, outbound http, etc.), utilizing the saved/restored context, to keep track of the latency of these operations and to correlate them to the incoming request.
```
matthewloring: In the first sentence: maybe "event-driven execution model"? I think concurrency models with language level threads would also be considered asynchronous.

README.md
```diff
@@ -198,6 +199,15 @@ A `TraceApi` object is guaranteed to be returned by both of these calls, even if
...
+
+This `async-listener` based mechanism works great in most cases, but it does have some limitations that can prevent us from being able to properly track execution context:
+
+* It is possible to use JavaScript code that does its own queuing of callback functions – simulating async hops. For example, one may write a http request buffering library that queues requests and then performs them in a batch in one shot. In such a case, when all the callbacks fire, they seem to be executing on behalf of the wrong incoming context. This problem is called the pooling problem or the [user-space queuing problem][queueing-problem], and is a fundamental limitation of JavaScript. If your application uses such code, you will notice that certain portions of your outbound RPCs do not get traced. In such cases we try to work around the problem through monkey patching, or work with the library authors to fix the code to properly propagate context. However, finding problematic code is not always trivial.
```
matthewloring: First sentence: maybe "simulating async hops" can be removed or replaced with "effectively merging asynchronous execution contexts"?

Third sentence: "they seem to be executing on behalf of the wrong incoming context" -> "they will execute in the context which flushed the queue instead of the contexts which added the callbacks to the queue". In practice that context usually ends up being correct for one of the callbacks just incorrect for all others (depending on the exact flushing policy).

Fifth sentence: "you will notice that RPCs from many requests are showing up under a single trace or that certain portions of your outbound RPCs do not get traced"

README.md
ofrobots: I am not sure if I want to be too wordy up here. If it is too wordy, we should skip this from here and let the details section explain this out.

matthewloring: I would be fine adding the mention of untranspiled async/await in the async/await section of the detailed explaination below.

README.md
```diff
@@ -198,6 +199,15 @@ A `TraceApi` object is guaranteed to be returned by both of these calls, even if
 
 A fully detailed overview of the `TraceApi` object is available [here](doc/trace-api.md).
 
+## How does automatic tracing work?
+
```
kjin: I see both 'execution context' and 'request context for execution of a request' in consecutive sentences. I think, trying to read this from the perspective of someone who is completely unfamiliar with 'context', that this might be confusing enough to warrant either a rewording or a separate explanation on the difference.

Given that the title of this section is "How does automatic tracing work?" I'd suggest possibly reversing the order of points in this paragraph. For example, it could be written to first talk about what automatic tracing entails (in well-known framework and RPC modules, we automatically inject code to start, annotate, and end trace spans, grouped by request), then the challenge that this poses (that at any given execution context, esp. after crossing asynchronous boundaries such as callbacks and Promise.thens, it is not always possible to determine which request is being served), and then how we approach this problem now (use continuation-local-storage on top of async-listener to propagate information about the current request across asynchronous boundaries).

README.md
ofrobots: done. ptal.

README.md
ofrobots: done.

README.md
ofrobots: done

README.md
ofrobots: I have addressed the comment in the first paragraph. I am not 100% convinced that reversing the flow of the paragraphs works better – how strongly do you feel about this?

kjin: Fairly strongly. As this section is meant to explain how automatic tracing works, I think it would read better to talk about injecting code right away, and then go into the details. Right now it feels like we're jumping straight into the details, without a high-level overview first.

ofrobots: Can you propose a re-wording of this section?

ofrobots: ping

kjin: I propose changing the first two paragraphs to the following:

```markdown
The Trace Agent automatically patches well-known modules to insert calls to functions that start, label, and end spans to measure latency of RPCs (such as mysql, redis, etc.) and incoming requests (such as express, hapi, etc.). As each RPC is typically performed on behalf of an incoming request, we must make sure that this association is accurately reflected in span data. To provide a uniform, generalized way of keeping track of which RPC belongs to which incoming request, we rely on the [continuation-local-storage][] module to propogate "trace context" across asynchronous boundaries by storing trace information in continuation-local storage.

This method, which relies on [async-listener][] to preserve continuations over asynchronous boundaries, works great in most cases. However, it does have some limitations that can prevent us from being able to properly propagate trace context:
```

This requires an additional link at the bottom:

```markdown
[continuation-local-storage]: https://www.npmjs.com/package/continuation-local-storage
```

ofrobots: PTAL

kjin: LGTM. Thanks!

ofrobots:
The node 8 failure is real, but unrelated.

PR #522, opened 2017-06-23
Update dev dep on got

matthewloring:


matthewloring:
This change verifies that the trace agent no longer breaks got >5.

matthewloring:
Updated in https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/563.

PR #528, opened 2017-07-18
Fix breaking change introduced in 2.1.1

kjin:


src/trace-api.js
```diff
@@ -53,20 +53,37 @@ var nullSpan = {};
...
  * be configured.
  */
-function TraceAgent(name, logger, config) {
+function TraceAgent(name) {
   this.pluginName_ = name;
+  this.disable(); // start disabled
+}
+
+/**
+ * TODO
```
matthewloring: Provide a more descriptive TODO.

ofrobots:
It would be good to start following [conventional commits](https://conventionalcommits.org/).

PR #529, opened 2017-07-19
build: use conventional commit

ofrobots:
Use https://conventionalcommits.org/ with the [Angular style][].

Use `commitlint` to ensure that the commit messages follow the style
`husky` to ensure the lint is applied at git commit time.

To reedit the commit message after a lint failure, one can do something
like:
```
git commit --edit --file=.git/COMMIT_EDITMSG'
```
[Angular style]: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
     "changelog": "./bin/run-changelog.sh",
     "coverage": "./bin/run-test.sh -c",
     "bump": "./bin/run-bump.sh",
-    "closure": "./node_modules/.bin/closure-npc"
+    "closure": "./node_modules/.bin/closure-npc",
+    "commitmsg": "commitlint -e"
   },
   "keywords": [
     "google",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -31,12 +34,14 @@
     "glob": "^7.0.3",
     "google-auto-auth": "^0.7.0",
     "got": "^5.7.1",
+    "husky": "^0.14.3",
```
matthewloring: How does this get invoked?

ofrobots: It probably installs the githooks as part of the install script

https://github.com/typicode/husky/blob/master/package.json#L11

DominicKramer:
I just want to make sure I understand how this will be used.

If we try to commit, and the linter identifies an error, it will abort the commit and output the error.  That is why we need to use `git commit --edit --file=.git/COMMIT_EDITMSG` instead of `git commit --amend`.

If there is only linter warnings, does the commit still go through, so that we need to use `git commit --amend` to go back and fix anything identified in the warning?

ofrobots:
Yes.

PR #533, opened 2017-07-28
fix: add explicit active flag in trace agent

kjin:
Using the custom span API while the Trace Agent is disabled causes an error, since the agent is in an inconsistent state where it uses its enabled implementation without being explicitly enabled. (Note that this doesn't affect plugin usage because plugins are not loaded if the trace agent is disabled).

This change adds an `active_` field in TraceAgent instead of relying on whether the namespace exists. This ensures that the agent is actually disabled completely.

### DIFF
--- a/src/trace-api.js
+++ b/src/trace-api.js
@@ -50,7 +50,8 @@ var nullSpan = {};
  */
 function TraceAgent(name) {
   this.pluginName_ = name;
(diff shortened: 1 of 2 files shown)


kjin:
@DominicKramer The included test does just that. It fails if the fix is not included.

I provided a description of how the error could manifest in the PR summary.

PR #534, opened 2017-07-28
fix: don't warn about trace loading before itself on windows

kjin:
On Windows, the Trace Agent emits a warning saying that it was loaded before itself, due to module names using system-dependent org/module separators.

### DIFF
--- a/index.js
+++ b/index.js
@@ -36,7 +36,7 @@ var modulesLoadedBeforeTrace = [];
 
 var traceAgent = new TraceAgent('Custom Span API');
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -35,9 +36,10 @@ var modulesLoadedBeforeTrace = [];
 
 var traceAgent = new TraceAgent('Custom Span API');
 
+var traceModuleName = '@google-cloud' + path.sep + 'trace-agent';
```
matthewloring: Does `path.join('@google-cloud', 'trace-agent')` work here?

PR #538, opened 2017-08-10
Async hooks based context tracking

matthewloring:
Introduces a lower overhead, more correct context tracking
mechanism based on async hooks that is available when using the
trace agent on newer versions of node (>=8). To enable async 
hooks based context, set the GCLOUD_TRACE_NEW_CONTEXT environment
variable. This change is not expected to have observable behavior
difference but due to the nature of context propagation it is 
difficult to say for sure without real world testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
   - psql -c "alter user postgres with password 'Password12!';" -U postgres
 
 env:
-  - CXX=g++-4.8
+  - CXX=g++-4.8 GCLOUD_TRACE_NEW_CONTEXT=1
 
 script:
   - ./bin/run-test.sh -c
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,8 +23,8 @@ install:
   - npm install
   - ps: ForEach ($fix in Get-ChildItem -Directory test/plugins/fixtures/*) { cd $fix; npm install --silent }
   - ps: cd ..\..\..\..
-  # set GCLOUD_PROJECT
   - SET GCLOUD_PROJECT=0
+  - SET GCLOUD_TRACE_NEW_CONTEXT=1
 
 before_test:
   - SET PATH=C:\Program Files\MySql\MySQL Server 5.7\bin;%PATH%
--- a/index.js
+++ b/index.js
@@ -20,7 +20,10 @@ var filesLoadedBeforeTrace = Object.keys(require.cache);
 
 // Load continuation-local-storage first to ensure the core async APIs get
 // patched before any user-land modules get loaded.
-require('continuation-local-storage');
+if (require('semver').satisfies(process.version, '<8') &&
+    process.env.GCLOUD_TRACE_NEW_CONTEXT) {
+  require('continuation-local-storage');
+}
 
 var path = require('path');
 var cls = require('./src/cls.js');
--- a/src/cls-ah.js
+++ b/src/cls-ah.js
@@ -1,0 +1,113 @@
+/**
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const asyncHook = require('async_hooks');
+
+const wrappedSymbol = Symbol('context_wrapped');
+let contexts = {};
+let current = {};
+
+asyncHook.createHook({init, before, destroy}).enable();
+
+function Namespace() {}
+
+Namespace.prototype.get = function(k) {
+  return current[k];
+};
+
+Namespace.prototype.set = function(k, v) {
+  current[k] = v;
+};
+
+Namespace.prototype.run = function(fn) {
+  const oldContext = current;
+  current = {};
+  const res = fn();
+  current = oldContext;
+  return res;
+};
+
+Namespace.prototype.runAndReturn = Namespace.prototype.run;
+
+Namespace.prototype.bind = function(cb) {
+  if (cb[wrappedSymbol] || !current) {
+    return cb;
+  }
+  const boundContext = current;
+  const contextWrapper = function() {
+    const oldContext = current;
+    current = boundContext;
+    const res = cb.apply(this, arguments);
+    current = oldContext;
+    return res;
+  };
+  contextWrapper[wrappedSymbol] = true;
+  Object.defineProperty(contextWrapper, 'length', {
+    enumerable: false,
+    configurable: true,
+    writable: false,
+    value: cb.length
+  });
+  return contextWrapper;
+};
+
+const eventEmitterMethods =
+  [ 'addListener', 'on', 'once', 'prependListener', 'prependOncelistener' ];
+
+// This function is not technically needed and all tests currently pass without it
+// (after removing call sites). While it is not a complete solution, restoring
+// correct context before running every request/response event handler reduces
+// the number of situations in which userspace queuing will cause us to lose context.
+Namespace.prototype.bindEmitter = function(ee) {
+  var ns = this;
+  eventEmitterMethods.forEach(function(method) {
+    const oldMethod = ee[method];
+    ee[method] = function(e, f) { return oldMethod.call(this, e, ns.bind(f)); };
+  });
+};
+
+var namespace = new Namespace();
+
+// AsyncWrap Hooks
+
+function init(uid, provider, parentUid, parentHandle) {
+  contexts[uid] = current;
+}
+
+function before(uid) {
+  if (contexts[uid]) {
+    current = contexts[uid];
+  }
+}
+
+function destroy(uid) {
+  delete contexts[uid];
+}
+
+module.exports = {
+  createNamespace: function() {
+    return namespace;
+  },
+
+  destroyNamespace: function() {
+    current = {};
+    contexts = {};
+  },
+
+  getNamespace: function() { return namespace; }
+};
--- a/src/trace-api.js
+++ b/src/trace-api.js
@@ -24,6 +24,9 @@ var Trace = require('./trace.js');
 var SpanData = require('./span-data.js');
 var uuid = require('uuid');
 var TracingPolicy = require('./tracing-policy.js');
+var semver = require('semver');
+
+var ROOT_SPAN_STACK_OFFSET = semver.satisfies(process.version, '>=8') ? 0 : 2;
 
 /**
  * Phantom implementation of the trace api. When disabled, a TraceAgent instance
@@ -170,7 +173,7 @@ TraceAgent.prototype.runInRootSpan = function(options, fn) {
       options.name, /* Span name */
       parentId, /* Parent's span ID */
       true, /* Is root span */
-      options.skipFrames ? options.skipFrames + 2 : 2);
+      ROOT_SPAN_STACK_OFFSET + (options.skipFrames || 0));
     rootContext.span.kind = 'RPC_SERVER';
     cls.setRootContext(rootContext);
     return fn(rootContext);
--- a/test/plugins/common.js
+++ b/test/plugins/common.js
@@ -21,9 +21,18 @@ proxyquire('gcp-metadata', {
   'retry-request': require('request')
 });
 
+var semver = require('semver');
+
 var logger = require('@google-cloud/common').logger;
 var trace = require('../..');
 var cls = require('../../src/cls.js');
+if (semver.satisfies(process.version, '>=8')) {
+  // Monkeypatch the monkeypatcher
+  var oldIt = global.it;
+  global.it = function it(title, fn) {
+    return oldIt.call(this, title, cls.createNamespace().bind(fn));
+  };
+}
 var tracePolicy = require('../../src/tracing-policy.js');
 var TraceWriter = require('../../src/trace-writer.js');
 
--- a/circle.yml
+++ b/circle.yml
@@ -5,3 +5,4 @@ machine:
     - redis
   environment:
     GCLOUD_PROJECT: 0
+    GCLOUD_TRACE_NEW_CONTEXT: 1
--- a/src/cls.js
+++ b/src/cls.js
@@ -16,8 +16,10 @@
 
 'use strict';
 
-
-var cls = require('continuation-local-storage');
+var semver = require('semver');
+var cls = semver.satisfies(process.version, '>=8') &&
+          process.env.GCLOUD_TRACE_NEW_CONTEXT ?
+            require('./cls-ah.js') : require('continuation-local-storage');
 
 /** @const {string} */
 var TRACE_NAMESPACE = 'com.google.cloud.trace';
--- a/test/plugins/test-trace-grpc.js
+++ b/test/plugins/test-trace-grpc.js
@@ -284,6 +284,19 @@ Object.keys(versions).forEach(function(version) {
 
       grpc = require(versions[version]);
 
+      var oldRegister = grpc.Server.prototype.register;
+      grpc.Server.prototype.register = function register(n, h, s, d, m) {
+        var result = oldRegister.call(this, n, h, s, d, m);
+        var oldFunc = this.handlers[n].func;
+        this.handlers[n].func = function() {
+          if (cls.getRootContext()) {
+            cls.setRootContext(null);
+          }
+          return oldFunc.apply(this, arguments);
+        };
+        return result;
+      };
+
       // This metadata can be used by all test methods.
       metadata = new grpc.Metadata();
       metadata.set('a', 'b');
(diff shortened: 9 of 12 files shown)


matthewloring:
I'm still working through the right test plan for this and I'm not sold on the way namespaces are represented in the async hook based implementation but all the high level structure is there and the tests are passing. I would be interested to get some early feedback on this, especially on the required changes to the tests (grpc required the most invasive changes).

src/trace-api.js
```diff
@@ -165,11 +166,17 @@ TraceAgent.prototype.runInRootSpan = function(options, fn) {
     // Create a new root span, and invoke fn with it.
     var traceId = incomingTraceContext.traceId || (uuid.v4().split('-').join(''));
     var parentId = incomingTraceContext.spanId || '0';
+    var frameOffset;
+    if (semver.satisfies(process.version, '>=8')) {
```
evanlucas: since this won't change at runtime, should this be cached?

matthewloring: Thanks for the suggestion! semver checks on the critical path should now be gone.

index.js
```diff
@@ -20,7 +20,9 @@ var filesLoadedBeforeTrace = Object.keys(require.cache);
 
 // Load continuation-local-storage first to ensure the core async APIs get
 // patched before any user-land modules get loaded.
-require('continuation-local-storage');
+if (require('semver').satisfies(process.version, '<8')) {
```
ofrobots: Note that you're loading `semver` before `cls`. Hopefully this is fine?

src/cls-ah.js
```diff
@@ -0,0 +1,146 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
```
ofrobots: Is 2015 correct?

src/trace-api.js
```diff
@@ -169,7 +172,8 @@ TraceAgent.prototype.runInRootSpan = function(options, fn) {
       options.name, /* Span name */
       parentId, /* Parent's span ID */
       true, /* Is root span */
-      options.skipFrames ? options.skipFrames + 2 : 2);
+      options.skipFrames ?
+        options.skipFrames + ROOT_SPAN_STACK_OFFSET : ROOT_SPAN_STACK_OFFSET);
```
ofrobots: `ROOT_SPAN_STACK_OFFSET + (options.skipFrames || 0)`

test/plugins/common.js
```diff
@@ -21,9 +21,18 @@ proxyquire('gcp-metadata', {
...
+var semver = require('semver');
+
 var logger = require('@google-cloud/common').logger;
 var trace = require('../..');
 var cls = require('../../src/cls.js');
+if (semver.satisfies(process.version, '>=8')) {
+  // Monkeypatch the monkeypatcher
+  var oldIt = global.it;
+  global.it = function it(title, fn) {
+    return oldIt.call(this, title, cls.createNamespace().bind(fn));
```
ofrobots: So you basically are running each test in a top-level namespace. This behaviour has changed between cls and cls-ah?

src/cls-ah.js
```diff
@@ -0,0 +1,146 @@
...
+const contexts = {};
+var current = new Context();
+
+const hook = asyncHook.createHook({init, before, after, destroy});
+hook.enable();
+
+// Core Zone Api
+
+function Context() {
+  this.properties = {};
```
ofrobots: discussion: use Symbol for extra privacy? Debugging would be harder though.

src/cls-ah.js
```diff
@@ -0,0 +1,146 @@
...
+    current = thisContext;
+    const res = cb.apply(this, arguments);
+    current = oldContext;
+    return res;
+  };
+  wrapped[wrappedSymbol] = true;
+  return wrapped;
+};
+
+Context.prototype.run = function(cb, applyThis, applyArgs) {
```
ofrobots: nit: cb implies async callback. Prefer to call it 'fn'.

src/cls-ah.js
```diff
@@ -0,0 +1,146 @@
...
+  const oldContext = current;
+  current = new Context();
```
ofrobots: Question: there are three contexts objects in this function: `current`, `new Context` and `this`. I believe `this === current` at the start of the function?

src/cls-ah.js
```diff
@@ -0,0 +1,146 @@
...
+  return current ? current.get(k) : null;
+};
+
+Namespace.prototype.set = function(k, v) {
+  if (current) {
+    current.set(k, v);
+  }
+};
+
+Namespace.prototype.run = function(cb) {
```
ofrobots: cb vs. fn

src/cls-ah.js
```diff
@@ -0,0 +1,146 @@
...
+  if (current) {
+    return current.run(cb);
+  } else {
+    return cb();
+  }
+};
+
+Namespace.prototype.runAndReturn = function(cb) {
```
ofrobots: cb vs fn

src/cls-ah.js
```diff
@@ -0,0 +1,146 @@
...
+  if (current) {
+    return current.run(cb);
+  } else {
+    return cb();
+  }
+};
+
+Namespace.prototype.bind = function(cb) {
```
ofrobots: cb vs fn

src/cls-ah.js
ofrobots: What is the difference with `Namespace#run`?

index.js
matthewloring: Yup, `semver` never calls `require` so this should be safe.

src/cls-ah.js
matthewloring: Fixed.

src/cls-ah.js
matthewloring: Fixed.

src/cls-ah.js
matthewloring: Fixed.

src/cls-ah.js
matthewloring: Fixed.

src/cls-ah.js
matthewloring: fixed.

src/trace-api.js
matthewloring: Fixed.

test/plugins/common.js
matthewloring: This should be unobservable for now since nobody else can store anything in our context (without requiring it by filename out of our module).

src/cls-ah.js
matthewloring: Context has been removed in the latest iteration and just replaced with a top level object.

src/cls-ah.js
matthewloring: This code has been cleaned up with the removal of Context.

src/cls-ah.js
matthewloring: These now share a code path in the new implementation.

src/cls-ah.js
```diff
@@ -0,0 +1,146 @@
...
+  if (current) {
+    return current.wrap(cb);
+  } else {
+    return cb;
+  }
+};
+
+Namespace.prototype.bindEmitter = function(ee) {
```
matthewloring: Good catch, adding them.

test/plugins/common.js
```diff
@@ -21,9 +21,18 @@ proxyquire('gcp-metadata', {
   'retry-request': require('request')
 });
 
+var semver = require('semver');
+
 var logger = require('@google-cloud/common').logger;
 var trace = require('../..');
 var cls = require('../../src/cls.js');
+if (semver.satisfies(process.version, '>=8')) {
+  // Monkeypatch the monkeypatcher
```
matthewloring: :speak_no_evil:

matthewloring:
It looks like appveyor is hanging on hiredis compilation but everything else looks clean. Comments should be addressed. I think this is ready for another look. @kjin @ofrobots

.travis.yml
```diff
@@ -19,7 +19,7 @@ before_install:
   - psql -c "alter user postgres with password 'Password12!';" -U postgres
 
 env:
-  - CXX=g++-4.8
+  - CXX=g++-4.8 GCLOUD_TRACE_NEW_CONTEXT=1
```
ofrobots: Do these need to be on separate lines?
https://docs.travis-ci.com/user/environment-variables/#Defining-public-variables-in-.travis.yml

appveyor.yml
```diff
@@ -25,6 +25,7 @@ install:
   - ps: cd ..\..\..\..
   # set GCLOUD_PROJECT
```
ofrobots: wow, such a useful comment ;)

index.js
```diff
@@ -20,7 +20,10 @@ var filesLoadedBeforeTrace = Object.keys(require.cache);
 
 // Load continuation-local-storage first to ensure the core async APIs get
 // patched before any user-land modules get loaded.
-require('continuation-local-storage');
+if (require('semver').satisfies(process.version, '<8') &&
```
ofrobots: In the future if `semver` adds a dependency, it would not be caught by the `filesLoadedBeforeTrace` and the requisite test.

src/cls-ah.js
```diff
@@ -0,0 +1,118 @@
...
+
+Namespace.prototype.run = function(fn) {
+  const oldContext = current;
+  current = {};
+  const res = fn();
+  current = oldContext;
+  return res;
+};
+
+Namespace.prototype.runAndReturn = function(fn) {
```
ofrobots: `runAndReturn = run` would work?

src/cls-ah.js
```diff
@@ -0,0 +1,118 @@
...
+const eventEmitterMethods =
+  [ 'addListener', 'on', 'once', 'prependListener', 'prependOncelistener' ];
+
+Namespace.prototype.bindEmitter = function(ee) {
+  for (var method of eventEmitterMethods) {
+    patchEventEmitterMethod(ee, method, this);
+  }
+};
+
+var n = null;
```
ofrobots: use a less terse variable name.

src/cls-ah.js
```diff
@@ -0,0 +1,118 @@
...
+    return cb;
+  }
+  const boundContext = current;
+  const wrapped = cb.length === 0 ? function() {
+    const oldContext = current;
+    current = boundContext;
+    const res = cb.apply(this, arguments);
+    current = oldContext;
+    return res;
+  } : function(a) {
```
ofrobots: There is no difference between the functions other that the second one shows a single but unused argument. What is going on here?

src/cls-ah.js
```diff
@@ -0,0 +1,118 @@
...
+
+function after(uid) {}
+
+function destroy(uid) {
+  delete contexts[uid];
+}
+
+module.exports = {
+  createNamespace: function() {
+    n = new Namespace();
```
ofrobots: Given that `n` is a singleton, you don't expect `createNamespace` to be called more than once, correct? If that is the case, assert rather than clobbering.

src/cls-ah.js
```diff
@@ -0,0 +1,118 @@
...
+
+'use strict';
+
+const asyncHook = require('async_hooks');
+
+const wrappedSymbol = Symbol('context_wrapped');
+const contexts = {};
+var current = {};
+
+const hook = asyncHook.createHook({init, before, after, destroy});
```
ofrobots: `after` is a nop. Why do you need to pass it?

src/cls-ah.js
```diff
@@ -0,0 +1,118 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const asyncHook = require('async_hooks');
+
+const wrappedSymbol = Symbol('context_wrapped');
+const contexts = {};
```
ofrobots: Would a `Map` be preferrable?

src/cls-ah.js
```diff
@@ -0,0 +1,118 @@
...
+var current = {};
+
+const hook = asyncHook.createHook({init, before, after, destroy});
+hook.enable();
+
+function Namespace() {}
+
+Namespace.prototype.get = function(k) {
```
ofrobots: Is it possible for me to set a value to `null`? You use `null` as a sentinel value which would be indistinguishable from the value being `null`.

src/cls-ah.js
```diff
@@ -0,0 +1,118 @@
...
+Namespace.prototype.runAndReturn = function(fn) {
+  return this.run(fn);
+};
+
+Namespace.prototype.bind = function(cb) {
+  if (cb[wrappedSymbol] || !current) {
+    return cb;
+  }
+  const boundContext = current;
+  const wrapped = cb.length === 0 ? function() {
```
ofrobots: Suggest naming the function. Perhaps `withContext` or `restoreContext`?

src/cls-ah.js
```diff
@@ -0,0 +1,118 @@
...
+  const oldMethod = ee[method];
+  ee[method] = function(e, f) { return oldMethod.call(this, e, ns.bind(f)); };
+}
+
+const eventEmitterMethods =
+  [ 'addListener', 'on', 'once', 'prependListener', 'prependOncelistener' ];
+
+Namespace.prototype.bindEmitter = function(ee) {
+  for (var method of eventEmitterMethods) {
+    patchEventEmitterMethod(ee, method, this);
```
ofrobots: Can you inline `patchEventEmitterMethod` here?

.travis.yml
matthewloring: Separate lines means disjoint run configurations (one set of tests at each node version with `CXX=g++-4.8` and another set of tests at each node version with `GCLOUD_TRACE_NEW_CONTEXT=1`. Writing them on the same line applies both env vars to each test configuration. https://docs.travis-ci.com/user/environment-variables/#Defining-Multiple-Variables-per-Item

src/cls-ah.js
matthewloring: `Map` captures the intended usage of the variable better but it did not have better performance the last time I looked into it.

src/cls-ah.js
matthewloring: Fixed.

src/cls-ah.js
matthewloring: Fixed.

src/cls-ah.js
matthewloring: Fixed.

src/cls-ah.js
matthewloring: Fixed.

appveyor.yml
matthewloring: Removed.

index.js
matthewloring: This is true. We can reorder things and add logic to remove `semver` from the list of modules loaded before trace. However, it seems highly unlikely that `semver` would take a dependency on module that would be problematic for us. Also, since it is required by our module instead of by the user, any issues it causes should manifest in our own tests.

src/cls-ah.js
matthewloring: All named.

src/cls-ah.js
matthewloring: This was experimentation to get the correct value of `Function.length` for wrapped functions. I've updated this to be more clear/complete.

src/cls-ah.js
matthewloring: We now use `undefined` to signify no value.

src/cls-ah.js
matthewloring: The new approach uses `Object#defineProperty`. @bmeurer are there any expected performance implications of redefining a function's length? I played around with some microbenchmarks that didn't show any impact but am concerned the microbenchmarks may not show everything.

src/cls-ah.js
matthewloring: The contract I'm aiming for is that calling `createNamespace` -> `destroyNamespace` -> `createNamespace` gives you completely fresh state as was the case with `continuation-local-storage`. To achieve this, I've updated `createNamespace`/`deleteNamespace` to wipe out context related state and leave the `namespace` singleton alone.

matthewloring:
@ofrobots All comments should be addressed. PTAL.

src/cls-ah.js
bmeurer: It could impact `F.p.bind`, but should be fine otherwise.

matthewloring: Would the impact be in performance alone or is there a correctness concern here? Also, would the issue arise if the function is already bound or if `F.p.bind` is called later?

bmeurer: I think it should be fine both correctness and performance wise.

matthewloring: Thanks!

src/cls-ah.js
```diff
@@ -0,0 +1,110 @@
...
+
+'use strict';
+
+const asyncHook = require('async_hooks');
+
+const wrappedSymbol = Symbol('context_wrapped');
+let contexts = {};
+let current = {};
+
+const hook = asyncHook.createHook({init, before, destroy});
```
ofrobots: nit: Get rid of the variable?

src/cls-ah.js
```diff
@@ -0,0 +1,110 @@
...
+hook.enable();
+
+function Namespace() {}
+
+Namespace.prototype.get = function(k) {
+  return current ? current[k] : undefined;
```
ofrobots: When would current be `null`?

src/cls-ah.js
matthewloring: Done.

src/cls-ah.js
matthewloring: This can no longer be the case. Removed the checks.

test/plugins/common.js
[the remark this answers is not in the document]
Qard: 🙉

src/cls-ah.js
```diff
@@ -0,0 +1,109 @@
...
+    writable: false,
+    value: cb.length
+  });
+  return wrapped;
+};
+
+const eventEmitterMethods =
+  [ 'addListener', 'on', 'once', 'prependListener', 'prependOncelistener' ];
+
+Namespace.prototype.bindEmitter = function(ee) {
```
ofrobots: As discussed, this needs a comment about why it is need.

matthewloring: Done.

Issue #525, opened 2017-07-16
Provide an example with GCLOUD_DIAGNOSTICS_CONFIG usage


oshalygin:
Reading the README it states that this environment variable can point to the configuration file but its unclear on what that actually means.  Here are the questions that run through my head:

1.  Is this configuration file exported as a javascript object? 
2. Is this configuration file exported as a JSON file?
3. Does this file include the agent startup code? (`require('@google-cloud/trace-agent').start({......`)

I'll try to figure it out on my own and i'll update this issue and follow it up with a PR if there is any interest in this.

oshalygin:
Confirmed what works:

This is performed with just step 1.  To get this working I have the following file in my project:

_./utilities/trace.js_
```javascript

module.exports = {

  // Log levels: 0-disabled,1-error,2-warn,3-info,4-debug
  // Default: 1
  logLevel: 2,

  // Default: true
  enabled: true,

  // If true, information about query parameters and results will be
  // attached to spans representing database operations.
  // Default: false
  enhancedDatabaseReporting: true,

  // The maximum number of characters reported on a label value. This
  // cannot exceed 16383, the maximum value accepted by the service.
  // Default: 512
  maximumLabelValueSize: 512,

  // @type {number} max number of frames to include on traces (0 disables)
  // Default: 10
  stackTraceLimit: 10,

  // We buffer the captured traces for `flushDelaySeconds` before publishing
  // to the trace API; unless the buffer fills up before then.
  // See `bufferSize`.
  // Default: 30
  flushDelaySeconds: 30,

  // If paths are present in this array, then these paths will be ignored before
  // `samplingRate` based decisions are made. Paths must include a leading
  // forward slash and be of the form:
  //   /componentOne/componentTwo/...
  // Paths can additionally be classified by regex in which case any path matching
  // any provided regex will be ignored.
  // We ignore the health checker probes (/_ah/health) by default.
  ignoreUrls: [
    '/_ah/health',
    '/healthz'
  ],

  // An upper bound on the number of traces to gather each second. If set to 0,
  // sampling is disabled and all traces are recorded. Sampling rates greater
  // than 1000 are not supported and will result in at most 1000 samples per
  // second. Some Google Cloud environments may further limit this rate.
  // Default: 10
  samplingRate: 10,

  // The number of transactions we buffer before we publish to the trace
  // API, unless we hit `flushDelaySeconds` first.
  // Default: 1000
  bufferSize: 1000,

[...]

PR #539, opened 2017-08-11
feat: Add GCLOUD_TRACE_CONFIG env config

oshalygin:
- This implementation provides a means of setting
  the configuration file from the environment variable,
  GCLOUD_TRACE_CONFIG.

Addresses: #525 

If the decision is to change the order, I can do that as well and update the tests accordingly.

### DIFF
--- a/index.js
+++ b/index.js
@@ -53,6 +53,7 @@ for (var i = 0; i < filesLoadedBeforeTrace.length; i++) {
  * @return A normalized configuration object.
  */
 function initConfig(projectConfig) {
+
   var envConfig = {
     logLevel: process.env.GCLOUD_TRACE_LOGLEVEL,
     projectId: process.env.GCLOUD_PROJECT,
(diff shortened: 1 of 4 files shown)


oshalygin:
If this is approved, I suggest we also add language back to the `README`

index.js
```diff
@@ -62,7 +63,28 @@ function initConfig(projectConfig) {
...
+    try {
+      envSetConfig = require(path.resolve(process.env.GCLOUD_TRACE_CONFIG));
+    } catch (error) {
+
+      var logger = common.logger({
+        level: common.logger.LEVELS[2],
+        tag: '@google-cloud/trace-agent',
+      });
+
+      logger.error('The environment variable GCLOUD_TRACE_CONFIG was set,' +
```
matthewloring: My personal preference would be to just require the value of `GCLOUD_TRACE_CONFIG` directly if it is present. This would reduce the chance that a user believes their environmental configuration is in effect when it really failed to load in case they miss the log (especially since many configuration may not be immediately observable such as trace buffer size, flush delay, or sampling rate). WDYT? If this seems reasonable I'm happy to update at landing time.

matthewloring:
Definitely agree that this should be documented in the `README`. If you have a preference on the wording you're welcome to include it in this PR or we can add it afterwards (if we did that it would probably read something like https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/commit/ead940235f2f3652a88e8154cc68ca10048b56b3). Thanks for putting this together!

index.js
oshalygin: That actually makes sense, and I agree with you, I rather have it vomit in trying to start the application/container so that its super obvious.  I'll update it shortly.

oshalygin:
@matthewloring Added the documentation back to the `README` from the commit you linked 👍

oshalygin:
Much appreciated thanks @matthewloring.  Thanks a bunch for the discussion/help.

matthewloring:
Anytime :smile:

PR #541, opened 2017-08-17
Fix http-e2e for node 8.4.0

matthewloring:
This change removes failing http2 tests introduced in node v8.4.0 from our CI. There is no need to run https/http2 tests on every PR.


ofrobots:
It would be good to add some explanation for posterity.

My guess is that the old glob is now matching http2?

PR #542, opened 2017-08-22
feat: add public API for root span id

ofrobots:
Make it possible for external users to get a unique identified for the
root span. This can be used to uniquely identify the top level
incoming http request context for example.


matthewloring:
Typo in the commit message `identified` -> `identifier`.

src/trace-api.js
```diff
@@ -33,6 +33,7 @@ var TracingPolicy = require('./tracing-policy.js');
 var phantomApiImpl = {
   enhancedDatabaseReportingEnabled: function() { return false; },
   runInRootSpan: function(opts, fn) { return fn(null); },
+  // getRootSpanId: function() { return null; },
```
matthewloring: Commented

src/trace-api.js
```diff
@@ -177,6 +178,22 @@ TraceAgent.prototype.runInRootSpan = function(options, fn) {
...
 
 /**
+ * Returns a unique identifier for the current root span. This can be used to
+ * uniquely identify the current root span. If there is no current root span,
+ * or if we have lost context, this will return null. The structure and the
+ * length of the returned string should be treated opaquely - the only guarantee
+ * is that the value would unique for every root span.
+ * @returns {string} the root span id, or null if it doesn't exist
+ */
+TraceAgent.prototype.getRootSpanId = function() {
```
draffensperger: Could we call this `getRootTraceId` instead?

matthewloring: We had previously discussed having `getCurrentTraceId` as the name but opted for this one because `TraceId` is an trace service concept and our agent API is built around the concept of spans. We felt this name would be more consistent with the rest of the existing API. However, for the logging use case, I think it is important that it is the trace id and not some other unique identifier so that the trace service can associate logs with the correct trace. Renaming it makes sense to me.

ofrobots: The suggestion does make sense from a trace service API point of view, however, the [public JS API ](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/doc/trace-api.md#trace-spans)we expose from the module uses concepts like `runInRootSpan`. We don't differentiate Root Trace and Root Span to our users, so `RootSpanId` might make more sense. WDYT?

draffensperger: Yeah, I think I'm OK with calling it `getRootSpanId` given the context of the JS API.

Eventually if we separate out the context propagation library from the trace library and make it a shared dependency of logging and trace, then we could re-evaluate what to name it then.

ofrobots: It seems that I totally missed @matthewloring 's comment above – the browser tab was stale.

How about we don't use any trace / span terminology at all? E.g. something `getCurrentContextId`? Perhaps `getCurrentTraceId` is fine.

draffensperger: I like `getCurrentContextId` as it does feel pretty general and in the situation where we propagate context without trace sampling being enabled, it helps subtly clarify that there is not actually a trace associated with this ID even though we are passing around the context.

matthewloring: `getCurrentContextId` sounds good to me.

src/trace-api.js
```diff
@@ -33,6 +33,7 @@ var TracingPolicy = require('./tracing-policy.js');
 var phantomApiImpl = {
   enhancedDatabaseReportingEnabled: function() { return false; },
   runInRootSpan: function(opts, fn) { return fn(null); },
+  getCurrentContext: function() { return null; },
```
draffensperger: Should this be `getCurrentContextId`?

ofrobots:
Renamed to `getCurrentContextId`. Will land once the CI is green.

PR #543, opened 2017-08-24
get rid of commitlint

ofrobots:
We tried this experiment, I am not sure it is successful:
Too onerous. Unfriendly to non-collaborators. Doesn't work well with github squash and rebase workflow.

Opinions @GoogleCloudPlatform/node-team ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,7 @@
     "changelog": "./bin/run-changelog.sh",
     "coverage": "./bin/run-test.sh -c",
     "bump": "./bin/run-bump.sh",
-    "closure": "./node_modules/.bin/closure-npc",
-    "commitmsg": "commitlint -e"
+    "closure": "./node_modules/.bin/closure-npc"
   },
   "keywords": [
     "google",
@@ -24,8 +23,6 @@
     "node": ">=0.12"
   },
   "devDependencies": {
-    "@commitlint/cli": "^3.0.3",
-    "@commitlint/config-angular": "^3.0.3",
     "@google-cloud/datastore": "^1.0.2",
     "changelog-maker": "^2.2.2",
     "closure-npc": "*",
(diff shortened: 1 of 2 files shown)


kjin:
LGTM. We should still follow conventional commits, especially since we've already got the ball rolling... mostly. It should be possible today to write a GitHub integration to enforce this by failing a check if the PR's title doesn't satisfy conventional commits, since squash+merge'd commits take on the title of the PR itself. Enforcing the body is a little different issue... I've emailed GitHub support about whether there's a feature for Squash and Merge to take on the contents of the initial PR description.

So in short, I think we should still enforce this, but at the GitHub integrations level rather than at the git CLI level.

package.json
```diff
@@ -10,8 +10,7 @@
     "changelog": "./bin/run-changelog.sh",
     "coverage": "./bin/run-test.sh -c",
     "bump": "./bin/run-bump.sh",
-    "closure": "./node_modules/.bin/closure-npc",
-    "commitmsg": "commitlint -e"
+    "closure": "./node_modules/.bin/closure-npc"
```
matthewloring: Just since it's highlighted in this change, should we remove this at some point?

ofrobots: Done.

AdriVanHoudt:
There is https://github.com/conventional-changelog/validate-commit-msg just an fyi

ofrobots:
@AdriVanHoudt Thanks! The main challenge we saw was that commit hook time validation puts the burden on people trying to contribute to the repo. In reality the commit time message doesn't matter but rather the message we use at squash-and-merge time in GitHub, which does not get validated.

marionebl:
Sorry to read `commitlint` did not match your use case. Paul Irish recently started work on https://github.com/paulirish/commitlintbot. Have not tried it myself but it looks as if this could solve the Squash/Merge problem.

kjin:
Thanks for pointing us in that direction! Looking at the high level
description it seems like this could be very useful to us.

PR #548, opened 2017-08-30
Trace API function for trace writer project ID

draffensperger:
This will be used by the best-effort logs correlation changes to be able to format the `LogEntry.trace` field in the specified `projects/[PROJECT-ID]/traces/[TRACE-ID]` format (per https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry).


src/trace-api.js
```diff
@@ -197,6 +199,23 @@ TraceAgent.prototype.getCurrentContextId = function() {
...
 
 /**
+ * Returns the projectId that was either configured or auto-discovered by the
+ * TraceWriter. Note that the auto-discovery is done asynchronously, so this
+ * will return null until the projectId auto-discovery completes.
+ */
+TraceAgent.prototype.getWriterProjectId = function() {
+  var traceWriter;
+  try {
+     traceWriter = TraceWriter.get();
```
matthewloring: The current behavior is that `TraceWriter.get` will fail if the writer has not been created with a call to `TraceWriter.create`. However, even after create is called, the project id may not be available until it has been resolved here: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/src/trace-writer.js#L99. I would recommend returning the value of `this.config_.projectId`. If this value is truthy, it will be used as the project id for the lifetime of the trace agent and the metadata service will never be consulted. If it is falsey, it will be overwritten with the asynchronously resolved id from the metadata service.

draffensperger: Done. That feels much cleaner.

test/test-trace-api.js
```diff
@@ -180,7 +181,13 @@ describe('Trace Interface', function() {
       traceAPI.runInRootSpan({name: 'root', url: 'root'}, function(rootSpan) {
         var id = traceAPI.getCurrentContextId();
         assert.strictEqual(id, rootSpan.trace.traceId);
-      });      
+      });
+    });
+
+    it('should return get the project ID if set in config', function() {
```
matthewloring: It would be good to have a test that checks that project id becomes available after asynchronous resolution (currently this behavior relies on the trace api and trace writer sharing the same mutable config object which may not always be the case). I don't want to hold up this PR though, maybe you could add a TODO?

draffensperger: Yeah, that's a good point. Just added a test case to `test-index.js` as `index.js` is where the shared `config` assignment takes place.

matthewloring: Hmm, the test looks good but it is failing. I can try to help debug next week.

draffensperger: Turned out I needed to clear the project environment variable.

draffensperger:
No problem. If all is good, can you go ahead and merge this?

PR #553, opened 2017-09-07
test: restore Function#length property in wrapped mocha test functions when using continuation-local-storage with node 8

kjin:
See: #550


ofrobots:
Why was #550 closed? It had a LGTM. What has changed since #550?

kjin:
@ofrobots CircleCI had trouble with #550 because I closed and re-opened that PR. So this PR is the same changes on a different branch. I'll merge this when the CI passes.

PR #554, opened 2017-09-07
Convert the Trace Agent to TypeScript (part 1)

kjin:
This change adds a new `gulp`-based build system for the Trace Agent, and does a superficial refactor of the code to work with this new build system.

I've separated this change into a few commits, each of which should be completely self-contained. I'd recommend looking at each commit separately (almost all new code is in the first commit). Here's a description of what each commit does:

## `build: add typescript build files`

This commit adds `tsconfig.json` and `gulpfile.js`, and amends other files to reflect build system changes.

I've written a new set of gulp tasks to exhaustively cover compilation, as well as all the CI functionality we had before:

* Unit tests (`gulp test`)
* Coverage reporting for unit tests (`gulp coverage`) (TODO(kjin): Verify this in CI)
* System tests (`gulp system-test`)
* Linting (`gulp lint`)

There are a lot of "helper" tasks -- the general rule being that `x.y` is a helper task for `x`.

I also added a `test.single` task, which runs a mocha test in the same process as gulp. This should be useful for debugging.

Notably, there is a `--dev` flag that compilation-related tasks consume. This relaxes `tsconfig` compilation options. As files receive types in the next PR, this flag will go away.

## `refactor: add convert-to-ts`

Adds a script as described below.

## `refactor: run convert-to-ts`

This commit is the result of running the `convert-to-ts` helper module. It's a purely mechanical operation which:

* Renames source and test files to have the `.ts` extension
* Removes `.js` from module requires
* Adds `export default {}` to the bottom of each renamed file (as `tsc` mistakes them for scripts otherwise)

## `src: src changes`

This commit sums up the minimum changes in `src` needed to make `gulp compile --dev` compile source files without any errors.

## `test: test changes`

This commit sums up the minimum changes in `test` needed to make `gulp test --dev` pass all unit tests.

## `revert: remove convert-to-ts`

This commit does exactly what it describes.


package.json
```diff
@@ -23,21 +23,39 @@
...
     "proxyquire": "^1.4.0",
+    "remap-istanbul": "^0.9.5",
     "request": "^2.81.0",
+    "source-map-support": "^0.4.16",
     "standard-version": "^4.2.0",
+    "through2": "^2.0.3",
     "timekeeper": "^1.0.0",
-    "tmp": "0.0.31"
+    "tmp": "0.0.31",
+    "tslint": "^5.7.0",
```
ofrobots: If you use `gts check`, would this dependency still be needed?

gulpfile.js
```diff
@@ -0,0 +1,306 @@
...
+/**
+ * Runs tslint on files in src/, with linting rules defined in tslint.json.
+ */
+gulp.task('lint', 'Emits linting errors found in src/ and test/.', () => {
+  if (util.env.dev) {
+    console.log('Skipping lint as the \'--dev\' option was passed.');
+    return merge2([]);
+  }
+  const program = require('tslint').Linter.createProgram(tsconfigPath);
+  return gulp.src(['src/**/*.ts', 'test/**/*.ts'])
```
ofrobots: This duplicates logic and configuration between `tsconfig.json`. Can we use `gts` for linting and formatting.

gulpfile.js
```diff
@@ -0,0 +1,306 @@
...
+
+gulp.task('format', 'Emits format errors found in src/ and test/.', () => {
+  if (util.env.dev) {
+    console.log('Skipping format as the \'--dev\' option was passed.');
+    return merge2([]);
+  }
+  throw new Error('Not implemented.');
+});
+
+gulp.task('clean', 'Deletes transpiled code.', () => {
```
ofrobots: Use `gts clean`.

gulpfile.js
```diff
@@ -0,0 +1,306 @@
...
+  return del(outDir);
+});
+
+/**
+ * Transpiles TypeScript files in src/ to JavaScript according to the settings found in tsconfig.json.
+ * Currently, all errors are emitted twice. This is being tracked here:
+ * https://github.com/ivogabe/gulp-typescript/issues/438
+ */
+gulp.task('compile', 'Transpiles src/.',
+  makeCompileFn({ transpile: ['*.ts', 'src/**/*.ts'] }));
```
ofrobots: The globs are duplicated from tsconfig. Can we use `gts` for this?

index.ts
```diff
@@ -16,6 +16,14 @@
 
 'use strict';
 
+declare global {
+  namespace NodeJS {
+    export interface Global {
+      _google_trace_agent: any;
```
matthewloring: Will this type be updated to an "agent"?

test/plugins/common.ts
```diff
@@ -15,6 +15,15 @@
  */
 'use strict';
 
+import * as mocha from 'mocha';
+declare global {
```
matthewloring: Do mocha type definitions not provide this?

test/plugins/test-trace-hapi.ts
```diff
@@ -49,7 +49,7 @@ describe('hapi', function() {
   });
 
   Object.keys(versions).forEach(function(version) {
-    if (version.substring(4) > 10 && semver.satisfies(process.version, '<4')) {
+    if (Number(version.substring(4)) > 10 && semver.satisfies(process.version, '<4')) {
```
matthewloring: It might be better to make this a typescript cast instead of a JS one.

kjin:
@ofrobots I'm going to move all build stuff out of the `gulpfile` and into a `scripts/` directory, which will get transpiled along with `src` and `test`. `gts` works for tasks such as lint/format checking and cleaning, but there are still things I would like to see as JS (TS) scripts, such as installing fixtures.

On a semi-related note, `index.ts` and `config.ts` currently reside outside `src`. I'm going to move them in `src`... @ofrobots @matthewloring let me know if you have any objections.

test/plugins/test-trace-hapi.ts
kjin: Typescript casts, AFAIK don't affect the value of the casted object -- so I don't think that would work here.

In `ts-node`, `'1' as number` yields an typecast error.

test/plugins/common.ts
kjin: It does for `it` in the global scope, but now the `it` property on the `global` object. I'm not entirely clear on why there's a distinction, but `tsc` complains if this definition isn't provided.

index.ts
kjin: Yes, in the next part.

tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "extends": "./node_modules/gts/tsconfig-google.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "build",
+    "lib": ["es2015", "es2015.generator"],
```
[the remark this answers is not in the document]
kjin: I found that `es2015.generator` isn't strictly necessary, just an artifact of my experimentation. Removed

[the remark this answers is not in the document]
kjin: `es2015.generator` isn't actually necessary; I've removed it.

gulpfile.js
kjin: Done

gulpfile.js
kjin: Done

gulpfile.js
kjin: Done

package.json
kjin: No, removed

scripts/utils.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const BUILD_DIRECTORY = 'build';
+
+export const globP: (pattern: string) => Promise<string[]> = pify(glob);
+export const ncpP: (src: string, dest: string) => Promise<void> = pify(ncp);
+export const statP: (path: string) => Promise<Stats> = pify(stat);
+export const readFileP: (path: string, encoding?: string) => Promise<Buffer|string> = pify(readFile);
+
+export async function existsP(path: string): Promise<boolean> {
+  let exists = true;
```
matthewloring: Not sure which is cleaner but this could also be something like:

```javascript
return new Promise(res => { fs.access(path, err => { res(!err); }) });
```

test/plugins/test-trace-hapi.ts
matthewloring: Yup, that makes sense.

test/test-cls-ah.ts
```diff
@@ -21,67 +21,50 @@ var semver = require('semver');
 
 if (semver.satisfies(process.version, '<8') || !process.env.GCLOUD_TRACE_NEW_CONTEXT) {
   console.log('Skipping cls-ah tests on node version without async hooks');
-  return;
-}
-
-var cls = require('../src/cls-ah'/*.js*/);
+} else {
```
matthewloring: Why is this else needed?

package.json
```diff
@@ -58,5 +76,75 @@
...
...rvice=github\n[coveralls-url]: https://coveralls.io/github/GoogleCloudPlatform/cloud-trace-nodejs?branch=master\n[david-dev-image]: https://david-dm.org/GoogleCloudPlatform/cloud-trace-nodejs/dev-status.svg\n[david-dev-url]: https://david-dm.org/GoogleCloudPlatform/cloud-trace-nodejs?type=dev\n[david-image]: https://david-dm.org/GoogleCloudPlatform/cloud-trace-nodejs.svg\n[david-url]: https://david-dm.org/GoogleCloudPlatform/cloud-trace-nodejs\n[gcloud-sdk]: https://cloud.google.com/sdk/gcloud/\n[npm-image]: https://badge.fury.io/js/%40google-cloud%2Ftrace-agent.svg\n[npm-url]: https://npmjs.org/package/@google-cloud/trace-agent\n[queueing-problem]: https://github.com/groundwater/nodejs-symposiums/tree/master/2016-02-26-Errors/Round1/UserModeQueuing\n[service-account]: https://console.developers.google.com/apis/credentials/serviceaccountkey\n[snyk-image]: https://snyk.io/test/github/GoogleCloudPlatform/cloud-trace-nodejs/badge.svg\n[snyk-url]: https://snyk.io/test/github/GoogleCloudPlatform/cloud-trace-nodejs\n[travis-image]: https://travis-ci.org/GoogleCloudPlatform/cloud-trace-nodejs.svg?branch=master\n[travis-url]: https://travis-ci.org/GoogleCloudPlatform/cloud-trace-nodejs\n",
```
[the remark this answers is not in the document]
kjin: Um... wow. Not sure how that got there 🙃

test/test-cls-ah.ts
kjin: `tsc` doesn't allow top-level `return` statements.

scripts/utils.ts
kjin: Actually, I can do this which might be cleaner than either:

```js
return statP(path).then(
  () => Promise.resolve(true),
  () => Promise.resolve(false)
);
```

scripts/run-tests.ts
```diff
@@ -0,0 +1,49 @@
+import * as path from 'path';
+import { BUILD_DIRECTORY, globP, forkP } from './utils';
+
+function nodule(nodule: string) {
```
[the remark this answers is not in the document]
kjin: Done

scripts/decrypt-service-account-key.ts
```diff
@@ -0,0 +1,34 @@
...
+  const {
+    encrypted_18363a01ae87_key: key,
+    encrypted_18363a01ae87_iv: iv,
+  } = process.env;
+
+  if (!key || !iv) {
+    throw new Error('Environment insufficient for decrypting service account key');
+  }
+  
+  const path = 'node-team-test-d0b0be11c23d.json';
```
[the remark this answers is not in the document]
kjin: Done

package.json
```diff
@@ -58,5 +76,75 @@
...
+  "readmeFilename": "README.md",
+  "gitHead": "cb00e053c6a13e5e94594a1b3c4684f75506f9e2",
```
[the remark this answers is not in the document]
kjin: Removed

package.json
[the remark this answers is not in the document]
kjin: Removed stuff that was added there somehow... looking into why this happened.

scripts/utils.ts
```diff
@@ -13,14 +13,15 @@ export const ncpP: (src: string, dest: string) => Promise<void> = pify(ncp);
...
-    await statP(path);
-  } catch (e) {
-    exists = false;
-  }
-  return exists;
+export function nodule(nodule: string) {
+  return path.relative(BUILD_DIRECTORY, `node_modules/${nodule}`);
+}
+
+export function existsP(path: string): Promise<boolean> {
```
matthewloring: No need to change this further but I'm curious if `statP(path) as Promise<boolean>` would also work? Seems unfortunate to create additional promises just for type conversion.

kjin: I'm not sure that would work -- we're specifically turning a rejection into a resolution with a `false` value.

matthewloring: Ah yes, that makes sense.

config.ts
```diff
@@ -137,3 +137,5 @@ module.exports = {
       minorVersion: null
     }
 };
+
+export default {};
```
ofrobots: Is this necessary? Style guide forbids default exports.

Here and elsewhere.

index.ts
```diff
@@ -16,6 +16,14 @@
 
 'use strict';
 
+declare global {
+  namespace NodeJS {
```
ofrobots: (disclosure: I may not understand namespaces). Is this necessary? The style guide forbids namespaces.

index.ts
```diff
@@ -26,16 +34,16 @@ if (require('semver').satisfies(process.version, '<8') ||
...
-var pluginLoader = require('./src/trace-plugin-loader.js');
-var TraceWriter = require('./src/trace-writer.js');
+var constants = require('./src/constants'/*.js*/);
+var traceUtil = require('./src/util'/*.js*/);
+var TraceAgent = require('./src/trace-api'/*.js*/);
+var pluginLoader = require('./src/trace-plugin-loader'/*.js*/);
+var TraceWriter = require('./src/trace-writer'/*.js*/);
 
-var modulesLoadedBeforeTrace = [];
+var modulesLoadedBeforeTrace: any[] = [];
```
ofrobots: could this be `string[]`?

package.json
```diff
@@ -2,42 +2,66 @@
   "name": "@google-cloud/trace-agent",
   "version": "2.1.3",
   "description": "Node.js Support for StackDriver Trace",
-  "main": "index.js",
-  "repository": "GoogleCloudPlatform/cloud-trace-nodejs",
+  "main": "build/index.js",
```
ofrobots: Eventually would this become `build/src/index.js`?

package.json
```diff
@@ -2,42 +2,66 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/GoogleCloudPlatform/cloud-trace-nodejs.git"
```
ofrobots: Any reason for this change? The previous terse format looks easier to understand to me.

package.json
```diff
@@ -2,42 +2,66 @@
...
   },
   "keywords": [
     "google",
     "tracing",
     "profiling"
   ],
-  "author": "Google Inc.",
+  "author": {
+    "name": "Google Inc."
+  },
```
ofrobots: Any reason for this change?

scripts/decrypt-service-account-key.ts
```diff
@@ -0,0 +1,34 @@
+import { createDecipheriv } from 'crypto';
+import { createReadStream, createWriteStream } from 'fs';
+
+export default async function() {
+  const {
```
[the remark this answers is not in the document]
ofrobots: 💜

scripts/decrypt-service-account-key.ts
```diff
@@ -0,0 +1,34 @@
...
+  const decipher = createDecipheriv(
+    'aes-256-cbc',
+    Buffer.from(key, 'hex'),
+    Buffer.from(iv, 'hex')
+  );
+
+  const readStream = createReadStream(`${filename}.enc`);
+  const writeStream = createWriteStream(filename);
+
+  await new Promise((resolve, reject) => {
```
ofrobots: nit: returning the promise rather than awaiting would also work, right?

scripts/index.ts
```diff
@@ -0,0 +1,83 @@
+const [bin, script, ...steps] = process.argv;
```
ofrobots: 💛

scripts/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "../node_modules/gts/tsconfig-google.json",
+  "compilerOptions": {
+    "rootDir": "..",
+    "outDir": "../build",
+    "lib": ["es2015"],
+    "target": "es5"
```
ofrobots: nit: `tsconfig-google.json` has the same `lib` and `target` options. Unless you plan to override, you should drop them from config here.

tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "extends": "./node_modules/gts/tsconfig-google.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "build",
+    "lib": ["es2015"],
+    "target": "es2015",
+    "module": "commonjs",
```
ofrobots: nit: I believe these these three lines can be omitted because they are the same as `tsconfig-google.json`?

config.ts
kjin: It's not necessary in this file specifically, but I've found that omitting any sort of export in a file causes the objects declared in that file to be dumped in the global scope rather than module scope upon being imported through `import`. Adding this line is a workaround to prevent that; in a future PR, all of these lines will be removed.

index.ts
kjin: AFAIK Yes - this block augments the `NodeJS` namespace defined in type definitions for Node. The alternative would to put this (and other declarations) in a `d.ts` file -- which I'm planning to do in a future PR.

index.ts
kjin: Done -- I do plan to more comprehensively add types in future PRs

package.json
kjin: Yep, in a future PR

package.json
kjin: Inadvertant from `gts init` - reverted

scripts/decrypt-service-account-key.ts
kjin: Done

scripts/tsconfig.json
kjin: Done

tsconfig.json
kjin: `lib` and `module` are the same -- `target` is actually not since we want to preserve generators for the Koa plugin.

PR #560, opened 2017-09-19
refactor: ts conversion of config, trace-writer, span-data

kjin:
I'd recommend looking at the commits separately.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,0 +1,73 @@
+// TODO(kjin): Unify these definitions with those of the Debugger Agent.
+
+declare module 'gcp-metadata' {
+  import * as http from 'http';
+
+  // TODO: Determine if the signature of the callback on these methods are
+  //       correct.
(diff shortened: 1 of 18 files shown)


src/index.d.ts
```diff
@@ -0,0 +1,73 @@
+// TODO(kjin): Unify these definitions with those of the Debugger Agent.
```
matthewloring: I would prefer to call this file something different since it is not actually related to the module entrypoint.

src/span-data.ts
```diff
@@ -18,19 +18,19 @@
 
 declare global {
   interface CallSite {
```
matthewloring: Could this type definition go in the node type definitions?

src/index.d.ts
kjin: Done

src/span-data.ts
kjin: Done

PR #561, opened 2017-09-19
refactor: ts conversion of trace and span-data

kjin:


src/trace.ts
```diff
@@ -16,16 +16,17 @@
...
-function Trace(projectId, traceId) {
-  this.projectId = projectId;
-  this.traceId = traceId;
-  this.spans = [];
-}
+import { TraceSpan } from './trace-span';
 
-module.exports = Trace;
+export class Trace {
+  public readonly spans: TraceSpan[] = [];
```
matthewloring: Do you have any style thoughts on declaring fields in multiple places (inside constructor + at the top of the class)?

src/trace-span.ts
```diff
@@ -16,51 +16,52 @@
...
- * Sets or updates a label value.
- * @param {string} key The label key to set.
- * @param {string} value The new value of the label.
+ * Describes a trace span.
  */
-TraceSpan.prototype.setLabel = function(key, value) {
-  this.labels[key] = value;
-};
+export class TraceSpan {
+  public readonly labels: TraceSpanLabels = {};
```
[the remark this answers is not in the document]
kjin: I believe class members cannot have `const` as a modifier.

src/trace.ts
kjin: It's one or the other, declaring in both places is a TS error.

Placing in the constructor is a style guide preference: https://g3doc.corp.google.com/javascript/typescript/g3doc/styleguide.md?cl=head#class-members

src/trace-span.ts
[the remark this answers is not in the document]
matthewloring: Ah ok.

src/trace.ts
matthewloring: Sorry, I meant declaring some of a classes fields at the top and others in the constructor. It seems like it would be nice to have one place to look to see what state is maintained by a class.

kjin: I agree that it's nicer to have it all in one place; the style guide is not specific on whether fields should be split. Maybe we can re-organize all classes as a followup.

PR #562, opened 2017-09-20
refactor: ts conversion of util

kjin:


test/plugins/common.ts
```diff
@@ -196,6 +196,15 @@ function hasContext() {
   return !!cls.getRootContext();
 }
 
+function notNull<T>(arg: T | null) : T {
+  assert.notStrictEqual(arg, null);
```
matthewloring: This seems redundant with the check below.

kjin: Yep, the below error will never be thrown. It's only there for type safety; I'll add a comment to describe this.

FYI, I think we should prefer an error thrown by the `assert` module, which is why I didn't bother elaborating on the error object thrown below.

PR #567, opened 2017-09-26
refactor: ts conversion of trace-agent

kjin:
This change converts `trace-agent.ts` to TypeScript.

A caveat -- this change removes switching between `TraceAgent.prototype` and `phantomApiImplementation`, instead opting for having all member functions on the `TraceAgent` class return default values if the Trace Agent is disabled. The reason for this is that ES6 class prototype members aren't iterable in the same way as before. While there might be a workaround for this, I figured that I might as well remove the switching mechanism altogether (it could be considered bad style).


src/trace-api.ts
```diff
@@ -16,296 +16,348 @@
...
-TraceAgent.prototype.getResponseTraceContext = function(
-    incomingTraceContext, isTraced) {
-  var traceContext = util.parseContextFromHeader(incomingTraceContext);
-  if (!traceContext) {
-    return '';
-  }
-  traceContext.options = (traceContext.options || 0) & isTraced;
-  return util.generateTraceContext(traceContext);
-};
+    const rootSpan: cls.RootContext = cls.getRootContext();
```
ofrobots: is this type definition inferrable?

src/trace-span.ts
```diff
@@ -36,7 +36,7 @@ export class TraceSpan {
   constructor(
     public readonly name: string,
     public readonly spanId: number,
-    public readonly parentSpanId: number
+    public readonly parentSpanId: number | string
```
ofrobots: From the rest of the changes, I cannot see why this definition needed to change. What's the reason here?

kjin: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/567/files#diff-4fae65ee00e417727e40216b35f1cbe7R221

Incoming span ID is a UInt64, so AFAIK we need it to be a string to represent it exactly.

src/trace-api.ts
```diff
@@ -16,296 +16,348 @@
...
+  };
+
+  /**
+   * Gets the value of enhancedDatabaseReporting in the trace agent's
+   * configuration object.
+   * @returns A boolean value indicating whether the trace agent was configured
+   * to have an enhanced level of reporting enabled.
+   */
+  enhancedDatabaseReportingEnabled(): boolean {
+    if (this.config_) {
```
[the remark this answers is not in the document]
kjin: Done

src/trace-api.ts
```diff
@@ -16,296 +16,348 @@
...
-  // TODO: When converting this file to TS, remove this cast.
-  return (rootSpan as SpanData).trace.traceId;
-};
+    return namespace.runAndReturn(() => {
+      // Attempt to read incoming trace context.
+      let incomingTraceContext: {
+        traceId?: string,
+        spanId?: string,
+        options?: number
+      } = {};
```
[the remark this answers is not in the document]
kjin: Done

src/trace-api.ts
kjin: Yep, done

src/trace-span.ts
ofrobots: Would it make sense for it to be only a `string` then? AFAIK we don't need to do arithmetic on it.

kjin: Yep -- but to keep things consistent I'll do likewise for other references to span IDs.

PR #568, opened 2017-09-26
refactor: ts conversion of plugin loader

kjin:
This change turns `trace-plugin-loader.ts` into TS.

### DIFF
--- a/src/trace-plugin-loader.ts
+++ b/src/trace-plugin-loader.ts
@@ -81,7 +81,9 @@ export interface Intercept {
   intercept: <T>(module: T, agent: TraceAgent) => T;
 }
 
-export type Plugin = (Patch | Intercept)[];
+export type Instrumentation = Patch | Intercept;
+
+export type Plugin = Array<Instrumentation>;
 
 // type guards
 
--- a/src/trace-plugin-loader.ts
+++ b/src/trace-plugin-loader.ts
@@ -15,26 +15,106 @@
  */
 'use strict';
 
+// TODO(kjin)
+// Module augmentation is implemented in this PR:
+// https://github.com/DefinitelyTyped/DefinitelyTyped/pull/19612
+// Do this correctly when it's landed and released.
+import _Module = require('module');
+const Module: {
+  _resolveFilename(request: string, parent?: NodeModule): string;
+  _load(request: string, parent?: NodeModule, isMain?: boolean): any;
+} = _Module as any;
+
+import { Logger } from '@google-cloud/common';
+import * as path from 'path';
+import * as semver from 'semver';
+import * as shimmer from 'shimmer';
 import * as util from './util';
-import { TraceAgent } from './trace-api';
+import { TraceAgent, TraceAgentConfig } from './trace-api';
 
-var Module = require('module');
-var shimmer = require('shimmer');
-var path = require('path');
-var semver = require('semver');
+/**
+ * An interface representing config options read by the plugin loader, which includes
+ * TraceAgent configuration as well.
+ */
+export interface PluginLoaderConfig extends TraceAgentConfig {
+  plugins: {
+    [pluginName: string]: string;
+  };
+}
+interface InternalPatch<T> {
+  file: string;
+  module?: T;
+  patch: (module: T, agent: TraceAgent) => void;
+  unpatch?: (module: T) => void;
+}
 
-var plugins = Object.create(null);
-var intercepts = Object.create(null);
-var activated = false;
+interface InternalIntercept<T> {
+  file: string;
+  module?: T;
+  intercept: (module: T, agent: TraceAgent) => T;
+}
 
-var logger_;
+interface InternalPlugin {
+  file: string;
+  patches: {
+    [patchName: string]: {
+      [file: string]: InternalPatch<any> | InternalIntercept<any>;
+    }
+  };
+  agent: TraceAgent;
+}
 
-function checkLoadedModules() {
-  for (var moduleName in plugins) {
+interface PluginStore {
+  [pluginName: string]: InternalPlugin;
+}
+
+export interface Patch {
+  file?: string;
+  versions?: string;
+  patch: (module: any, agent: TraceAgent) => void;
+  unpatch?: (module: any) => void;
+}
+
+export interface Intercept {
+  file?: string;
+  versions?: string;
+  intercept: <T>(module: T, agent: TraceAgent) => T;
+}
+
+export type Plugin = (Patch | Intercept)[];
+
+// type guards
+
+function isPatch(obj: Patch | Intercept): obj is Patch {
+  return !!(obj as Patch).patch;
+}
+
+function isIntercept(obj: Patch | Intercept): obj is Intercept {
+  return !!(obj as Intercept).intercept;
+}
+
+function isInternalPatch<T>(
+    obj: InternalPatch<T> | InternalIntercept<T>): obj is InternalPatch<T> {
+  return !!(obj as InternalPatch<T>).patch;
+}
+
+function isInternalIntercept<T>(
+    obj: InternalPatch<T> | InternalIntercept<T>): obj is InternalIntercept<T> {
+  return !!(obj as InternalIntercept<T>).intercept;
+}
+
+let plugins: PluginStore = Object.create(null);
+let intercepts = Object.create(null);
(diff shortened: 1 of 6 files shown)


src/trace-plugin-loader.ts
```diff
@@ -15,26 +15,113 @@
...
+  };
+  agent: TraceAgent;
+}
+
+interface PluginStore {
+  [pluginName: string]: InternalPlugin;
+}
+
+export interface Patch {
+  file?: string;
```
ofrobots: nit: `file` is not optional in `InternalPatch`. Is this intentional?

src/trace-plugin-loader.ts
```diff
@@ -15,26 +15,113 @@
...
+  unpatch?: (module: any) => void;
+}
+
+export interface Intercept {
+  file?: string;
+  versions?: string;
+  intercept: <T>(module: T, agent: TraceAgent) => T;
+}
+
+export type Plugin = (Patch | Intercept)[];
```
ofrobots: style guide prefers `Array<>` syntax.

ofrobots: it might make sense to give a concrete type name to `Patch|Intercept`.

src/trace-plugin-loader.ts
```diff
@@ -43,15 +130,15 @@ function checkLoadedModules() {
...
+    const first = process._preload_modules[0];
     if (first !== '@google-cloud/trace-agent') {
       logger_.error('Tracing might not work as ' + first +
             ' was loaded with --require before the trace agent was initialized.');
     }
   }
 }
 
-function checkPatch(patch) {
+function checkPatch(patch: any) {
```
ofrobots: Can we avoid `any`?

src/trace-plugin-loader.ts
kjin: Yes, omitting `file` is equivalent to specifying that the export of the module itself should be patched (from the [plugin guide](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/doc/plugin-guide.md))

src/trace-plugin-loader.ts
kjin: We can avoid it by explicitly setting its type to `{ file?: any, patch?: any, unpatch?: any, intercept?: any }`, or something more complex. I don't think it's worthwhile to set the type more precisely than `any`, given that this function is doing type-checking of some sort on its own.

src/trace-plugin-loader.ts
kjin: `Array<>` syntax done.

Thoughts on `type Instrumentation = Patch | Intercept`? I realize the term is overloaded in the current file, but the variables named `instrumentation` are all internal and can be renamed.

src/trace-plugin-loader.ts
ofrobots: My concern was around the fact that for internal patches we require that the `file` must always be specified. If we already have the code (and tests) to deal with optional `file` field, do we really get anything out of mandating `file` for internal patches?

ofrobots: Put differently, I want to make sure there is value we get out of having different types for `InternalPatch` and `Patch`.

kjin: It's true that data store in an `InternalPatch` is derived mostly from `Patch`, but we already want to hold other information, such as `module`, a pointer to the loaded module being patched. Since this other information is purely implementation detail, it wouldn't be appropriate to expose it to the user (`Patch` is a user-facing interface).

Given that there must be a distinction between `InternalPatch` and `Patch`, I don't see why we should make `InternalPatch::file` optional. In my mind, any internal interface should be as strictly-defined as possible, while anything user-facing should be as loose as needed -- this is a principle that we already apply to the agent configuration object in `index.ts::initConfig`: to the user, every configuration field appears to be optional, but internally, we copy the user-set fields into an object with a more strictly-defined shape, substituting default values for user-omitted fields.

ofrobots: The one reason I can think of is that in the current structure they are unrelated interfaces. I would prefer one to be a subtype of the other. That leads to more code sharing.

kjin: Between internal and external representations of some data:

1. The external representation should be at least more permissive than the internal representation, allowing users to omit fields.  The implementation will populate the internal representation with default values (`InternalPatch::file`)
2. The internal representation may have extra fields that are not relevant in the external representation (`InternalPatch::module`)
3. One should be a subtype of the other, or they should otherwise share common code in some way

My belief is that these points are generally applicable, beyond just interface definitions. For example, the first two points apply to functions as well: functions accept optional arguments, initialize them to default values if omitted, and also have function-scope "implementation detail" variables.

These three things are all nice to have (feel free to dispute that though), but AFAIK in TypeScript we can only pick two, because in (1) the internal representation is a subtype of the external representation, and vice-versa in (2). Given this, I would prefer to have the first two despite the slightly improved code brevity provided by (3), because the first two have to do with code correctness, and the third code readability, which is arguably less important.

Maybe a different way of putting this is -- between a public-facing and implementation-specific representation, which should be a subtype of the other?

ofrobots: Can't we get all three? This is valid TS.

```ts
interface Patch {
  file?: string;
}

interface InternalPatch extends Patch {
  file: string;
}

function foo(p: Patch) {}

const q: InternalPatch = { file: 'foo' };

foo(q);
```

Reducing code duplication doesn't just give us brevity, it is also leads to fewer opportunities for bugs or for code to incidentally diverge.

kjin: Oops, you're right

src/trace-plugin-loader.ts
```diff
@@ -15,26 +15,109 @@
...
+}
+
+export type Instrumentation<T> = Patch<T> | Intercept<T>;
+
+export type Plugin = Array<Instrumentation<any>>;
+
+interface InternalPatch<T> extends Patch<T> {
+  file: string;
+  module?: T;
+  patch: (module: T, agent: TraceAgent) => void;
```
ofrobots: can patch & unpatch definitions be omitted? The should get inherited from Patch.

src/trace-plugin-loader.ts
```diff
@@ -15,26 +15,109 @@
...
+  unpatch?: (module: T) => void;
+}
+
+interface InternalIntercept<T> extends Intercept<T> {
+  file: string;
+  module?: T;
+  intercept: (module: T, agent: TraceAgent) => T;
```
ofrobots: likewise.

src/trace-plugin-loader.ts
```diff
@@ -15,26 +15,109 @@
...
+  [pluginName: string]: InternalPlugin;
+}
 
-var plugins = Object.create(null);
-var intercepts = Object.create(null);
-var activated = false;
+// type guards
 
-var logger_;
+function isPatch<T>(obj: Patch<T> | Intercept<T>): obj is Patch<T> {
```
ofrobots: `obj: Instrumentation<T>`

src/trace-plugin-loader.ts
```diff
@@ -15,26 +15,109 @@
...
 
-var Module = require('module');
-var shimmer = require('shimmer');
-var path = require('path');
-var semver = require('semver');
+interface InternalPlugin {
+  file: string;
+  patches: {
+    [patchName: string]: {
+      [file: string]: InternalPatch<any> | InternalIntercept<any>;
```
ofrobots: Maybe create an alias as InternalInstrumentation?

src/trace-plugin-loader.ts
ofrobots: Could we do better than `any` now?

src/trace-plugin-loader.ts
```diff
@@ -76,12 +159,12 @@ function activate(logger, config) {
...
 
-  var pluginConfig = config.plugins;
-  for (var moduleName in pluginConfig) {
+  const pluginConfig = config.plugins;
+  for (const moduleName in pluginConfig) {
     if (!pluginConfig[moduleName]) {
       continue;
     }
-    var agent = new TraceAgent(moduleName);
+    const agent: TraceAgent = new TraceAgent(moduleName);
```
ofrobots: nit: type can be inferred.

PR #569, opened 2017-10-03
refactor: ts conversion of index.ts

kjin:
In addition to converting index.ts to use ES6 features and types, this also renames `TraceWriterOptions` to `TraceWriterConfig`.


src/types.d.ts
```diff
@@ -76,6 +76,10 @@ declare module '@google-cloud/common' {
   
   export function logger(options?: LoggerOptions | string): Logger;
 
+  export namespace logger {
```
ofrobots: can we avoid namespaces?

src/types.d.ts
```diff
@@ -106,6 +110,19 @@ declare module '@google-cloud/common' {
   }
 }
 
+// There is a @types/extend, but it's not expressive enough.
```
ofrobots: Can we submit this as a PR?

kjin: Yep -- but while our attention is focused on it, what do you think of dropping `extend` for `Object.assign`? Since we already target Node 4+, which is the criteria for using the latter.

ofrobots: sgtm.

src/types.d.ts
kjin: Done; almost removed in `shimmer`.

src/types.d.ts
kjin: Done -- changed to reference the PR (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/20266)

kjin: I'll remove the `extend` dependency in a future PR.

index.ts
```diff
@@ -69,13 +79,13 @@ function initConfig(projectConfig) {
...
-  var envSetConfig = {};
+  let envSetConfig: Config = {};
   if (process.env.hasOwnProperty('GCLOUD_TRACE_CONFIG')) {
-    envSetConfig = require(path.resolve(process.env.GCLOUD_TRACE_CONFIG));
+    envSetConfig = require(path.resolve(process.env.GCLOUD_TRACE_CONFIG)) as Config;
   }
   // Configuration order of precedence:
   // Default < Environment Variable Set Configuration File < Project
-  var config = extend(true, {}, defaultConfig, envSetConfig,
+  const config = extend(true, { forceNewAgent_: false }, defaultConfig, envSetConfig,
```
ofrobots: Why was this `forceNewAgent_` necessary? Is it not in `defaultConfig`?

While you're here can you also edit the comment above to add the missing `envConfig` to the precedence list.

src/plugin-interfaces.ts
```diff
@@ -0,0 +1,163 @@
+import { Constants } from './constants';
```
ofrobots: copyright

src/plugin-interfaces.ts
```diff
@@ -0,0 +1,163 @@
...
+import { TraceLabels } from './trace-labels';
+
+export type Func<T> = (...args: any[]) => T;
```
ofrobots: nit: use a more descriptive name perhaps.

src/plugin-interfaces.ts
```diff
@@ -0,0 +1,163 @@
...
+
+export interface Intercept<T> {
+  file?: string;
+  versions?: string;
+  intercept: (module: T, agent: TraceAgent) => T;
+}
+
+export type Instrumentation<T> = Patch<T> | Intercept<T>;
+
+export type Plugin = Array<Instrumentation<any>>;
```
ofrobots: nit: newline

src/span-data.ts
```diff
@@ -45,7 +46,7 @@ interface StackFrame {
 // Auto-incrementing integer
 let uid = 1;
 
-export class SpanData {
+export class SpanData implements ISpanData {
```
ofrobots: See 'Naming Style' section of the style guide on `I` prefix for interfaces. Disallowed.

Also two different files exporting the same type name is not great. Use different names.

src/trace-api.ts
```diff
@@ -20,6 +20,7 @@ import { Logger } from '@google-cloud/common';
 import * as cls from './cls';
 import { Constants } from './constants';
 import * as is from 'is';
+import { Func, SpanOptions, RootSpanOptions, TraceAgent as ITraceAgent } from './plugin-interfaces';
```
ofrobots: same comment about `ITraceAgent`

src/plugin-interfaces.ts
kjin: Not sure if there is anything more specific than `Func` (or `Function`) for something this broad.

src/span-data.ts
kjin: The externally facing types in `plugin-interface` should take precedence when it comes to name brevity; I don't want to call them `TraceAgentInterface`, etc.

I could follow in the footsteps of `grpc` and name the classes `*Impl*, but I'd want to do that as a separate change. For now, I hope it's acceptable to have `SpanDataInterface`?

src/plugin-interfaces.ts
kjin: done

index.ts
kjin: It's not set in `defaultConfig`. As it's test-only, I don't think we ever want it to be exposed.

Comment edited!

Issue #574, opened 2017-10-17
Improve docs on ignoreUrls option


amsibamsi:
Found it confusing to know how the ignoreUrls option exactly works. From my limited understanding of the source it does do string matching, so specifying '/' will actually ignore all URLs. So exact matches need to be regular expressions, like /^\/$/. Can you please verify if this is how it works and then maybe add more detailed docs?

kjin:
Yes -- each input string in `ignoreUrls` is treated as a regex against which the potentially ignored URLs will be matched, so '/' will end up ignoring everything. (In addition -- the URL is _also_ ignored if it's an exact match of an `ignoreUrls` element -- something we will remove as a semver-major change in the future.)

I'll change the docs to make it more reflective of its function. Thanks for bringing this to our attention!

kjin:
I've updated the contents of `config.ts` to better describe `ignoreUrls`.

amsibamsi:
Thanks a lot!

PR #575, opened 2017-10-17
docs: update config.ts

kjin:
Various changes to in-code documentation to hopefully make things clearer. `ignoreUrls` is the important one.

Addresses #574

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -55,7 +55,7 @@ export interface Config {
 
   // URLs that partially match any regex in ignoreUrls will not be traced.
   // In addition, URLs that are _exact matches_ of strings in ignoreUrls will
-  // also not be traced.
+  // also not be traced (this is deprecated behavior and will be removed in v3).
   // URLs should be expected to be in the form of:
   //   /componentOne/componentTwo...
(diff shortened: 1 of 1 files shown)


config.ts
```diff
@@ -16,52 +16,52 @@
...
-  // If paths are present in this array, then these paths will be ignored before
-  // `samplingRate` based decisions are made. Paths must include a leading
-  // forward slash and be of the form:
-  //   /componentOne/componentTwo/...
-  // Paths can additionally be classified by regex in which case any path matching
-  // any provided regex will be ignored.
-  // We ignore the health checker probes (/_ah/health) by default.
-  ignoreUrls?: string[];
+  // URLs that partially match any regex in ignoreUrls will not be traced.
+  // In addition, URLs that are _exact matches_ of strings in ignoreUrls will
```
ofrobots: Given the comment here: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/issues/574#issuecomment-337298135, you expect the exact match behaviorr to be removed in a semver major change in the future. If that's the case, it would be could to call this out as deprecated behavior so that users can avoid depending upon it.

kjin: Good call, done

PR #576, opened 2017-10-18
style: start lint/format for src files

kjin:
This change applies `gts` rules for linting, formatting, and tsc compiler options for source files (not including plugins).

Note on the several tsconfig files:

We want files matching `*.ts` and `src/*.ts` to match the TS style guide, but files that are not fully converted (plugins and tests) should be excluded. To accomplish this I have a `tsconfig.full.json` which compiles _all_ files with non-strict rules, and `tsconfig.json` to additionally compile source files under strict rules. Eventually, `tsconfig.full.json` will disappear.

This works well with `gts check` because `gts` will read only `tsconfig.json` to get a list of files to lint and format.


scripts/index.ts
```diff
@@ -16,7 +16,7 @@ async function run(steps: string[]) {
         'npm',
         [
           'run',
-          ...moduleAndArgs.slice(1)
+          moduleAndArgs.slice(1).join('-')
```
ofrobots: This seems like a logical rather than style change. Intentional?

src/trace-api.ts
```diff
@@ -75,36 +76,37 @@ export class TraceAgent implements TraceAgentInterface {
   public readonly constants = Constants;
   public readonly labels = TraceLabels;
 
-  private pluginName_: string;
-  private logger_: Logger;
-  private config_: TraceAgentConfig;
+  private pluginName: string;
```
ofrobots: This is not semver major?

tsconfig.json
```diff
@@ -1,18 +1,13 @@
 {
-  "extends": "./node_modules/gts/tsconfig-google.json",
+  "extends": "./tsconfig.full.json",
```
ofrobots: is tsconfig.full.json different from tsconfig-google.json? If so, what's the rationale here?

tsconfig.full.json
```diff
@@ -0,0 +1,18 @@
...
+    "noImplicitAny": false,
+    "noImplicitThis": false,
+    "noEmitOnError": false
+  },
+  "include": [
+    "*.ts",
+    "src/**/*.ts",
+    "test/**/*.ts"
+  ]
+}
```
ofrobots: new line

tsconfig.json
```diff
@@ -1,18 +1,13 @@
...
+    "noImplicitThis": true,
+    "noEmitOnError": true
   },
   "include": [
     "*.ts",
-    "src/**/*.ts",
-    "test/**/*.ts"
+    "src/*.ts"
   ]
 }
```
ofrobots: add new line while you're here.

scripts/index.ts
kjin: This is a build script -- wanted to support npm commands that have a hyphen in its name.

src/trace-api.ts
kjin: `pluginName_` is an undocumented field. I would not consider this semver-major.

tsconfig.json
kjin: Edited abstract above. Hope it makes sense

tsconfig.full.json
kjin: Done

tsconfig.json
kjin: Done

PR #578, opened 2017-10-19
test: cleanups and ensure module installs correctly

kjin:
In commit order:
1. Make build script transpile on the fly with `ts-node`, and restrict the set of files being packaged.
1. Remove local copy of `extend` types in favor of `@types/extend@3.0.0`.
1. Use `/** ... */` instead of `// ...` in `config.ts` because `tsc` erases the latter when producing type definitions.
1. Remove `'use strict'` in source files because `tsc` automatically adds them.
1. Re-order imports (as presumably enforced by clang-format 1.1.0).
1. Add a new "check install" test. This test essentially installs a packed form of this module in a temporary directory and tries to run `tsc` there. This ensures that issues don't arise during installation or compilation for users of this module. This is being tested on AppVeyor and Travis now.
1. CI fails on the above commit because of missing dependent type definitions used only in internal files, so further restrict the set of type definitions being packaged and change `index.ts` to reference the `TraceAgent` interface rather than its implementation.


kjin:

config.ts
```diff
@@ -16,111 +16,145 @@
 
 import * as path from 'path';
 
-// Default configuration
+/**
+ * Available configuration options.
+ */
```
jinwoo: Nit. I'd prefer a one-liner for short comments like this. Like
```typescript
/** Available configuration options. */
```

config.ts
```diff
@@ -16,111 +16,145 @@
...
-  //            responsible for terminating the application.
-  //   `flushAndExit`: Handle the uncaught exception, make a best effort attempt
-  //            to publish the traces to the API, and then terminate the
-  //            application after a delay. Note that the presence of other
-  //            uncaught exception handlers may choose to terminate the
-  //            application before the buffer has been flushed to the network.
+  /**
+   * Specifies the behavior of the trace agent in the case of an uncaught
+   * exception. Possible values are:
+   *   `ignore` Take no action. The process may terminate before all the traces
```
jinwoo: Missing colon after \`ignore\`.

scripts/check-install.ts
```diff
@@ -0,0 +1,28 @@
...
+ * 1. In a temporary directory, installs from the `npm pack` of this directory
+ * 2. Compiles a top-level file in that directory that imports this module
+ */
+export default async function() {
+  // This script assumes that you don't already have a TGZ file
+  // in your current working directory.
+  const installDir = await tmpDirP();
+  console.log(installDir);
+  await spawnP('npm', ['pack']);
+  const tgz = (await globP(`${process.cwd()}/*.tgz`)).join(' '); // should be only one
```
jinwoo: If there should be only one .tgz file, why not programmatically check it?

scripts/check-install.ts
```diff
@@ -0,0 +1,28 @@
...
+  await spawnP('npm', ['install', 'typescript', '@types/node', tgz], {
+    cwd: installDir
+  });
+  // use-module.ts is a fixture that imports the Trace Agent
+  await ncpP('./test/fixtures/use-module.ts', `${installDir}/index.ts`);
+  await spawnP(`node_modules${path.sep}.bin${path.sep}tsc`, ['index.ts'], {
+    cwd: installDir
+  });
+  console.log('`npm install` + `tsc` test was successful.');
+}
```
jinwoo: Nit. Missing newline.

scripts/index.ts
```diff
@@ -22,6 +23,9 @@ async function run(steps: string[]) {
       continue;
     }
     switch (step) {
+      case 'check-install':
+        await checkInstall();
```
jinwoo: Looks like all these `await`s in this `switch` statement are just "fire-and-forget" things. Do we need `await`?

scripts/check-install.ts
kjin: Done

scripts/check-install.ts
kjin: Makes sense, added throwing an error.

config.ts
kjin: Done

config.ts
kjin: Done

scripts/index.ts
kjin: In this place we need `await`, since this switch statement takes place in a for loop iterating over "step" names. It's true that not every usage of `await` is needed, but for consistency's sake I prefer using it over *Sync operations.

scripts/check-install.ts
```diff
@@ -12,11 +12,14 @@ export default async function() {
   const installDir = await tmpDirP();
   console.log(installDir);
   await spawnP('npm', ['pack']);
-  const tgz = (await globP(`${process.cwd()}/*.tgz`)).join(' '); // should be only one
+  const tgz = (await globP(`${process.cwd()}/*.tgz`));
```
jinwoo: Delete the outer parens?

scripts/index.ts
jinwoo: Sorry I missed the for loop.

scripts/check-install.ts
kjin: Done

PR #581, opened 2017-10-23
chore: update dep: gcp-metadata

ofrobots:
Fixes slow environment detection when running off of GCP. See
https://github.com/stephenplusplus/gcp-metadata/issues/10.


ofrobots:
Superceded by #590.

PR #586, opened 2017-10-30
fix: extend supported gRPC version range to 1.7

### LINKED ISSUE #585: Trace Agent won't work with gRPC 1.7
The Trace Agent will crash when trying to start with gRPC 1.7 (not the pre-release). We're hoping to get a fix out before it's released.

cc @murgatroid99 (FYI)

kjin:
Fixes #585


src/plugins/plugin-grpc.ts
```diff
@@ -469,26 +469,44 @@ function unpatchServer(server) {
...
+    patch: patchClient,
+    unpatch: unpatchClient
+  },
+  {
+    file: 'src/metadata.js',
+    versions: '1.7',
+    patch: patchMetadata,
+    // patchMetadata doesn't modify the module exports of metadata.js.
+    // So it's safe to have provide a no-op unpatch function.
+    unpatch: function unpatchMetadata() {}
```
jinwoo: This duplicates the `unpatchMetadata()` function. Maybe define one at the top-level and use it from both places? Or use an arrow function instead? `unpatch: () => {}`

kjin: Good catch -- removed duplicate comment as well

kjin:
All CI passed at 474fb76... Merging since #585 is blocking gRPC 1.7.

PR #590, opened 2017-11-06
chore: upgrade gcp-metadata to 0.4

kjin:
Supercedes #581

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ env:
   - CXX=g++-4.8 GCLOUD_TRACE_NEW_CONTEXT=1
 
 script:
+  - if [ $TRAVIS_NODE_VERSION -ge 8 ]; then yarn list; fi
   - npm run check-install && npm run coverage && npm run system-test && npm run non-interference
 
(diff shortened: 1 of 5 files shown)


kjin:
AppVeyor is green: https://ci.appveyor.com/project/matthewloring/cloud-trace-nodejs/build/1.0.1839

I believe the check didn't show up because NuGet was down for a large portion of the day today.

.travis.yml
```diff
@@ -22,6 +22,7 @@ env:
   - CXX=g++-4.8 GCLOUD_TRACE_NEW_CONTEXT=1
 
 script:
+  - if [ $TRAVIS_NODE_VERSION -ge 8 ]; then yarn list; fi
```
ofrobots: Why is this needed?

kjin: We don't print package structure for Node 8 tests so I wanted to add that for debuggability -- but I realized that we don't for any Node version tests.

I'll remove the conditional in an upcoming PR... don't want to wait for CI to pass again before landing this change.

PR #591, opened 2017-11-07
src: patch https in Node 8.9 and 9.0

### LINKED ISSUE #589: https.request doesn't work in Node 8.9+
Context: https://github.com/nodejs/node/pull/16395

kjin:
Fixes #589 

Loading either `http` or `https` will cause both to be patched for tracing. This is in line with behavior in <8.9, as `https.request` used to depend on `http.request`.

The only new user-facing change is in the addition of the `https` config field: disabling causes the specific behavior of not patching `https` only if it's loaded before `http` in user-space code. It's a niche use-case so I think this is justifiable, but please let me know your thoughts.

By nature of our loader system this is uncharted territory, so feel free to give suggestions if you think there's a better approach.

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -164,6 +164,7 @@ export const defaultConfig = {
     'grpc': path.join(__dirname, 'src/plugins/plugin-grpc.js'),
     'hapi': path.join(__dirname, 'src/plugins/plugin-hapi.js'),
     'http': path.join(__dirname, 'src/plugins/plugin-http.js'),
+    'https': path.join(__dirname, 'src/plugins/plugin-https.js'),
     'knex': path.join(__dirname, 'src/plugins/plugin-knex.js'),
     'koa': path.join(__dirname, 'src/plugins/plugin-koa.js'),
     'mongodb-core': path.join(__dirname, 'src/plugins/plugin-mongodb-core.js'),
--- a/src/plugins/plugin-http.ts
+++ b/src/plugins/plugin-http.ts
@@ -174,7 +174,7 @@ module.exports = [
   },
   {
     file: 'https',
-    versions: '>=8.9.0',
+    versions: '=8.9.0 || >=9.0.0',
     patch: patchHttps,
     unpatch: unpatchHttps
   }
--- a/src/plugins/plugin-http.ts
+++ b/src/plugins/plugin-http.ts
@@ -44,16 +44,94 @@ function extractUrl(options) {
     (options.path || options.pathName || '/');
 }
 
-function unpatchHttp(http) {
-  shimmer.unwrap(http, 'request');
-  if (semver.satisfies(process.version, '>=8.0.0')) {
-    shimmer.unwrap(http, 'get');
-  }
+function isTraceAgentRequest (options, api) {
+  return options && options.headers &&
+    !!options.headers[api.constants.TRACE_AGENT_REQUEST_HEADER];
+}
+
+function makeRequestTrace(request, api) {
+  // On Node 8+ we use the following function to patch both request and get.
+  // Here `request` may also happen to be `get`.
+  return function request_trace(options, callback) {
+    if (!options) {
+      return request.apply(this, arguments);
+    }
+
+    // Don't trace ourselves lest we get into infinite loops
+    // Note: this would not be a problem if we guarantee buffering
+    // of trace api calls. If there is no buffering then each trace is
+    // an http call which will get a trace which will be an http call
+    if (isTraceAgentRequest(options, api)) {
+      return request.apply(this, arguments);
+    }
+
+    options = isString(options) ? url.parse(options) : merge({}, options);
+    options.headers = options.headers || {};
+
+    var uri = extractUrl(options);
+    var requestLifecycleSpan =
+        api.createChildSpan({name: getSpanName(options)});
+    if (!requestLifecycleSpan) {
+      return request.apply(this, arguments);
+    }
+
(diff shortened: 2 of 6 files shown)


config.ts
```diff
@@ -164,6 +164,7 @@ export const defaultConfig = {
     'grpc': path.join(__dirname, 'src/plugins/plugin-grpc.js'),
     'hapi': path.join(__dirname, 'src/plugins/plugin-hapi.js'),
     'http': path.join(__dirname, 'src/plugins/plugin-http.js'),
+    'https': path.join(__dirname, 'src/plugins/plugin-https.js'),
```
ofrobots: If I require `https`, that will recursively require `http` hitting the existing plugin. Would that not work?

kjin: That won't be the case -- the order of events happens like this:

1. call `require('https')`
1. load plugin for `https`
1. load `https`
    1. call `require('http')`
    1. load plugin for `http`
    1. load `http`
    1. patch `http` (and `https`)
1. patch `https` (no op)

We can't patch `https` in (3iv) because it's already been loaded.

So what I'm proposing here is for (3i) thru (3iv) to be moved under (2) rather than (3).

kjin:
AppVeyor CI: https://ci.appveyor.com/project/matthewloring/cloud-trace-nodejs/build/1.0.1848

kjin:
@ofrobots @GoogleCloudPlatform/node-team PTAL when AppVeyor CI is green.

src/plugins/plugin-https.ts
```diff
@@ -0,0 +1,25 @@
...
+
+'use strict';
+
+// Force load http to patch https with the http plugin.
+// https depends on http anyway, so this shouldn't cause an unnecessary load.
+require('http');
+
+module.exports = [];
+
+export default {};
```
ofrobots: default exports are not permitted (gts should catch this)

src/plugins/plugin-https.ts
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+'use strict';
+
+// Force load http to patch https with the http plugin.
+// https depends on http anyway, so this shouldn't cause an unnecessary load.
+require('http');
+
+module.exports = [];
```
ofrobots: It doesn't make sense to use both CJS style exports and ES6 exports in the same file. Which one takes precedence?

src/plugins/plugin-http.ts
```diff
@@ -44,118 +44,139 @@ function extractUrl(options) {
...
-  shimmer.unwrap(http, 'request');
-  if (semver.satisfies(process.version, '>=8.0.0')) {
-    shimmer.unwrap(http, 'get');
-  }
+function isTraceAgentRequest (options, api) {
+  return options && options.headers &&
+    !!options.headers[api.constants.TRACE_AGENT_REQUEST_HEADER];
+}
+
+function makeRequestTrace(request, api) {
```
ofrobots: This code reorganization makes it harder to review. Is anything changed in this file?

kjin: `makeRequestTrace` and `isTraceAgentRequest` are unchanged except they now each take in an `api` object. They were taken out of `patchHttp` so they can be shared with `patchHttps`.

src/plugins/plugin-https.ts
kjin: Changing this to `export = []`.

src/plugins/plugin-https.ts
kjin: Plugins aren't covered by gts -- as they have not been modified (save for the default export to mark the file as a module rather than a script).

That being said, since this is a new file, I can include this file in the list of files to check with gts.

src/plugins/plugin-http.ts
```diff
@@ -44,118 +44,139 @@ function extractUrl(options) {
...
-      options.headers = options.headers || {};
-
-      var uri = extractUrl(options);
-      var requestLifecycleSpan =
-          api.createChildSpan({name: getSpanName(options)});
-      if (!requestLifecycleSpan) {
-        return request.apply(this, arguments);
-      }
+function patchHttps(https, api) {
+  console.log('patch https');
```
ofrobots: remove this

src/plugins/plugin-http.ts
```diff
@@ -44,118 +44,139 @@ function extractUrl(options) {
...
+  shimmer.wrap(https, 'request', function requestWrap(request) {
+    return makeRequestTrace(request, api);
+  });
+  shimmer.wrap(https, 'get', function requestWrap(get) {
```
ofrobots: `https.get` didn't exist before 8.0.0 as well right? I think this code assumes that you will never reach here because of the semver range in `module.exports` array. It would better to have the logic similar (potentially shared) between http and https rather than having two slightly different copies.

src/plugins/plugin-http.ts
kjin: 😨

src/plugins/plugin-http.ts
kjin: I believe whether external/internal `request` is used for `get` is inconsistent between `http` and `https` across different versions of Node. AFAIK:

* `https.get` calls `https.request` for `<8.9.0` and (`>=8.9.1 && <9.0.0`)
* `http.get` calls `http.request` for `<8.0.0`

I think it's more readable to have slightly more duplicated code than have one patch and unpatch function and add an additional parameter that specifies the version range in which `get` must be separately patched/unpatched.

PR #594, opened 2017-11-13
fix: propagate context in koa tracing

### LINKED ISSUE #593: koa async await child spans not working
We just activated this in our project and we have 1 root span per request and no child spans in our trace... resulting in 1 span for an entire request.

I did some debugging to attempt to understand the problem by using the `traceApi` to create a child span in our first koa middleware. The result is always null (this is the same result the `pg` plugin gives). Enabling the logLevel 4, I see:
* `WARN:@google-cloud/trace-agent: Custom Span API: Attempted to create child span without root`
* `WARN:@google-cloud/trace-agent: pg: Attempted to create child span without root`

We **do** use native `async/await` and no babel transpiling. I am running with `GCLOUD_TRACE_NEW_CONTEXT=1`. I confirmed the `cls-ah.js` module is being loaded (indicating the `GCLOUD_TRACE_NEW_CONTEXT=1`) is working.

It seems the `async_hooks` stuff is not working in our case.

Our current setup:
* `node v8.9.1` (also tested with v8.9.1)
* `koa@v2.3.4`
* `@google-cloud/trace-agent@2.3.1`
* `pg@6.4.2`
* __no babel, only native js__

Any ideas what else might be wrong? What to try???

kjin:
Fixes #593

The Koa tracing plugin ignored the return value of `api.wrap`, resulting in trace context not being propagated correctly. This affects both Koa 1 and 2, but Koa 1 needs different patching logic altogether (as it relies on generator functions).

### DIFF
--- a/src/plugins/plugin-koa.ts
+++ b/src/plugins/plugin-koa.ts
@@ -16,12 +16,11 @@
 'use strict';
 
 const shimmer = require('shimmer');
-const co = require('co');
-var urlParse = require('url').parse;
+const urlParse = require('url').parse;
 
 function startSpanForRequest(api, req, res, next) {
-  var originalEnd = res.end;
-  var options = {
+  const originalEnd = res.end;
+  const options = {
     name: urlParse(req.url).pathname,
     url: req.url,
     traceContext: req.headers[api.constants.TRACE_CONTEXT_HEADER_NAME],
@@ -29,14 +28,14 @@ function startSpanForRequest(api, req, res, next) {
   };
   return api.runInRootSpan(options, function(root) {
     // Set response trace context.
-    var responseTraceContext =
+    const responseTraceContext =
       api.getResponseTraceContext(options.traceContext, !!root);
     if (responseTraceContext) {
       res.setHeader(api.constants.TRACE_CONTEXT_HEADER_NAME, responseTraceContext);
     }
     
     if (!root) {
-      return;
+      return next;
     }
 
     api.wrapEmitter(req);
--- a/src/plugins/plugin-koa.ts
+++ b/src/plugins/plugin-koa.ts
@@ -16,6 +16,7 @@
 'use strict';
 
 const shimmer = require('shimmer');
+const co = require('co');
 var urlParse = require('url').parse;
 
 function startSpanForRequest(api, req, res, next) {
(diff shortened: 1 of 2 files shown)


src/plugins/plugin-koa.ts
```diff
@@ -16,6 +16,7 @@
 'use strict';
 
 const shimmer = require('shimmer');
+const co = require('co');
 var urlParse = require('url').parse;
 
 function startSpanForRequest(api, req, res, next) {
```
jinwoo: It'd be good to declare the return type explicitly for a better readability.

src/plugins/plugin-koa.ts
```diff
@@ -16,6 +16,7 @@
 'use strict';
 
 const shimmer = require('shimmer');
+const co = require('co');
```
jinwoo: Why is this needed?

test/plugins/test-trace-koa.ts
```diff
@@ -42,7 +42,7 @@ describe('koa', function() {
   });
 
   Object.keys(appBuilders).forEach(function(version) {
-    describe(version, function() {
+    describe.skip(version, function() {
```
jinwoo: Why skip?

src/plugins/plugin-koa.ts
```diff
@@ -94,9 +101,9 @@ function createMiddleware2x(api) {
   return function middleware(ctx, next) {
     const req = ctx.req;
     const res = ctx.res;
-
-    startSpanForRequest(api, req, res, next);
-
+    console.log(next);
```
jinwoo: Delete?

src/plugins/plugin-koa.ts
```diff
@@ -94,9 +101,9 @@ function createMiddleware2x(api) {
...
+    next = startSpanForRequest(api, req, res, next);
+    console.log(next);
```
jinwoo: Delete?

src/plugins/plugin-koa.ts
jinwoo: Can you put some type declarations here again?

src/plugins/plugin-koa.ts
kjin: 😓

src/plugins/plugin-koa.ts
kjin: Done

kjin:
@jinwoo Yep, adding types for plugins is in the pipeline. I'm trying to figure out how to import types for multiple versions of the same module, which is the major blocker.

PR #599, opened 2017-11-15
refactor: move index.ts and config.ts into src

kjin:
Follow-up to https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/554#issuecomment-328688840

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@google-cloud/trace-agent",
   "version": "2.3.3",
   "description": "Node.js Support for StackDriver Trace",
-  "main": "build/index.js",
+  "main": "build/src/index.js",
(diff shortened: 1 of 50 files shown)


package.json
```diff
@@ -18,22 +18,22 @@
...
     "build/src/plugin-types.d.ts",
+    "build/src/trace-labels.d.ts",
     "build/src/plugins/*.js",
     "doc",
     "CHANGELOG.md",
     "LICENSE",
     "README.md"
   ],
-  "types": "build/index.d.ts",
+  "types": "build/src/index.d.ts",
```
ofrobots: While you're in here, move this next to the main so that changes to one are more visibly caught

Issue #582, opened 2017-10-24
Support for mysql2


julienvincent:
I have built a plugin that adds support for the [node-mysql2](https://github.com/sidorares/node-mysql2) module. Its based heavily on the existing `mysql` plugin in this repository and only required a few minor changes.

Is this something that GoogleCloudPlatform would like to consume and maintain, or should I publish it separately as a third party plugin? I will be happy to submit a PR if the former.

kjin:
We would be happy to see a PR opened for this! Given that the changes are minor, it doesn't sound like it would require much additional maintenance.

julienvincent:
@kjin Sorry for the late response. As the changes I made to the original `mysql` plugin are so minor, I am wondering what the best approach here would be.

It feels a little dirty to create a whole new plugin that duplicates 98% of the `mysql` plugins code, however as plugins modules are declared with `module.exports = [ ... ]` its difficult to import shared code from other plugins (not that you would really want to do this anyway).

Would it be acceptable to create a `mysql-shared.js` (or some such) that both the mysql and mysql2 plugins can require or would you rather keep plugins completely independent?

kjin:
That's a valid concern, esp. given that `mysql` and `mysql2` have very closely resembling public APIs -- and under different circumstances a shared JS module makes the most sense. However, I'd still prefer to keep the plugins independent, because it's nice to have each module's corresponding monkeypatches summed up into a single file. It helps reduce the barrier to entry for other users that also need custom tracing behavior for a module.

That being said, this is a really compelling point that we will consider going into the next semver-major version so I'm glad you brought that up. If you have any ideas on how to neatly handle duplicated code while still making the path for users to create custom plugins, please let us know (we are actively thinking about this). Our web framework plugins, to an extent, already see a certain amount of duplication, so this isn't the only instance where code is potentially being duplicated.

The `mysql` plugin is currently about ~60 lines of actual code, so it wouldn't be very difficult to recuperate differences if we decide to go down a shared code route in the near future.

kjin:
[...]

PR #607, opened 2017-11-27
Port mysql plugin to support mysql2

julienvincent:
This PR adds support for the `mysql2` npm module (as per #582).

I am unsure on how to get the test suite up and running correctly, if someone could point me in the right direction I'd be happy to also write tests for the added plugin 👍 

I will put comments on the relevant lines that changed from `plugin-mysql.js`

### DIFF
--- a/src/plugins/plugin-mysql2.ts
+++ b/src/plugins/plugin-mysql2.ts
@@ -1,0 +1,101 @@
+/**
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+var shimmer = require('shimmer');
+
+var SUPPORTED_VERSIONS = '^1.3.5';
+
+function createCreateQueryWrap(api) {
+  return function createQueryWrap(createQuery) {
+    return function createQuery_trace(sql, values, cb, config) {
+      var span = api.createChildSpan({
+        name: 'mysql-query'
+      });
+      var query = createQuery.apply(this, arguments);
+      if (!span) {
+        return query;
+      }
+      if (api.enhancedDatabaseReportingEnabled()) {
+        span.addLabel('sql', query.sql);
+        if (query.values) {
+          span.addLabel('values', query.values);
+        }
+      }
+      api.wrapEmitter(query);
+      if (query.onResult) {
+        query.onResult = wrapCallback(api, span, query.onResult);
+      } else {
+        query.on('end', function() {
+          span.endSpan();
+        });
+      }
+      return query;
+    };
+  };
+}
+
+function wrapCallback(api, span, done) {
+  var fn = function(err, res) {
+    if (api.enhancedDatabaseReportingEnabled()) {
+      if (err) {
+        span.addLabel('error', err);
+      }
+      if (res) {
+        span.addLabel('result', res);
+      }
+    }
+    span.endSpan();
+    if (done) {
+      done(err, res);
+    }
+  };
+  return api.wrap(fn);
+}
+
+function createWrapGetConnection(api) {
+  return function wrapGetConnection(getConnection) {
+    return function getConnection_trace(cb) {
+      return getConnection.call(this, api.wrap(cb));
+    };
+  };
+}
+
+module.exports = [
+  {
+    file: 'lib/connection.js',
+    versions: SUPPORTED_VERSIONS,
+    patch: function(Connection, api) {
+      shimmer.wrap(Connection, 'createQuery', createCreateQueryWrap(api));
+    },
+    unpatch: function(Connection) {
+      shimmer.unwrap(Connection, 'createQuery');
+    }
+  },
+  {
+    file: 'lib/pool.js',
+    versions: SUPPORTED_VERSIONS,
+    patch: function(Pool, api) {
+      shimmer.wrap(Pool.prototype, 'getConnection',
+                   createWrapGetConnection(api));
+    },
+    unpatch: function(Pool) {
+      shimmer.unwrap(Pool.prototype, 'getConnection');
+    }
+  }
+];
+
+export default {};
--- a/config.ts
+++ b/config.ts
@@ -168,6 +168,7 @@ export const defaultConfig = {
     'koa': path.join(__dirname, 'src/plugins/plugin-koa.js'),
     'mongodb-core': path.join(__dirname, 'src/plugins/plugin-mongodb-core.js'),
     'mysql': path.join(__dirname, 'src/plugins/plugin-mysql.js'),
+    'mysql2': path.join(__dirname, 'src/plugins/plugin-mysql2.js'),
     'pg': path.join(__dirname, 'src/plugins/plugin-pg.js'),
     'redis': path.join(__dirname, 'src/plugins/plugin-redis.js'),
     'restify': path.join(__dirname, 'src/plugins/plugin-restify.js')


src/plugins/plugin-mysql2.ts
```diff
@@ -0,0 +1,101 @@
...
+ */
+'use strict';
+
+var shimmer = require('shimmer');
+
+var SUPPORTED_VERSIONS = '^1.3.5';
+
+function createCreateQueryWrap(api) {
+  return function createQueryWrap(createQuery) {
+    return function createQuery_trace(sql, values, cb, config) {
```
julienvincent: requires an additional `config` argument

src/plugins/plugin-mysql2.ts
```diff
@@ -0,0 +1,101 @@
...
+      }
+      if (api.enhancedDatabaseReportingEnabled()) {
+        span.addLabel('sql', query.sql);
+        if (query.values) {
+          span.addLabel('values', query.values);
+        }
+      }
+      api.wrapEmitter(query);
+      if (query.onResult) {
+        query.onResult = wrapCallback(api, span, query.onResult);
```
julienvincent: Property `_callback` renamed to `onResult`

src/plugins/plugin-mysql2.ts
```diff
@@ -0,0 +1,101 @@
...
+    versions: SUPPORTED_VERSIONS,
+    patch: function(Pool, api) {
+      shimmer.wrap(Pool.prototype, 'getConnection',
+                   createWrapGetConnection(api));
+    },
+    unpatch: function(Pool) {
+      shimmer.unwrap(Pool.prototype, 'getConnection');
+    }
+  }
+];
```
julienvincent: Target files renamed to be lowercase (`Connection` => `connection`, `Pool` => `pool`)

src/plugins/plugin-mysql2.ts
```diff
@@ -0,0 +1,101 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+var shimmer = require('shimmer');
+
+var SUPPORTED_VERSIONS = '^1.3.5';
```
julienvincent: Different version from `mysql`

kjin:
Thanks for the PR! I'm taking a look now... in the meantime, there's a script [bin/docker-trace.sh](../tree/master/bin/docker-trace.sh) that contains commands for launching docker containers for each of these databases. To make unit tests work you'll need to have all four services running. This should be the only pre-requisite to running unit tests with `npm test`.

As the APIs are similar it may be sufficient to edit [test/plugins/test-trace-mysql.ts](../tree/master/test/plugins/test-trace-mysql.ts) to make it so that every unit test is run twice, once for `mysql` and `mysql2` each. The `mysql` library is currently being loaded from a fixture directory named at `fixtures/mysql2`, which might be confusing -- this refers to `mysql@2`. You should probably rename it so that there are co-existing fixtures `mysql-2` and `mysql2-1`.

If you don't have time to do this that's fine, don't hesitate to let me know and I will update the tests.

src/plugins/plugin-mysql2.ts
kjin: Do you know that this only works for mysql2 > 1.3.5? Is it possible to expand the range to start at a 1.x.0?

julienvincent:
Great! That sounds simple enough - I will finish up the tests tomorrow morning then 👍

src/plugins/plugin-mysql2.ts
```diff
@@ -0,0 +1,101 @@
...
+    };
+  };
+}
+
+module.exports = [
+  {
+    file: 'lib/connection.js',
+    versions: SUPPORTED_VERSIONS,
+    patch: function(Connection, api) {
+      shimmer.wrap(Connection, 'createQuery', createCreateQueryWrap(api));
```
kjin: Does this need to be `Connection.prototype`?

src/plugins/plugin-mysql2.ts
kjin: I believe this is unnecessary if `config` is not being used. We already make a call to `createQuery` using the `arguments` object.

julienvincent: In which case non of those arguments are really necessary, correct?

src/plugins/plugin-mysql2.ts
julienvincent: I will test and see

src/plugins/plugin-mysql2.ts
julienvincent: No, createQuery isn't a prototypal property. Observe: https://github.com/sidorares/node-mysql2/blob/master/lib/connection.js#L586

julienvincent:
I'm getting the following failed test when running locally:

```bash
  1) test-trace-connect
       should remove trace frames from stack:

      Uncaught AssertionError [ERR_ASSERTION]: 'Namespace.runAndReturn [as runAndReturn]' == 'middleware'
      + expected - actual

      -Namespace.runAndReturn [as runAndReturn]
      +middleware

      at ClientRequest.<anonymous> (/Users/julienvincent/code/cloud-trace-nodejs/test/plugins/test-trace-connect.ts:127:16)
      at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:565:21)
      at HTTPParser.parserOnHeadersComplete (_http_common.js:116:23)
      at Socket.socketOnData (_http_client.js:454:20)
      at addChunk (_stream_readable.js:252:12)
      at readableAddChunk (_stream_readable.js:239:11)
      at Socket.Readable.push (_stream_readable.js:197:10)
      at TCP.onread [as _originalOnread] (net.js:589:20)
      at TCP.onread (/Users/julienvincent/code/cloud-trace-nodejs/node_modules/async-listener/glue.js:188:31)
```

As far as I can see this has nothing to do with the changes I made. Any ideas?

kjin:
Oops - this test failure has to do with the `GCLOUD_TRACE_NEW_CONTEXT` environmental variable; this should be set to `1`. Sorry -- that slipped my mind earlier.

src/plugins/plugin-mysql2.ts
kjin: Ah, ok. I think I confused `query` and `createQuery`

src/plugins/plugin-mysql2.ts
kjin: Ahh, yeah. For some reason I thought `cb` was being used which would be why I thought the args were spelled out in the first place.

I guess they aren't necessary but let's keep it for readability's sake.

julienvincent:
I'm struggling with a strange issue now that I can't piece together. @kjin Perhaps you could help me understand whats happening here:

Defining a patcher for `lib/pool.js` of mysql2 breaks the constructor of `Pool`. This seems to happen whether or not I alter properties on its prototype - simply defining the patcher breaks it. The error I get in the tests is as follows:

```bash
  test-trace-mysql-2
    ✓ should perform basic operations
    ✓ should propagate context
    ✓ should remove trace frames from stack
    ✓ should work with events
    ✓ should work without events or callback (54ms)
    ✓ should perform basic transaction

  test-trace-mysql2-1
    1) "before all" hook
    2) "after all" hook


  6 passing (368ms)
  2 failing

  1) test-trace-mysql2-1
       "before all" hook:
     TypeError: Pool is not a constructor
      at Object.module.exports.createPool (test/plugins/fixtures/mysql2-1/node_modules/mysql2/index.js:17:10)
      at Context.<anonymous> (/Users/julienvincent/code/cloud-trace-nodejs/test/plugins/test-trace-mysql.ts:44:20)

  2) test-trace-mysql2-1
       "after all" hook:
     TypeError: Cannot read property 'end' of undefined
      at Context.<anonymous> (/Users/julienvincent/code/cloud-trace-nodejs/test/plugins/test-trace-mysql.ts:48:11)
```
and the exports of the `plugin-mysql2.ts` look as follows:

```js
module.exports = [
  {
    file: 'lib/connection.js',
    versions: SUPPORTED_VERSIONS,
    patch: function(Connection, api) {
      shimmer.wrap(Connection, 'createQuery', createCreateQueryWrap(api));
    },
    unpatch: function(Connection) {
      shimmer.unwrap(Connection, 'createQuery');
    }
  },
  {
    file: 'lib/pool.js',
    versions: SUPPORTED_VERSIONS,
    patch: function(Pool, api) {
      // shimmer.wrap(Pool.prototype, 'getConnection',
      //              createWrapGetConnection(api));
    },
    unpatch: function(Pool) {
      // shimmer.unwrap(Pool.prototype, 'getConnection');
    }
  }
];
```

As you can see the patcher itself is not doing anything to `Pool`, this leads me to believe that something is happening within `trace-plugin-loader.ts` to cause this - though I can't see anything obvious that would cause this.

What I have done for now is change the pool patcher to target `index.js` which seems to work as expected, but I would still like to figure out whats happening here as it may reveal a bug in the plugin loader.

julienvincent:
I don't think the tests that just failed on CircleCI were due to my changes - could you retry the build?

kjin:
@julienvincent I looked into the lib/pool.js patch issue you mentioned yesterday... I'm as baffled as you are. Investigating more today.

julienvincent:
I can say that I was able to patch that file fine while using version `2.2.0` of this module within my personal project (before starting this PR) and using the above plugin as an external plugin. Perhaps a change has been made since then to cause this?

kjin:
Yes, it seems like it.

I think I understand the issue now -- will open a bug imminently. In the meantime, if the patch to `index.js` is sufficient, let's use that.

julienvincent:
Aha! Circular references - that makes sense now.

Yes the patch to `index.js` seems to work fine 👍

PR #609, opened 2017-11-28
fix: add projectId to config type definitions

kjin:
Somehow I left out `projectId` from the `Config` object type definitions (probably because it doesn't have a default value)... this change adds it.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -126,6 +126,12 @@ export interface Config {
   ignoreContextHeader?: boolean;
 
   /**
+   * The ID of the Google Cloud Platform project with which traces should
+   * be associated.
+   */
+  projectId?: string;
+
+  /**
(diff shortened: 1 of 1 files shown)


src/config.ts
```diff
@@ -125,6 +125,12 @@ export interface Config {
    */
   ignoreContextHeader?: boolean;
 
+  /**
+   * The ID of the Google Cloud Platform project with which traces should
+   * be associated.
+   */
+  projectId?: string;
```
jinwoo: Qq. Shouldn't this be required? What happens if the project id is not given?

kjin: It's not required. Project ID can also be either read from the metadata service, or the environment.

kjin: On that note, let me add a comment to further elaborate on this.

jinwoo: Thanks.

PR #613, opened 2017-11-29
fix: listenerAttached should be set to true

jinwoo:
Otherwise, multiple listeners are attached when the client code attaches multiple `data` listeners.

### DIFF
--- a/test/plugins/test-trace-http.ts
+++ b/test/plugins/test-trace-http.ts
@@ -45,6 +45,9 @@ describe('test-trace-http', function() {
 
   afterEach(function() {
     common.cleanTraces();
+  });
+
+  after(function() {
     server.close();
   });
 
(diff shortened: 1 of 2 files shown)


test/plugins/test-trace-http.ts
```diff
@@ -45,6 +45,9 @@ describe('test-trace-http', function() {
 
   afterEach(function() {
     common.cleanTraces();
+  });
+
+  after(function() {
```
kjin: Looking at the CI, it seems like the creation of the server should be in a `beforeEach` rather than the `server.close` call be moved to `after`.

jinwoo: Yeah, I just noticed the failures and am looking. I'm wondering how tests worked when the server was started only once and closed after each test.

jinwoo: I was mistaken. `server.close()` doesn't invalidate the server itself, but just stops accepting connections. So it should be done from `afterEach()`. Reverted that change.

PR #615, opened 2017-11-29
fix: wrong test/assertion in test-trace-http

jinwoo:
The wrong test is 'should not trace api requests'. The header name to
check should be TRACE_AGENT_REQUEST_HEADER, not
TRACE_CONTEXT_HEADER_NAME. Also the assertion should be done against the
number of spans, not against the number of traces.

Also fix the order of arguments to `assert.equal()` calls. The first
argument must be the actual result and the second the expected value.


jinwoo:
CI failures seem like a flakiness. @kjin can you rerun Appveyor?

jinwoo:
@kjin All tests passed now. Can you take a look?

PR #619, opened 2017-12-01
fix: use shimmer to wrap event emitters in async_hooks-based cls

### LINKED ISSUE #600: got@7 behavior changes when being traced
The `got` module (at version 7) doesn't response body decompression when `http` is patched by the Trace Agent. We are working with a few dependency modules to resolve this -- in the meantime, if you observe this behavior, please try got@6 or ignore incoming URLs that trigger outgoing HTTP requests through got (or, as a more heavy-handed approach, disable the http plugin entirely with the config option `{ 'plugins': { 'http': null } }`).

kjin:
Fixes #600 for `async_hooks`-based context tracking.

Background -- got 7+ does gzip decompression by piping the `http.IncomingMessage` stream through an `unzip` transform, and then through a `stream.PassThrough` transform. To make the shape of the resulting `stream.PassThrough` stream look like `http.IncomingMessage`, it copies enumerable properties from the initial `http.IncomingMessage` stream to the new one. Listener functions like `on` are not enumerable, so they don't get copied to the `stream.PassThrough` (and they shouldn't be). [[source](https://github.com/sindresorhus/decompress-response/blob/master/index.js)]

Our way of patching event emitters causes functions like '`on`' to be enumerable, so the `stream.PassThrough` object has the wrong methods on it. This causes gzip decompression to fail.

This change uses `shimmer` to wrap event emitters, since it preserves property enumerability when overriding object properties.

### DIFF
--- a/src/cls-ah.ts
+++ b/src/cls-ah.ts
@@ -16,6 +16,7 @@
 
 import * as asyncHook from 'async_hooks';
 import {Context, Func, Namespace as CLSNamespace} from 'continuation-local-storage';
+import * as shimmer from 'shimmer';
 
 const wrappedSymbol = Symbol('context_wrapped');
 let contexts: {[asyncId: number]: Context;} = {};
@@ -97,10 +98,13 @@ class AsyncHooksNamespace implements CLSNamespace {
     EVENT_EMITTER_METHODS.forEach((method) => {
       // TODO(kjin): Presumably also dependent on MS/TS-#15473.
       // tslint:disable:no-any
-      const oldMethod = (ee as any)[method];
-      (ee as any)[method] = function(this: {}, event: string, cb: Func<void>) {
-        return oldMethod.call(this, event, ns.bind(cb));
-      };
+      if ((ee as any)[method]) {
+        shimmer.wrap(ee, method, (oldMethod) => {
+          return function(this: {}, event: string, cb: Func<void>) {
+            return oldMethod.call(this, event, ns.bind(cb));
+          };
+        });
+      }
       // tslint:enable:no-any
     });
   }


src/cls-ah.ts
```diff
@@ -97,10 +98,11 @@ class AsyncHooksNamespace implements CLSNamespace {
     EVENT_EMITTER_METHODS.forEach((method) => {
       // TODO(kjin): Presumably also dependent on MS/TS-#15473.
       // tslint:disable:no-any
-      const oldMethod = (ee as any)[method];
-      (ee as any)[method] = function(this: {}, event: string, cb: Func<void>) {
-        return oldMethod.call(this, event, ns.bind(cb));
-      };
+      shimmer.wrap(ee, method, (oldMethod) => {
+        return function(this: {}, event: string, cb: Func<void>) {
+          return oldMethod.call(this, event, ns.bind(cb));
```
jinwoo: Nit. AIUI `bind()` is slow. Can this be
```typescript
return oldMethod.call(this, event, (...args: Array<{}>) => cb.call(ns, ...args));
```
?

I'm not actually sure which one is better.

jinwoo: Why is `ns` needed here? Isn't it identical to `this`? In that case, woud
```typescript
return oldMethod.call(this, event, (...args: Array<{}>) => this.call(...args));
```
work?

kjin: The name `bind` is overloaded... but this doesn't refer to `Function#bind`, but rather `CLSNamespace#bind`.

kjin: `ns !== this` on line 103, because the value of `this` at line 103 is dynamically scoped.

jinwoo: Oh, I see. Didn't know about CLS.

PR #621, opened 2017-12-05
refactor: add web framework plugin types and script to fetch types

### LINKED ISSUE #620: Koa route path labels are likely never applied
The Trace Agent [expects](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/f1ca3a39e101c3b5a5a858e2a2a2c42549dbeb67/src/plugins/plugin-koa.ts#L59) the Koa route path to exist in `ctx.req.route.path`; however, neither Koa nor `koa-router` write to this field.

kjin:
Fixes #620

This change adds TypeScript types for all web framework tracing plugins and subjects them to `gts`, ~~In addition, it introduces an npm verb to download type definitions separately into a directory distinct from `node_modules`. The purpose of this is to depend on several versions of type definitions, something that isn't supported by npm. Currently, this doesn't apply to any type definitions (Koa 1 and 2 have different types, but @types/koa@1 doesn't exist, and the minimal type definitions for koa 1 can share types with koa 2).~~

~~I'm considering avoiding this entirely (it adds a considerable amount of complexity to the codebase), and having the Trace Agent simply depend on one version of each module instead. So far this is feasible, but it might not be in the long term. I would like to get thoughts on this.~~

This change also fixes a bug where framework-specific routing trace labels aren't generated because the Trace Agent expects them on the wrong property. This is apparent in `koa` (#620) and `connect` (because AFAIK connect's route path _is_ the original URL).

Update: Changed this so that plugin types are simply installed under devDependencies. There is no script to fetch types anymore. I might add that later, thinking of crossing that bridge when the times comes (currently, besides Koa 1 which has no `@types` module anyway, there are no modules whose interfaces change dramatically between versions)


scripts/index.ts
```diff
@@ -2,6 +2,7 @@ const [bin, script, ...steps] = process.argv;
 
 import { checkInstall } from './check-install';
 import { encryptCredentials, decryptCredentials } from './credentials';
+import { getPluginTypes } from './get-plugin-types';
```
jinwoo: Where is `get-plugin-types`? And why is it imported?

src/plugins/plugin-connect.ts
```diff
@@ -13,26 +13,50 @@
...
+import {connect_3} from './types';
+
+type Connect3 = typeof connect_3;
+// Connect docs note that routed requests have an originalUrl property.
+// https://github.com/senchalabs/connect/tree/3.6.5#appuseroute-fn
+type Request = IncomingMessage&{originalUrl?: string};
+
+const SUPPORTED_VERSIONS = '3.x';
+
+function getFirstHeader(req: IncomingMessage, key: string) {
```
jinwoo: Maybe specify the return type as `string`?

src/plugins/plugin-connect.ts
```diff
@@ -42,8 +66,8 @@ function createMiddleware(api) {
       api.wrapEmitter(req);
       api.wrapEmitter(res);
 
-      var url = (req.headers['X-Forwarded-Proto'] || 'http') +
-        '://' + req.headers.host + req.originalUrl;
+      const url = (req.headers['X-Forwarded-Proto'] || 'http') + '://' +
```
jinwoo: Better to use a template string.

src/plugins/plugin-express.ts
```diff
@@ -55,16 +58,16 @@ function patchModuleRoot(express, api) {
       api.wrapEmitter(req);
       api.wrapEmitter(res);
 
-      var url = req.protocol + '://' + req.hostname + req.originalUrl;
+      const url = req.protocol + '://' + req.hostname + req.originalUrl;
```
jinwoo: template string

src/plugins/plugin-hapi.ts
```diff
@@ -13,40 +13,51 @@
...
-  };
+import {PluginTypes} from '..';
+
+import {hapi_16} from './types';
+
+type Hapi16Module = typeof hapi_16;
+
+const SUPPORTED_VERSIONS = '8 - 16';
+
+function getFirstHeader(req: IncomingMessage, key: string) {
```
jinwoo: return type

src/plugins/plugin-hapi.ts
```diff
@@ -56,36 +67,36 @@ function createMiddleware(api) {
       api.wrapEmitter(req);
       api.wrapEmitter(res);
 
-      var url = (req.headers['X-Forwarded-Proto'] || 'http') +
-      '://' + req.headers.host + req.url;
-    
+      const url = (req.headers['X-Forwarded-Proto'] || 'http') + '://' +
```
jinwoo: template string

src/plugins/plugin-koa.ts
```diff
@@ -13,134 +13,170 @@
...
+
+type Koa1Module = typeof koa_1;
+type Koa2Module = typeof koa_2;
+// routePath is populated if the user uses the koa-route module.
+type KoaContext = (koa_1.Context|koa_2.Context)&{routePath?: string | RegExp};
+
+interface KoaModule<T> {
+  // TypeScript isn't expressive enough, but KoaModule#use should return `this`.
+  // tslint:disable-next-line:no-any
+  readonly prototype: {use: (m: T) => any};
```
jinwoo: I'm not sure but would something like this work?

```typescript
readonly prototype: {use: (this: any, m: T) => this};
```

src/plugins/plugin-koa.ts
```diff
@@ -13,134 +13,170 @@
...
+}
+
+// Function signature for createMiddleware[2x]
+type CreateMiddlewareFn<T> = (api: PluginTypes.TraceAgent) => T;
+// Function signature for a function that returns the value of the "next"
+// middleware function parameter, wrapped to propagate context based on the
+// propagateContext flag. The type of "next" differs between Koa 1 and 2.
+type GetNextFn<T> = (propagateContext: boolean) => T;
+
+function getFirstHeader(req: IncomingMessage, key: string) {
```
jinwoo: return type

src/plugins/plugin-koa.ts
```diff
@@ -13,134 +13,170 @@
...
-      return next;
+      return getNext(false);
     }
 
     api.wrapEmitter(req);
     api.wrapEmitter(res);
 
-    const url = (req.headers['X-Forwarded-Proto'] || 'http') +
-      '://' + req.headers.host + req.url;
+    const url = (req.headers['X-Forwarded-Proto'] || 'http') + '://' +
```
jinwoo: template string

src/plugins/plugin-koa.ts
```diff
@@ -13,134 +13,170 @@
...
-  return function* middleware(next) {
-    /* jshint validthis:true */
-    const req = this.req;
-    const res = this.res;
-
-    next = startSpanForRequest(api, req, res, next);
-
+function createMiddleware(api: PluginTypes.TraceAgent): koa_1.Middleware {
+  // Koa 1 type definitions use any here.
+  // tslint:disable-next-line:no-any
```
jinwoo: Doesn't seem needed?

scripts/index.ts
kjin: Oops -- removed this

This was part of a previous commit, back when I would fetch plugin types separately from `npm install`. The most recent commit removed this script but I forgot to include this change

src/plugins/plugin-koa.ts
kjin: The best we can do is

```ts
readonly prototype: {use: (this: any, m: T) => typeof this};
```

which loses the information that it returns `this`.

src/plugins/plugin-koa.ts
kjin: Removed

src/plugins/plugin-koa.ts
kjin: Done

src/plugins/plugin-koa.ts
kjin: Done

src/plugins/plugin-hapi.ts
kjin: Done

src/plugins/plugin-hapi.ts
kjin: Done

src/plugins/plugin-express.ts
kjin: Done

src/plugins/plugin-connect.ts
kjin: Done

src/plugins/plugin-connect.ts
kjin: Actually it would have to be `string | null` since it might not exist. (Small detail -- choosing `| null` over `| undefined` since I don't want to pass around undefined values)

kjin:
New changes are in 1393994 only

kjin:
Ignoring codecov for now... will be fixed in follow-up PR.

PR #624, opened 2017-12-13
fix: upgrade semver to ^5.4.1

kjin:
We use `semver.prerelease`, which is introduced in semver 5.2. This causes issues because we only necessitate that semver 5.0 or higher be installed.


kjin:
Blocked by #625.

ofrobots:
@kjin: can we get this out in a release?

PR #625, opened 2017-12-13
fix: use new node error constructor type definitions

kjin:
`@types/node` recently introduces built-in `CallSite` and `ErrorConstructor` type definitions, so this change removes them. (The compilation will fail otherwise because the definitions are slightly incompatible.)


ofrobots:
I think you should bump the version of `@types/node` in `package.json` to make sure we get the new types.

kjin:
@ofrobots good idea, done

PR #630, opened 2018-01-04
fix: extend supported gRPC version range to 1.8

kjin:
* Bump supported gRPC version range to 1.8
* Change gRPC fixture versions to be `~` instead of `^` (to fix CI failure after gRPC 1.8 was released)
* datastore test: Change checked error code from HTTP 401 (unauthorized) to gRPC 16 (unauthenticated) -- reflects change in gRPC 1.8
* For convenince, add `npm run script` npm script


stevenaldinger:
heads up, we needed this update and gave it a whirl (you had opened the PR 22 minutes prior) and we got "unknown errors" in our grpc client suddenly. I'm at work right now and all proprietary unfortunately but things went from simply not showing up in the cloud trace UI to actually breaking the client itself. i was already using a fork with a semver patch (one line of code change in the loader) and merged this in/rebuilt and had no luck. wish i could help more, I'm not a type scripter or git expert so very well could be user error, I see the CI tests passed.

kjin:
@stevenaldinger This PR shouldn't result in any additional behavior changes because [JavaScript changes in gRPC 1.8 seem to be minimal](https://gist.github.com/kjin/48c261cca7a7a2fb6712fef6b2e8434e). Do you know if this is an issue with gRPC itself? It's possible considering the unknown errors you got -- to my knowledge the Trace Agent isn't emitting or changing gRPC errors.

stevenaldinger:
Honestly the code base I tried using it with is such a mess I felt guilty even posting that. In about 3 hours I'll reply back here when I can try it with my setup at home. Very very believable it'd be something outside of your control here, especially with what you just said.

stevenaldinger:
Had reverted to the current stable version on npm and got the same error after a while, safe to say its not you. Thanks a lot for this PR! Can't wait to use it

kjin:
Thanks for keeping us updated!

PR #643, opened 2018-01-11
refactor: use setHeader to set trace context header

### LINKED ISSUE #639: Seeing performance issues with higher outbound http client requests
This is on node 8.9.4 and `@google-cloud/trace-agent@2.3.3`

7.2% seems a bit high to me?

The top of the JavaScript portion of the cpu profile looks like this:

```
[JavaScript]:
   ticks  total  nonlib   name
   3887    9.8%   10.0%  Builtin: KeyedLoadIC_Megamorphic
   3206    8.1%    8.2%  Builtin: KeyedStoreIC_Megamorphic
   2876    7.2%    7.4%  LazyCompile: *<anonymous> /opt/app/node_modules/lodash.merge/index.js:973:37
```

I'm *thinking* it is coming from https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/38d4673e433bc1ae156bf9d962c767ca0703ef7f/src/plugins/plugin-http.ts#L68 but could be wrong. I don't have it narrowed down to a reproducible test case yet, but I'm working on that now.

kjin:
Fixes #639 

The HTTP tracing plugin should not be trying to clone the HTTP request options object if it's not even clear if the request is being traced. This change ~~moves the clone to after this check has been made~~ removes the `merge` call in favor of using `setHeader` to mutate headers after `http.request` has been called.

### DIFF
--- a/src/plugins/plugin-http.ts
+++ b/src/plugins/plugin-http.ts
@@ -18,14 +18,10 @@
 var shimmer = require('shimmer');
 var url = require('url');
 var isString = require('is').string;
-var merge = require('lodash.merge');
 var httpAgent = require('_http_agent');
 var semver = require('semver');
 
 function getSpanName(options) {
-  if (isString(options)) {
-    options = url.parse(options);
-  }
   // c.f. _http_client.js ClientRequest constructor
   return options.hostname || options.host || 'localhost';
 }
@@ -65,22 +61,35 @@ function makeRequestTrace(request, api) {
       return request.apply(this, arguments);
     }
 
-    options = isString(options) ? url.parse(options) : merge({}, options);
-    options.headers = options.headers || {};
+    var uri;
+    if (isString(options)) {
+      // save the value of uri so we don't have to reconstruct it later
+      uri = extractUrl(options);
+      options = url.parse(options);
+    }
 
-    var uri = extractUrl(options);
     var requestLifecycleSpan =
         api.createChildSpan({name: getSpanName(options)});
     if (!requestLifecycleSpan) {
       return request.apply(this, arguments);
     }
 
+    if (!uri) {
+      uri = extractUrl(options);
+    }
+
     requestLifecycleSpan.addLabel(api.labels.HTTP_METHOD_LABEL_KEY,
                                   options.method);
     requestLifecycleSpan.addLabel(api.labels.HTTP_URL_LABEL_KEY, uri);
-    options.headers[api.constants.TRACE_CONTEXT_HEADER_NAME] =
+
+    var headers = options.headers ? Object.assign({}, options.headers) : {};
+    headers[api.constants.TRACE_CONTEXT_HEADER_NAME] =
         requestLifecycleSpan.getTraceContext();
-    var req = request.call(this, options, function(res) {
+    // Clone the options object to pass to request options,
+    // using the new set of headers.
+    var requestOptions = Object.assign({}, options, { headers: headers });
+
+    var req = request.call(this, requestOptions, function(res) {
       api.wrapEmitter(res);
       var numBytes = 0;
       var listenerAttached = false;
--- a/src/plugins/plugin-http.ts
+++ b/src/plugins/plugin-http.ts
@@ -50,7 +50,7 @@ function makeRequestTrace(request, api) {
   // Here `request` may also happen to be `get`.
   return function request_trace(options, callback) {
     if (!options) {
-      return request.apply(this, arguments);
+      return request(options, callback);
     }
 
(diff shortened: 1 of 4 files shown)


jinwoo:
This change improves the performance only when the client code is not tracing. The performance degradation still happens if the code is traced, right? I'm not sure if #639 is only for the former, not for the latter. Is there a way to improve for both cases?

I'd imagine most production services would want to enable tracing all the time (at least for some of its traffic), and they wouldn't want the performance to be impacted severely.

kjin:
@jinwoo As long we're claiming a guarantee that the HTTP request options object is not being mutated (this could be up for discussion in the spirit of improving performance), I'm not sure there's a way around cloning the headers object. I'm not sure what the traffic patterns are like in #639 but it might very well be possible that the number of outgoing HTTP requests indeed far exceeds the number of traced requests (we by default limit it to 10 traces/second).

evanlucas:
@kjin would yall actually consider an option for mutating the request object? That seems like it would be the smallest performance hit. https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/issues/639 is a socket.io server that makes outbound http requests on every socket pong that is received, so yes, the outbound requests greatly exceed the inbound requests for most cases.

I also agree with @jinwoo. It would be ideal to tune for the production if we can :]

jinwoo:
@kjin Do you mean that this new code will degrade the performance only for 10 requests/second at maximum, while the old code will degrade for all the requests? If that is the case, I think this is in a much better state, if not ideal.

kjin:
Looking through the docs I'm reminded that `http.setHeader` exists. As the implementation of `http.request` internally copies the headers object already (at least in v4/6/8/9) we might be able to have our cake and eat it too :)

src/plugins/plugin-http.ts
```diff
@@ -65,21 +61,27 @@ function makeRequestTrace(request, api) {
       return request.apply(this, arguments);
     }
 
-    options = isString(options) ? url.parse(options) : merge({}, options);
-    options.headers = options.headers || {};
+    var uri;
```
jinwoo: This can be
```typescript
const uri = extractUrl(options);
```
because `extractUrl()` is called whether `options` is a string or not, right?

ofrobots:
@evanlucas 
> @kjin would yall actually consider an option for mutating the request object?

This was a bug. We were cloning the request options for all requests rather than the sampled requests. I expect that this change will fix the performance issue even without the change to avoid mutation.

src/plugins/plugin-http.ts
kjin: It's a small optimization but I split it so that we wouldn't reconstruct a `uri` from an `options` object if no span is being created.

jinwoo: Makes sense!

kjin:
I changed the wrapped `http.get` to circumvent calling underlying `http.get` altogether, so that we can set the trace context header before the `req.end` is called. This might have implications if the user changed `http.get`, but because it itself is a convenience wrapper over `http.request` I don't think this would be an issue (arguably the user should also be patching `http.request` instead).

PTAL (will make minor changes if CI is still failing)

kjin:
@GoogleCloudPlatform/node-team @evanlucas PTAL

src/plugins/plugin-http.ts
```diff
@@ -137,20 +141,37 @@ function patchHttp(http, api) {
...
+      // Re-implement http.get. This needs to be done (instead of using
+      // makeRequestTrace to patch it) because we need to set the trace
+      // context header before the returned ClientRequest is ended.
+      // The Node.js docs state that the only differences between request and
+      // get are that (1) get defaults to the HTTP GET method and (2) the
+      // returned request object is ended immediately.
+      // The former is already true (at least in supported Node versions up to
+      // v9), so we simply follow the latter.
+      // Ref: https://nodejs.org/dist/latest/docs/api/http.html#http_http_get_options_callback
+      return function makeGetTrace() {
```
evanlucas: Any reason to not explicitly pass the two arguments here? Last I checked, apply() has a performance hit over an explicit function call

kjin: `apply` seemed incrementally less intrusive just because it doesn't modify the implicit argument -- I didn't know it incurs a performance hit over the explicit call. Since `this` is irrelevant I'll change this.

kjin:
CircleCI is backed up... a previous Node 6 build passed (differences are not specific to Node 6) https://circleci.com/gh/GoogleCloudPlatform/cloud-trace-nodejs/1289

Landing.

PR #646, opened 2018-01-12
chore: 2.4.1 release proposal

kjin:


kjin:
Going to wait on #648 as I ended up disabling http2 in #643 for no reason

PR #647, opened 2018-01-12
chore: add npm publish job to circle ci

kjin:
(To be merged after v2.4.1 is released)


.circleci/config.yml
```diff
@@ -119,3 +136,18 @@ jobs:
...
     <<: *unit_tests
+  publish_npm:
+    docker:
+      - image: node:8
+        user: node
+    steps:
+      - checkout
+      - run:
+          name: Set NPM authentication.
+          command: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
```
jinwoo: For this to work, you have to set the `NPM_TOKEN` env var in the circleci setting. The token that we're using is in valentine.

kjin: Done. Thanks!

PR #649, opened 2018-01-13
refactor: types for http tracing plugin

kjin:


src/plugins/plugin-http.ts
```diff
@@ -14,41 +14,65 @@
...
-'use strict';
-var shimmer = require('shimmer');
-var url = require('url');
-var isString = require('is').string;
-var httpAgent = require('_http_agent');
-var semver = require('semver');
-
-function getSpanName(options) {
+import * as httpModule from 'http';
+import {Agent, ClientRequest, ClientRequestArgs, get, request} from 'http';
```
jinwoo: You shouldn't import `get` and `request`. That'll make the `http` module loaded at run time before the it's instrumented. You should import only types using this syntax.

jinwoo: In other words, you should look at the transpiled .js file and see if the `http` module is `require`'d there. It is a problem if it is.

kjin: Good catch -- these imports get erased at compile time because they are only used for types.

jinwoo: Ah, I missed that they are being used only for types. Sorry for the noise.

kjin:
Ignoring codecov check for now... it will change drastically in #651 anyways

PR #654, opened 2018-01-18
Change span ID to use random bytes

draffensperger:
While running a test application that was composed of multiple Node.js apps, each with tracing enabled, we found that span IDs would conflict when the multiple apps were simultaneously started, because each started their span IDs at index 1. This switches to use secure randomness for span IDs to help ensure uniqueness even when multiple Node apps are started at the same time.

### DIFF
--- a/src/span-data.ts
+++ b/src/span-data.ts
@@ -42,11 +42,13 @@ interface StackFrame {
   column_number?: number;
 }
 
-// This evaluates to a large whole number such that any fractional number
-// between 0 and 1 multiplied by it will always give a whole number.
-const SPAN_ID_MAX = (1 << 30) * (1 << 30);
+// Span IDs in the X-Cloud-Trace-Context header are 64-bit unsigned integers
+// This is the largest JavaScript number that is less than the max unsigned
+// 64-bit int, that is less than 2^64-1.
+const SPAN_ID_MAX = 18446744073709550000;
 
 function randomSpanId() {
+  // No rounding is needed because SPAN_ID_MAX is bigger than 1 / Number.EPSILON
   return String(Math.random() * SPAN_ID_MAX);
 }
 
--- a/src/span-data.ts
+++ b/src/span-data.ts
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-import * as crypto from 'crypto';
 import * as util from 'util';
 
 import {Constants} from './constants';
@@ -43,21 +42,12 @@ interface StackFrame {
   column_number?: number;
 }
 
-// Use 6 bytes of randomness only as JS numbers are doubles not 64-bit ints.
-const SPAN_ID_RANDOM_BYTES = 6;
-
-// Use the faster crypto.randomFillSync when available (Node 7+) falling back to
-// using crypto.randomBytes.
-const spanIdBuffer = Buffer.alloc(SPAN_ID_RANDOM_BYTES);
-const randomFillSync = crypto.randomFillSync;
-const randomBytes = crypto.randomBytes;
-const spanRandomBuffer = randomFillSync ?
-    () => randomFillSync(spanIdBuffer) :
-    () => randomBytes(SPAN_ID_RANDOM_BYTES);
+// This evaluates to a large whole number such that any fractional number
+// between 0 and 1 multiplied by it will always give a whole number.
+const SPAN_ID_MAX = (1 << 30) * (1 << 30);
 
 function randomSpanId() {
-  // tslint:disable-next-line:ban Needed to parse hexadecimal.
-  return parseInt(spanRandomBuffer().toString('hex'), 16).toString();
+  return String(Math.random() * SPAN_ID_MAX);
 }
 
 export class SpanData implements SpanDataInterface {
--- a/src/span-data.ts
+++ b/src/span-data.ts
@@ -15,6 +15,7 @@
  */
 
 import * as util from 'util';
+import * as crypto from 'crypto';
 
 import {Constants} from './constants';
 import {SpanData as SpanDataInterface} from './plugin-types';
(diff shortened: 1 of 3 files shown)


src/span-data.ts
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import {randomBytes} from 'crypto';
```
evanlucas: It would probably be beneficial to try and use [`crypto.randomFillSync()`](https://nodejs.org/dist/latest-v8.x/docs/api/crypto.html#crypto_crypto_randomfillsync_buffer_offset_size) if it exists. It is ~30% faster last time I benchmarked it

kjin:
@draffensperger FYI: I opened a PR to update CONTRIBUTING.md (#655) to write about running unit tests locally. What was the problem you encountered with initializing test fixtures?

draffensperger:
Thanks for documenting that!

As for my specific issue, I was getting an error like this:
```
[ { Error: ENOENT: no such file or directory, mkdir '~/github/cloud-trace-nodejs/build/test/plugins/fixtures/google-cloud-datastore1/node_modules/grpc/node_modules/tar-pack/test/fixtures'
   errno: -2,
   code: 'ENOENT',
   syscall: 'mkdir',
   path: '~/github/cloud-trace-nodejs/build/test/plugins/fixtures/google-cloud-datastore1/node_modules/grpc/node_modules/tar-pack/test/fixtures' } ]
```

It turns out that I still had some untracked directories in `test/plugins/fixture` from when I had cloned a previous version of the repo. I deleted those directories, but kept the `.pem` files and then I was able to run the tests. I also needed to run `./bin/docker-trace.sh start` to start a Docker container with the databases for the test.

src/span-data.ts
draffensperger: Done. Thanks for the tip.

src/span-data.ts
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import * as crypto from 'crypto';
```
kjin: I'm wondering if the use of cryptographically-secure random bytes is necessary here. Would we be able to use Math.random() instead?

test/test-trace-header-context.ts
```diff
@@ -101,18 +101,28 @@ describe('test-trace-header-context', function() {
...
-        ';o=1');
+      var receivedTraceContext =
+          req.headers[Constants.TRACE_CONTEXT_HEADER_NAME];
+      var receivedTraceId = receivedTraceContext.split('/')[0];
+      var receivedSpanIdAndOptions =
+          receivedTraceContext.split('/')[1].split(';');
+      var receivedSpanId = receivedSpanIdAndOptions[0];
+      var receivedTraceOptions = receivedSpanIdAndOptions[1];
+      // Trace ID and trace options should be the same in sender and receiver.
+      assert.equal(receivedTraceId, sentTraceId);
```
kjin: Nit: Use `strictEqual` and `notStrictEqual` (I know we weren't using this before, but while you're making this change it would be nice)

draffensperger: Done.

src/span-data.ts
draffensperger: Sure, I switched this to use `Math.random()`.

My thinking with the crypto random was that if two Node processes run on a machine with a common clock (e.g. containers on the same VM), perhaps their Math.random would be seeded to the same number and they would produce colliding span IDs. But I think my concern is probably unwarranted and random number generators are probably smarter these days.

src/span-data.ts
```diff
@@ -43,21 +42,12 @@ interface StackFrame {
...
-// Use the faster crypto.randomFillSync when available (Node 7+) falling back to
-// using crypto.randomBytes.
-const spanIdBuffer = Buffer.alloc(SPAN_ID_RANDOM_BYTES);
-const randomFillSync = crypto.randomFillSync;
-const randomBytes = crypto.randomBytes;
-const spanRandomBuffer = randomFillSync ?
-    () => randomFillSync(spanIdBuffer) :
-    () => randomBytes(SPAN_ID_RANDOM_BYTES);
+// This evaluates to a large whole number such that any fractional number
+// between 0 and 1 multiplied by it will always give a whole number.
```
kjin: Interesting 🤔  do you have a source for this? I'm curious

Also, do you mean "any fractional number generated by `Math.random`"?

draffensperger: So it has to do with the fact that the number is bigger than 1 / Number.EPSILON, but I think it's clearer to have a more meaningful maximum span ID. I changed it to just be the largest JS number that's less than 1^64-1, which is the maximum span ID as it's a 64-bit unsigned integer.

src/span-data.ts
```diff
@@ -42,8 +42,15 @@ interface StackFrame {
...
-// Auto-incrementing integer
-let uid = 1;
+// Span IDs in the X-Cloud-Trace-Context header are 64-bit unsigned integers
+// This is the largest JavaScript number that is less than the max unsigned
+// 64-bit int, that is less than 2^64-1.
+const SPAN_ID_MAX = 18446744073709550000;
+
+function randomSpanId() {
+  // No rounding is needed because SPAN_ID_MAX is bigger than 1 / Number.EPSILON
+  return String(Math.random() * SPAN_ID_MAX);
```
kjin: It might be in practice that `Math.random()` generates numbers larger than `Number.EPSILON` but I don't think this is necessarily guaranteed, as there can certainly be numbers smaller than `Number.EPSILON` (take Number.MIN_VALUE for example).

I did an experiment [here](https://gist.github.com/kjin/321ce5514a0a393f35c6aa71ef038789) and observed some interesting results:

* Stringifying a number is probably the bottleneck, not rounding
* `crypto.randomFillSync` is ~5 times slower for 10000000 requests

But given that even each `randomFillSync` call takes about ~2 microseconds we may be over-optimizing in the first place. My initial comment on whether we should use `crypto` was less about saying that we should use `Math.random`, and more wondering about whether the slowdown justified the increase in randomness. Given the uncertainty about how multiplying a decimal number by a large value gives a uniform distribution of integers (at least it's uncertain in my mind) it may actually be better to use `crypto.randomFillSync`. What do you think? (Sorry if it seems like I'm flip-flopping, that wasn't my intention)

draffensperger: So `Math.random()` generates numbers between 0 and 1, so it almost always generates numbers larger than `Number.EPSILON`. Number.MIN_VALUE is negative. You're right though in that the distribution won't cover all possible numbers as by multiplying by Math.random though.

Thanks for benchmarking `randomFillSync` and good to know that it's not to slow. I'll switch it back to that.

I think ideally we would pull in a 64-bit integer library for Node and then generate all 8 random bytes, but I think 6 random bytes is OK (even given that the `number` double precision floats don't represent all of them exactly so it will be less than 6 bytes of randomness I think).

But all in all, much more random than an incrementing counter this is replacing.

draffensperger: OK, update done. I guess to clarify, I'm not saying we should for sure pull in an 64-bit integer library and do the full 8-bytes of crypto randomness (or even pull in a separate random number library!), because I think the 6-bytes of crypto randomness with some rounding error is fine overall and better than more dependencies.

kjin: Sounds good. For future reference -- `Number.MIN_VALUE = 5e-324`, the smallest positive number representable in JS.

draffensperger: Sorry - my bad on that. I was confused by the negative exponent I think and was not looking as deeply into this as I could have. Thanks for clarifying!

PR #655, opened 2018-01-18
doc: add details about running tests locally in CONTRIBUTING.md

kjin:


CONTRIBUTING.md
```diff
@@ -1,3 +1,50 @@
...
+`npm run script` is an alias for `ts-node -p ./scripts ./scripts/index.ts`.
+
+For example, to compile all scripts and then initialize test fixtures, the command to use would be one of:
+
+```bash
+# Option 1
+ts-node -p ./scripts ./scripts/index.ts npm-compile-all init-test-fixtures
+
+# Option 2
+npm run script npm-compile-all init-test-fixtures
```
jinwoo: Don't you need `--` between `script` and the rest of the arguments?

CONTRIBUTING.md
```diff
@@ -1,3 +1,50 @@
...
+
+# Option 3
+npm run compile-all
+npm run script init-test-fixtures
```
jinwoo: Same question here.

CONTRIBUTING.md
```diff
@@ -1,3 +1,50 @@
...
+npm run compile
+npm run init-test-fixtures
+npm test # Or "npm run script run-unit-tests"
+```
+
+A convenient one-liner for this (after `npm install`) is:
+
+```bash
+npm install
+GCLOUD_TRACE_NEW_CONTEXT=1 npm run script npm-compile-all init-test-fixtures run-unit-tests
```
jinwoo: And here.

CONTRIBUTING.md
kjin: We don't need it. I believe `npm run x` is mapped to its underlying script first, and then the rest of the arguments are passed to the result of that evaluation.

jinwoo: OK. Maybe I was misunderstanding something.

PR #658, opened 2018-01-23
test: replace web framework tracing tests

### LINKED ISSUE #617: hapi unit tests are flaky
While running tests for a PR that is unrelated to hapi, its unit tests are frequently failing.

kjin:
__R__

This change removes `test/plugins/test-trace-{express|connect|hapi|koa|restify}.ts`  in favor of a new `test/test-trace-web-frameworks.ts` which features most of the test cases, but run once per framework-version pair. The benefits of this are:

- Less duplication of code, which also means
  - Less mistake-prone
  - Minimal work required to add new test cases
  - Minimal work required to add testing for new frameworks/versions
  - It's easier to read and understand (in my opinion)
- It doesn't take as long (due to the different selection of test cases/versions)
- Re-write includes types
- Re-write includes liberal use of `async/await`
- Re-write reduces flakiness (by making abort request and span timing unit tests more forgiving)

The new list of test cases are derived from the following table of test cases (marked in bold) in each of the individual removed files:

  | hapi | connect | express | koa | restify
-- | -- | -- | -- | -- | --
__accurately measure get time (1 handler)__ | x | x | x | x | x
__accurately measure get time (2 handlers)__ |   | x |   |   | x
accurately measure get time (regex path) |   |   |   |   | x
accurately measure get time until res.end |   | x |   |   |  
accurately measure route time |   |   | x |   |  
accurately measure post time | x |   | x |   | x
accurately measure put time |   |   | x |   |  
accurately measure param time |   |   | x |   |  
accurately measure get time for a single middleware |   |   | x |   |  
accurately measure get time with custom handlers | x |   |   |   |  
accurately measure get time with custom hapi plugins | x |   |   |   |  
accurately measure get time with hapi server.after callback | x |   |   |   |  
accurately measure get time with custom hapi extension __(this is covered with the 2 handlers case)__ | x |   |   |   |  
__have proper labels__ | x | x | x | x | x
__remove trace frames from stack__ | x | x | x | x | x
__not include query parameters in span name__ | x | x | x | x | x
__set trace context__ | x | x | x | x | x
__not trace ignored URLs__ | x | x | x | x | x
__end span on abort__ | x | x | x | x | x
__handle errors from get__ |   | x | x |   |  
handle errors from middleware |   |   | x |   |  
__propagates context__ |   |   |   | x |  

I felt that it wasn't necessary to measure time for various HTTP verbs, as from a web framework serving perspective they are handled identically.

In addition, unit tests are no longer run for each version of `hapi` between 8 - 16; instead, they are run only for 8, 12, 15, and 16. (Fixes #617)

Also, I updated `test-trace-cluster.ts` to remove `doRequest` from `common.ts`. This actually helped expose an issue: #659

As tests use async/await I added a new compile command which compiles to `es2017` for Node 8+ and `es2015` for Node 4/6.


scripts/index.ts
```diff
@@ -37,79 +38,83 @@ async function run(steps: string[]) {
...
-      case 'decrypt-service-account-credentials':
-        const {
-          TRACE_SYSTEM_TEST_ENCRYPTED_CREDENTIALS_KEY: key,
-          TRACE_SYSTEM_TEST_ENCRYPTED_CREDENTIALS_IV: iv,
-        } = process.env;
-      
-        if (!key || !iv) {
-          console.log('> Environment insufficient to decrypt service account credentials');
+    } else if (step.startsWith('compile-')) {
+      const [_literalCompile, languageLevel, strict] = step.split('-');
```
jinwoo: `_literalCompile` is not used, right?

```ts
const [, languageLevel, strict] = step.split('-');
```
?

src/cls.ts
```diff
@@ -23,13 +23,16 @@ export type RootContext = SpanData|{} /* null span */|null;
...
-    require('continuation-local-storage');
+const useAsyncHooks: boolean = semver.satisfies(process.version, '>=8') &&
+    !!process.env.GCLOUD_TRACE_NEW_CONTEXT;
+
+const cls: typeof CLS =
+    useAsyncHooks ? require('./cls-ah') : require('continuation-local-storage');
 
 const TRACE_NAMESPACE = 'com.google.cloud.trace';
 
+export const ROOT_SPAN_STACK_OFFSET = useAsyncHooks ? 0 : 2;
```
jinwoo: What does this mean? Can you add some comments?

src/plugins/plugin-express.ts
```diff
@@ -55,7 +55,7 @@ function patchModuleRoot(express: Express4Module, api: PluginTypes.TraceAgent) {
       api.wrapEmitter(req);
       api.wrapEmitter(res);
 
-      const url = `${req.protocol}://${req.hostname}${req.originalUrl}`;
+      const url = `${req.protocol}://${req.headers.host}${req.originalUrl}`;
```
jinwoo: Curious. What's the diff between `req.hostname` and `req.headers.host`?

test/test-trace-web-frameworks.ts
```diff
@@ -0,0 +1,323 @@
+/**
+ * Copyright 2018 Google Inc. All Rights Reserved.
```
jinwoo: Nit. Should be Google "LLC" now. I think only newly added files can be updated.

test/trace.ts
```diff
@@ -0,0 +1,125 @@
...
+import {LabelObject, TraceWriter, traceWriter, TraceWriterConfig, TraceWriterSingletonConfig} from '../src/trace-writer';
+
+export {Config, PluginTypes};
+
+const traces = new Map<string, TraceSpan[]>();
+const allSpans: TraceSpan[] = [];
+
+class TestTraceWriter extends TraceWriter {
+  constructor(logger: common.Logger, config: TraceWriterConfig) {
+    super(logger, config);
```
jinwoo: I'm not sure but doesn't it just work if this constructor definition is removed?

test/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Constants
+ */
+
+// The duration to give a span when it's important
+export const DEFAULT_SPAN_DURATION = 200;
+// The acceptable window of variation in span duration
+export const ASSERT_SPAN_TIME_TOLERANCE_MS = 40;
+export const SERVER_SPAN_PREDICATE = (span: TraceSpan) =>
```
jinwoo: Any reason to use arrow functions for this? And it looks weird to use all-capital for a function name.

I find it easier to read to use a regular function definition for top-level functions.

test/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+    span.kind === 'RPC_SERVER' && span.name !== 'outer';
+
+/**
+ * Helper Functions
+ */
+
+// Convenience function that, when awaited, stalls for a given duration of time
+export const wait = (ms: number) =>
```
jinwoo: The same question for `wait` and `assertSpanDuration`.

scripts/index.ts
kjin: I wasn't aware of this syntax... added :)

src/cls.ts
kjin: Stack traces are captured when a root span is started. Because the stack trace height varies on the context propagation mechanism, to keep published stack traces uniform we need to remove the top-most frames when using the c-l-s module. Keep track of this number here.

^ Adding the above as a comment

src/plugins/plugin-express.ts
kjin: I believe `headers.host` includes the port number, but `hostname` doesn't.

test/test-trace-web-frameworks.ts
kjin: Done

test/trace.ts
kjin: Yep -- did not know this. Updated

test/utils.ts
kjin: Personally I've been using arrows unless they're class members -- but it's true that it's less readable. I've changed this (and the other functions)

jinwoo: Thanks. I think Google's TS style guide recommends that we prefer function declarations in this case:

> Prefer function declarations over assigning a function expression into a local variable (const x = function() {...};). TypeScript already disallows rebinding functions, so preventing overwriting a function declaration by using const is unnecessary.

PR #669, opened 2018-02-10
fix(package): update @google-cloud/common to version 0.16.1

jinwoo:
Closes #665

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "typescript": "~2.6.2"
   },
   "dependencies": {
-    "@google-cloud/common": "^0.15.1",
+    "@google-cloud/common": "^0.16.1",
     "continuation-local-storage": "^3.2.1",
     "extend": "^3.0.0",
     "gcp-metadata": "^0.4.1",


jinwoo:
@kjin AppVeyor tests are failing while CircleCI tests pass. Do you know why?

jinwoo:
This PR is not needed anymore because v0.15.2 of `@google-cloud/common` has been released, which fixes the security issue around `log-driver`.

But the AppVeyor test failure doesn't seem related to this change and should be looked at separately (/cc @kjin).

Closing this.

kjin:
@jinwoo I think it's because error message contents were changed in gRPC 1.9. From the Node 4 run it seems like there is a change detector test that specifically checks message equality (I've removed in another PR just now -- I think just checking the code should be fine).

It's a little odd that it didn't fail for Node 6, but I think that's attributable to fixtures being cached.

jinwoo:
@kjin Then why would CircleCI tests have passed with all versions of Node? Are CircleCI tests different from what is tested on AppVeyor?

PR #672, opened 2018-02-17
fix(package): update gcp-metadata to version 0.6.3

kjin:
Closes #689
Closes #697
Closes #682


src/trace-writer.ts
```diff
@@ -173,60 +174,49 @@ export class TraceWriter extends common.Service {
...
-          }
-          this.logger.info('Acquired ProjectId from metadata: ' + projectId);
-          this.config.projectId = projectId;
-          callback(null, projectId);
+    gcpMetadata.project({property: 'project-id', headers})
+        .then((res) => {
+          cb(null, res.data);  // project ID
+        })
+        .catch((err) => {
+          cb(err);
```
jinwoo: Previously the error message had detailed information. Does the new gcp-metadata provide similar information in the error?

test/nocks.ts
```diff
@@ -31,25 +31,38 @@ export function oauth2<T extends {} = {}>(validator?: (body: T) => boolean):
...
+export function projectId(status: number|(() => string), reply?: () => string) {
+  if (typeof status === 'function') {
+    reply = status;
+    status = 200;
+  }
   return nock('http://metadata.google.internal')
       .get('/computeMetadata/v1/project/project-id')
       .once()
-      .reply(reply);
+      .reply(200, reply, {'Metadata-Flavor': 'Google'});
```
jinwoo: I think you should use `status` instead of 200, right? Same for similar pieces of code below.

package.json
```diff
@@ -86,8 +86,8 @@
     "nyc": "^11.4.1",
     "once": "^1.4.0",
     "pify": "^3.0.0",
-    "proxyquire": "^1.4.0",
     "request": "^2.83.0",
+    "retry-axios": "^0.3.2",
```
jinwoo: Should this be moved to devDeps?

kjin: I think it already is in `devDeps`.

jinwoo: Ah, I was mistaken by the github UI. Sorry :)

src/trace-writer.ts
kjin: It displays the status code. I added back the "This may be due to a temporary server error; please try again later." part.

test/nocks.ts
kjin: Done

src/trace-writer.ts
```diff
@@ -15,8 +15,10 @@
  */
 
 import * as common from '@google-cloud/common';
+import {AxiosError} from 'axios';
```
jinwoo: Ugh.. This is an abstraction leak from gcp-metadata, by bypassing axios errors to the client code. /cc @JustinBeckwith

src/trace-writer.ts
```diff
@@ -173,60 +175,53 @@ export class TraceWriter extends common.Service {
...
-            }
-          }
-          if (err || !projectId) {
-            // We shouldn't observe a falsey projectId if there's no error
-            err = err || new Error('Project ID missing.');
-            callback(err);
-            return;
+    gcpMetadata.project({property: 'project-id', headers})
+        .then((res) => {
+          cb(null, res.data);  // project ID
```
jinwoo: Old code handled the case where the falsey project id is given. Don't we need to handle that case anymore?

kjin: I think the mention of "falsey projectId" was to satisfy the TypeScript compiler, since it could statically understand that there would either be an error or a projectId, but not both. I don't think this is needed anymore. If we get a falsey project ID here, it should be the responsibility of `gcp-metadata` to fix that.

src/trace-writer.ts
JustinBeckwith: Yeah, little bit.  What would you suggest?

jinwoo: An obvious answer would be "catch errors at the API boundaries and transform them into gcp-metadata's own errors". But I'm not sure how simple it'd be to do that.

I think at least we should document what kind of fields are available in the error thrown by gcp-metadata, and make it not be broken when we switch axios to something else, if ever, later.

PR #680, opened 2018-03-02
feat: ensure spans are non-null

### LINKED ISSUE #656: Create New Spans In gRPC Server RPC Methods
Hi Everyone,

I am trying to create additional trace spans inside my gRPC rpc methods. I reviewed the API docs and examples but have not seen any firm guidance or working implementation. When I do call trace#createChildSpan, a null is returned. What am I doing wrong here? Here is an abbreviated code block from my server.js file:

`const traceApi = require("@google-cloud/trace-agent").start({
  projectId: "festivals-183900",
  keyFilename: googleSvcAccount
});
server.addService(smtpTemplateProto.templatesPackage.SmtpTemplateSvc.service, {
addBindable: (call: any, cb: any) => {
    let clientCn: Client;
    let data: Object;
    let rollBack: boolean = false;
    let spanAddBindable: any;
    try {
      // create spanAddBindable span
      spanAddBindable = traceApi.createChildSpan({
        name: "Svc.BindableController#addBindable"
      });
      console.log(spanAddBindable);   <== evals as null!
      spanAddBindable.endSpan();
  }
  catch(err) {}
}
}
`

kjin:
Fixes #656

This PR changes the public API so that spans are never null (in `runInRootSpan` or `createChildSpan`). Instead, they now expose a `type` field which represents whether they are real or "virtual spans". The code docs for `SpanDataType` in [this file](https://github.com/kjin/cloud-trace-nodejs/blob/new-span-data/src/constants.ts) should explain the purpose of `type`.

Open questions:
1. Is this semver-major?
2. Does `SpanData#type` need to have distinct `ROOT` and `CHILD` values?

# Example

```js
import * as trace from '.';
import { createReadStream } from 'fs';

const traceApi = trace.start();

// original
{
  function doWork(onChunk) {
    const stream = createReadStream('large-file', 'utf8');
    traceApi.runInRootSpan({ name: 'outer' }, rootSpan => {
      if (rootSpan) {
        // only attempt to propagate context if root span isn't null
        traceApi.wrapEmitter(stream);
      }
      let count = 0;
      stream.on('data', onChunk);
      stream.on('end', () => {
        // rootSpan could be null
        if (rootSpan) {
          rootSpan.endSpan();
        }
      });
    });
  }

  doWork((data) => {
    const childSpan = traceApi.createChildSpan({ name: `inner-${count++}` });
    // childSpan could be null
    if (childSpan) {
      childSpan.addLabel('data-length', data.length);
      childSpan.endSpan();
    } else {
      // Why not? Unclear
    }
  });
}

// changed
{
  function doWork(onChunk) {
    const stream = createReadStream('large-file', 'utf8');
    traceApi.runInRootSpan({ name: 'outer' }, rootSpan => {
      if (rootSpan.type === traceApi.spanTypes.ROOT) {
        // only attempt to propagate context if this span is really being traced
        traceApi.wrapEmitter(stream);
      }
      let count = 0;
      stream.on('data', onChunk);
      stream.on('end', () => {
        rootSpan.endSpan();
      });
    });
  }

  doWork((data) => {
    const childSpan = traceApi.createChildSpan({ name: `inner-${count++}` });
    childSpan.addLabel('data-length', data.length);
    childSpan.endSpan();
    // Log if context was lost somehow
    if (childSpan.type === traceApi.spanTypes.UNCORRELATED) {
      console.log('we must have lost context');
    }
  });
}
```


ofrobots:
This would be semver major if existing user code *must* be changed in order to continue working. I haven't reviewed due to the DNR warning at the top, but my suspicion is that this is not semver major.

kjin:
@ofrobots I'll add an example of how it changed in the abstract shortly

src/cls.ts
```diff
@@ -51,16 +70,47 @@ export function getNamespace(): CLS.Namespace {
...
+ * sense to pass UNCORRELATED_SPAN, which is a value specifically reserved for
+ * when getRootContext is known to give an unusable value.
+ */
 export function setRootContext(rootContext: RootContext): void {
   getNamespace().set('root', rootContext);
 }
+
+// This is only used in tests (and is temporary), so it doesn't apply in the
+// comment in getRootContext about the possible values of namespace.get('root').
+// It's functionally identical to setRootContext(null).
```
ofrobots: *Its

src/plugins/plugin-grpc.ts
```diff
@@ -70,7 +70,7 @@ function patchClient(client, api) {
     return function clientMethodTrace() {
       // The span name will be of form "grpc:/[Service]/[MethodName]".
       var span = api.createChildSpan({ name: 'grpc:' + method.path });
-      if (!span) {
+      if (span.type !== api.spanTypes.CHILD) {
```
ofrobots: I don't really like this test.

Previously there was a simple way to check if the operation (`createChildSpan`) was succesful. Now we have to check if the operation produced a span of the correct type. This is burdensome for users.

src/span-data.ts
```diff
@@ -42,36 +41,49 @@ function randomSpanId() {
...
+   *                   when collecting the stack trace.
    */
   constructor(
-      readonly trace: Trace, spanName: string, parentSpanId: string,
-      private readonly isRoot: boolean, skipFrames: number) {
-    spanName =
-        traceUtil.truncate(spanName, Constants.TRACE_SERVICE_SPAN_NAME_LIMIT);
-    this.span = new TraceSpan(spanName, randomSpanId(), parentSpanId);
-    trace.spans.push(this.span);
+      trace: Trace, spanName: string, parentSpanId: string,
```
ofrobots: Why did you change the inline declaration of the `trace` property?

src/cls.ts
kjin: "It's" is correct: "It is functionally identical to (...)."

src/plugins/plugin-grpc.ts
kjin: This is a check that only needs to be done in plugins, where the expressiveness of distinguishing between untraced and uncorrelated requests can be useful. Most users should never need do to this check.

That being said, it might be nice to have a three-value enum instead with numeric values: `{ REAL = 0, UNTRACED = 1, UNCORRELATED = 2 }`. That way we can have a binary statement to check that spans are non-null: `if (!span.type)` and a comparison if we need to know why it's null: `if (span.type === UNTRACED)` (though obviously the field name should be changed from `type` to something more descriptive). What do you think?

src/span-data.ts
kjin: It looks cleaner. I forgot that the TS style guide prefers inline... will change.

src/cls.ts
```diff
@@ -17,9 +17,28 @@
...
+ * specifying whether this is on purpose or by accident, respectively).
+ *
+ * When we store an actual root span, the only information we need is its
+ * current trace/span fields.
+ */
+export type RootContext = ({
+  readonly span: TraceSpan;
+  readonly trace: Trace;
+  readonly type: SpanDataType.ROOT;
+}|{
```
ofrobots: nit: it would be good to name these anonymous types. Something like `RealRootContext` and `PhantomRootContext`?

src/cls.ts
ofrobots: you're right.

src/plugin-types.ts
```diff
@@ -43,12 +44,19 @@ export interface SpanData {
...
+   */
+  readonly type: SpanDataType;
+
   /**
    * Ends the span. This method should only be called once.
    */
   endSpan(): void;
 }
 
+// export type SpanData = RootSpanData|ChildSpanData;
```
ofrobots: remove this?

src/plugin-types.ts
```diff
@@ -99,7 +107,7 @@ export interface TraceAgent {
    * null as an argument.
    * @returns The return value of calling fn.
    */
-  runInRootSpan<T>(options: RootSpanOptions, fn: (span: SpanData|null) => T): T;
+  runInRootSpan<T>(options: RootSpanOptions, fn: (span: SpanData) => T): T;
```
ofrobots: I am pretty sure that this type change is not semver major. Did you think about this as well, and agree?

src/span-data.ts
ofrobots: Avoiding style changes in big PRs also makes the reviewers' life easy. If you really want them, put them in a standalone commit for ease of reviewing.

src/span-data.ts
```diff
@@ -83,18 +95,75 @@ export class SpanData implements SpanDataInterface {
...
+    <T extends SpanDataType>(spanType: T): SpanData&{readonly type: T} => {
+      return Object.freeze({
+        type: spanType,
+        getTraceContext() {
+          return '';
+        },
+        // tslint:disable-next-line:no-any
+        addLabel(key: string, value: any) {},
+        endSpan() {}
+      } as SpanData & {readonly type: T});
```
ofrobots: is this cast necessary?

src/span-data.ts
```diff
@@ -83,18 +95,75 @@ export class SpanData implements SpanDataInterface {
...
+  constructor(
+      trace: Trace, spanName: string, parentSpanId: string,
+      skipFrames: number) {
+    super(trace, spanName, parentSpanId, skipFrames);
+    this.span.kind = SpanKind.RPC_CLIENT;
+  }
+}
+
+// Helper function to generate static virtual trace spans.
+const createDummySpanData =
```
ofrobots: nit: this would be more readable as:
```ts
function createDummySpanData<T extends SpanDataType>(spanType: T): ... {
  // ...
}
```

Also, `Phantom` is probably a better phrase than `Dummy`.

src/plugins/plugin-grpc.ts
ofrobots: I do expect third party authors to write plugins. And the test is still 'wordy'. By reading the test, it is not obvious to me why the test is here. Logically the test the plugin authors want to write is this:

```
if (this request is being sampled) {
  do something
}
```

Instead we see a test that compares the `type` of a span. This doesn't explain what is happening or why comparing the type achieves it.

Perhaps you can add a boolean returning function on the API that would lead to more descriptive tests.

The idea of using integer valued enums with three numeric values but using the fact that `REAL = 0` as a short circuit sounds too hacky to me.

ofrobots:
Note the example you provide in OP
```
      if (rootSpan.type === traceApi.spanTypes.ROOT) {
        // only attempt to propagate context if this span is really being traced
        traceApi.wrapEmitter(stream);
      }
```

The test should be self-documenting. Right now you needed a comment to explain what the test is achieving.

src/span-data.ts
kjin: Done

src/span-data.ts
kjin: Yes, but curiously this works:

```ts
Object.freeze(Object.assign(
      {
        getTraceContext() {
          return '';
        },
        // tslint:disable-next-line:no-any
        addLabel(key: string, value: any) {},
        endSpan() {}
      },
      {type: spanType}))
```

src/plugin-types.ts
kjin: Done

src/plugin-types.ts
kjin: I guess the semver-major change is that if someone had significant logic branching based on the root span being null, then that code path will no longer execute. I don't expect users of the custom tracing API to encounter this at all, but I did have to change every plugin.

src/cls.ts
kjin: Done

kjin:
@ofrobots PTAL

src/cls.ts
```diff
@@ -21,6 +21,15 @@ import {SpanDataType} from './constants';
 import {UNCORRELATED_SPAN, UNTRACED_SPAN} from './span-data';
 import {Trace, TraceSpan} from './trace';
 
+export type RealRootContext = {
+  readonly span: TraceSpan; readonly trace: Trace; readonly type: SpanDataType
```
ofrobots: nit: Do these need to be on the same line?

src/trace.ts
```diff
@@ -30,7 +30,7 @@ export interface TraceSpan {
   kind: SpanKind;
   name: string;
   spanId: string;
-  parentSpanId: string;
+  parentSpanId?: string;
```
ofrobots: Why did this field become optional?

kjin: To match the [API data model](https://cloud.google.com/trace/docs/reference/v1/rest/v1/projects.traces#TraceSpan) more closely. This field is never read except in tests so I don't think it really makes a difference.

src/cls.ts
kjin: `clang-format` says: Yep

jinwoo: You should prefer `interface` over `type`. clang-format will format it more beautifully.

kjin: You're right. Updated

kjin:
Node 4 CI will pass in ~10 min (tests passed, but SSH connection is open)

PR #686, opened 2018-03-08
fix: rewrite plugin loader

kjin:
This change re-writes the Plugin Loader entirely.

The existing Plugin Loader code exists as a set of module-scope functions and variables. While there aren’t any major problems here, it’s worth re-writing it for the following reasons:
* To make it more idiomatic to TypeScript, especially now that it has type annotations
* To make the code more readable
* To expose a more testable interface (and improve unit tests)
* To depend on a third party module (require-in-the-middle) for monkeypatching
* To make future behavioral changes more easy to implement:
  * Support for multiple plugins for the same module
  * Support for plugins that don’t monkeypatch at all

User-facing changes:
- ~~`Intercept<T>` is no longer a type on the public interface of the Trace Agent; its definition was merged with `Patch<T>`. Arguably, they are all "patches". Users will still be warned if they provide both `patch` and `intercept`.~~ (Edit: This has been reverted.)
- We no longer warn if neither `patch` or `intercept` exist on a `Patch`. This is to prevent needless warnings when a module such as `knex` needs patches for some versions but not others.
- All log messages from the plugin loader have been revised to be more regex-searchable.
- We accept an undocumented `'[core]'` plugin key. A plugin loaded under this key would behave as if the key were any other core module, such as `'http'`. The plan is that at some point, we would use the `'[core]'` key to patch core modules rather than `'http'`, because it's somewhat misleading that `'http'` also patches `'https'`.

(All of these are up for discussion.)

### Notes for Reviewers

* This change used to also introduce multiple plugins for a single module. I decided against that since it's (1) too big of a change and (2) unclear whether this makes sense for an end-user. Just something to keep in mind if you see something weird like storing plugins as an array or two layers of version-checks (you should make a note of it if it doesn't make sense).
* The computed PR diff for `trace-plugin-loader.ts` is completely useless. I recommend looking at the commits separately -- the first commit is just the removal of the old file.


src/trace-plugin-loader.ts
```diff
@@ -16,278 +16,485 @@
...
-  return !!(obj as InternalPatch<T>).patch;
-}
+/**
+ * A class that represents wrapper logic around a user-provided plugin object
+ * to be applied to a single module.
+ */
+export class ModulePluginWrapper implements PluginWrapper {
+  // Sentinel value to indicate that a plugin has not been loaded into memory
+  // yet.
+  private static readonly NOT_LOADED: Plugin = [];
```
jinwoo: Would it make sense to export this from plugin-types.ts? Probably after naming it `NULL_PLUGIN` or something.

src/trace-plugin-loader.ts
```diff
@@ -16,278 +16,485 @@
...
-            ' was loaded before the trace agent was initialized.');
-        break;
+  isSupported(version: string): boolean {
+    // The plugin is lazily loaded here.
+    const plugin = this.getPluginExportedValue();
+    // Count the number of Patch/Intercept objects with compatible version
+    // ranges
+    let numFiles = 0;
+    plugin.forEach(instrumentation => {
+      const postLoadVersions = instrumentation.versions;
```
jinwoo: I'm confused, Why is this variable named *post*LoadVersions?

src/trace-plugin-loader.ts
```diff
@@ -16,278 +16,485 @@
...
+        numFiles++;
       }
+    });
+    // We consider a module to be unsupported if there are no Patch/Intercept
+    // objects with compatible version ranges at all.
+    return numFiles > 0;
+  }
+
+  unapplyAll() {
+    this.unpatchFns.reverse().forEach(fn => fn());
```
jinwoo: It's probably worthwhile to add a comment that unpatch should be done in the reverse order.

src/trace-plugin-loader.ts
```diff
@@ -16,278 +16,485 @@
...
+          `PluginWrapper#applyPlugin: [${logString}] Applying plugin.`);
+      if (instrumentation.patch) {
+        instrumentation.patch(
+            exportedValue, this.createTraceAgentInstance(logString));
+        // Queue a function to run if the plugin gets disabled.
+        if (instrumentation.unpatch) {
+          this.unpatchFns.push(() => {
+            this.logger.info(
+                `PluginWrapper#unapplyAll: [${logString}] Unpatching file.`);
+            instrumentation.unpatch!(exportedValue);
```
jinwoo: The compiler is actually correctly detecting a nullable error here. If the `unpatch` field is cleared (or changed) after `applyPlugin()` is called, this code will do a wrong thing. You should create a local variable that is closured over from this code.
```ts
if (instrumentation.unpatch) {
  const unpatch = instrumentation.unpatch;
  this.unpatchFns.push(() => {
    // ...
    unpatch(exportedValue);
  });
}
```

src/trace-plugin-loader.ts
```diff
@@ -16,278 +16,485 @@
...
+    return traceApi;
   }
 }
 
-export function activate(logger: Logger, config: PluginLoaderConfig): void {
-  if (activated) {
-    logger.error('Plugins activated more than once.');
-    return;
+/**
+ * A class that represents wrapper logic on top of plugins that patch core
```
jinwoo: What does a core module mean? Does it mean a Node's builtin module?

src/trace-plugin-loader.ts
```diff
@@ -16,278 +16,485 @@
...
+  /**
+   * Applies this object's underlying plugin patches to a file, returning the
+   * patched or intercepted value.
+   * @param moduleExports The module exports of the file.
+   * @param file The file path, relative to the module root.
+   * @param version The module version.
+   */
+  applyPlugin<T>(moduleExports: T, file: string, version: string): T {
+    return this.children.reduce((exportedValue, child) => {
+      return child.applyPlugin(exportedValue, file, version);
```
jinwoo: Nit. You can remove the curly braces and `return`.

src/trace-plugin-loader.ts
```diff
@@ -16,278 +16,485 @@
...
+  DEACTIVATED
+}
+
+/**
+ * A class providing functionality to hook into module loading and apply
+ * plugins to enable tracing.
+ */
+export class PluginLoader {
+  // Key on which core modules are stored.
+  static readonly CORE_MODULE = '[core]';
```
jinwoo: Nit. Most of the other fields also can be `readonly`, except `internalState`.

src/trace-plugin-loader.ts
```diff
@@ -16,278 +16,485 @@
...
+  // The current state of the plugin loader.
+  private internalState: PluginLoaderState = PluginLoaderState.NO_HOOK;
+
+  /**
+   * Constructs a new PluginLoader instance.
+   * @param logger The logger to use.
+   * @param config The configuration for this instance.
+   */
+  constructor(logger: Logger, config: PluginLoaderConfig) {
+    this.logger = logger;
```
jinwoo: Nit. This can be deleted by
```ts
constructor(private readonly logger: Logger, ...) {...}
```

src/trace-plugin-loader.ts
```diff
@@ -16,278 +16,485 @@
...
+    const nonCoreModules: string[] = [];
+    // Initialize ALL of the PluginWrapper objects here.
+    // See CorePluginWrapper docs for why core modules are processed
+    // differently.
+    const coreWrapperConfig: CorePluginWrapperOptions = {children: []};
+    Object.keys(config.plugins).forEach(key => {
+      const value = config.plugins[key];
+      // Core module plugins share a common key.
+      const coreModule = builtinModules.indexOf(key) !== -1 ||
+          key === PluginLoader.CORE_MODULE;
```
jinwoo: Isn't it better to do the latter comparison first because `indexOf()` is O(N)?

src/trace-plugin-loader.ts
```diff
@@ -16,278 +16,485 @@
...
-          } else {
-            return null;
-          }
         }
+        return exportedValue;
+      });
+      this.internalState = PluginLoaderState.ACTIVATED;
+      this.logger.info(`PluginLoader#activate: Activated.`);
+    } else if (this.internalState === PluginLoaderState.DEACTIVATED) {
+      throw new Error('Currently cannot re-activate plugin loader.');
```
jinwoo: Why is it so?

src/trace-plugin-loader.ts
```diff
@@ -16,278 +16,485 @@
...
-        function moduleAlreadyPatched(
-            instrumentation: InternalPlugin, moduleRoot: string) {
-          return instrumentation.patches[moduleRoot];
-        }
+  /**
+   * Deactivates the plugin loader, preventing additional plugins from getting
+   * loaded or applied, as well as unpatching any modules for which plugins
+   * specified an unpatching method.
+   */
+  deactivate(): PluginLoader {
```
jinwoo: `deactivate()` is idempotent while `activate()` is not. Would it be better to be consistent?

src/trace-plugin-loader.ts
kjin: > Just something to keep in mind if you see something weird like storing plugins as an array or two layers of version-checks (you should make a note of it if it doesn't make sense).

This is exactly the kind of artifact I was describing in the PR abstract. There used to be pre-load and post-load versions. Thanks for the catch!

src/trace-plugin-loader.ts
```diff
@@ -16,278 +16,485 @@
...
-          }
-          return originalModuleLoad.apply(this, arguments);
-        };
-      });
-}
+  /**
+   * Adds a search path for plugin modules. Intended for testing purposes only.
+   * @param searchPath The path to add.
+   */
+  static setPluginSearchPath(searchPath: string) {
```
jinwoo: I'd use a rather long and ugly name for a testing-only method. Something like `setPluginSearchPathForTestingOnly()`.

src/trace-plugin-loader.ts
```diff
@@ -16,278 +16,485 @@
...
+   * string, returning both (or just the name if it's the main module).
+   * @param moduleStr The module string; in the form of either `${module}` or
+   *   `${module}/${relPath}`
+   */
+  static parseModuleString(moduleStr: string): {name: string, file: string} {
+    // Canonicalize the name by using only '/'.
+    const parts = moduleStr.replace(/\\/g, '/').split('/');
+    let indexOfFile = 1;
+    if (parts[0].startsWith('@')) {  // for @org/package
+      indexOfFile = 2;
```
jinwoo: Just
```ts
const indexOfFile = parts[0].startsWith('@') ? 2 : 1
```
?

test/test-plugin-loader.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2017 Google Inc. All Rights Reserved.
+ * Copyright 2018 Google Inc. All Rights Reserved.
```
jinwoo: While you're changing the date, change 'Inc' to 'LLC' as well?

test/test-plugin-loader.ts
```diff
@@ -13,430 +13,329 @@
...
-    ]);
-    assert.strictEqual(require('module-h').createSentence(),
-      'bad tests don\'t make sense.');
-    // Activate plugin loader
-    pluginLoader.activate(logger, {
-      plugins: {
-        'module-h': 'module-h-plugin'
-      }
+
+    it('doesn\'t unpatch twice', () => {
```
jinwoo: Nit. Better to use a template string instead of escaping quotes.

src/plugin-types.ts
```diff
@@ -180,18 +180,11 @@ export interface TraceAgent {
...
-  file?: string;
-  versions?: string;
   intercept: (module: T, agent: TraceAgent) => T;
+  unpatch: (module: T) => void;
 }
 
-export type Instrumentation<T> = Patch<T>|Intercept<T>;
-
-export type Plugin = Array<Instrumentation<any>>;
+export type Plugin = Array<Partial<Patch<any>>>;
```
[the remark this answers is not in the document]
kjin: Strictly speaking it might not because the type reflects the exported value of a module, which can be anything, including non-objects.

src/trace-plugin-loader.ts
kjin: I'm not sure how it might be used by end users. Its value is only used internally by a class that is not exposed to users.

src/trace-plugin-loader.ts
kjin: Yes, I'll amend the comment to that effect.

src/trace-plugin-loader.ts
jinwoo: My comment was less about usages by end users but more about keeping the abstraction barrier. Here it is assumed that `Plugin` is actually an array. If plugin-types.ts exported `NULL_PLUGIN`, we wouldn't have to know if it's an array or an object or whatever.

Having said that, since other parts of code in this file assume that a plugin is an array, I don't think it buys much.

src/trace-plugin-loader.ts
kjin: There is currently no code path that causes it to be re-activated, because the Trace Agent can't be re-enabled after being disabled. So I didn't bother writing an implementation.

test/test-plugin-loader.ts
kjin: Done

test/test-plugin-loader.ts
kjin: Done

src/trace-plugin-loader.ts
kjin: Done

src/trace-plugin-loader.ts
kjin: Done

src/trace-plugin-loader.ts
kjin: Yep -- done

src/trace-plugin-loader.ts
kjin: Good idea, done

src/trace-plugin-loader.ts
kjin: `NOT_LOADED` is only used privately in `ModulePluginWrapper`; even the `Plugin` field that is initially assigned to it is inaccessible. (IOW even within this file, other classes don't know about it.) It might as well be a symbol -- the only reason it's not is to make type checking simple.

.gitignore
```diff
@@ -1,4 +1,5 @@
 node_modules/
+!test/fixtures/loader/node_modules/
```
DominicKramer: Why  is the `node_modules` directory under the `loader` directory not ignored?

src/plugin-types.ts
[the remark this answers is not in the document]
DominicKramer: I'm not sure if this is the case, but if all of the attributes in `Intercept` are optional, I would suggest just marking them all optional instead of using `Partial` because I think it is easier to understand.

src/plugins/plugin-knex.ts
```diff
@@ -18,7 +18,6 @@
 var shimmer = require('shimmer');
 var util = require('util');
 
-// knex 0.10.x and 0.11.x do not need patching
```
DominicKramer: Maybe it wouldn't hurt to keep this comment near the `versions: '>=0.10 <=0.11'` line below to make it explicit that those versions don't need to be patched.

.gitignore
kjin: They are test fixtures. The directory must be named `node_modules` to adhere to how the Node module loader looks for modules.

src/plugins/plugin-knex.ts
kjin: Done

.gitignore
DominicKramer: Ok that makes sense.

PR #687, opened 2018-03-08
chore: upgrade typescript to 2.7

kjin:


kjin:
@jinwoo Not necessarily, but the code path that reads the value of the nullable fields won't be exercised if `enable` wasn't called first (because we evaluate `this.isActive` first). The `!` annotation is there for this purpose -- to tell the compiler that it's statically known that these values can't be null.

kjin:
@jinwoo For the record, I tried making `isActive` return a type guard `this is { namespace: cls.Namespace }`. It doesn't work, but maybe this is worth opening a bug against TS.

PR #693, opened 2018-03-16
test: stop testing plugin functionality on appveyor

kjin:
The main source of flakiness in AppVeyor CI runs is timeouts stemming from `require()` calls taking up more than 4 seconds (this has been observed to be more than 10 seconds in some runs). The suspected reason for this is that we are running several databases on a VM with already-limited resources.

To circumvent this, we stop testing plugin functionality altogether, making the running of these database services entirely unnecessary. The primary reason for using AppVeyor in the first place remains to catch UNIX-specific code changes that are untested on Windows, and testing plugins may be out of that scope.

This also allows us to stop installing/caching plugin test fixtures, which also costs a significant amount of time. Some tests depend on the `express@4` fixture, but these appear to be testing the Trace Agent itself rather than the express plugin, so they now require the `express` devDependency instead.

The result of this PR should be that AppVeyor CI running times improve significantly, and have a much lower (hopefully 0%) false positive rate. As we work towards making the line between unit and integration tests more well-defined in the future, we should be able to get more code coverage when testing on Windows.

Also, I made a semi-related change to have the system test-specific dependencies in its own package.json. Since AppVeyor doesn't run the system test, there is no need to spend time installing unused devDependencies such as gRPC.

### DIFF
--- a/scripts/run-tests.ts
+++ b/scripts/run-tests.ts
@@ -20,7 +20,7 @@ export async function runTests(options: Options) {
   const includedFiles = ([] as string[])
     .concat(...await Promise.all(includeGlobs.map(glob => globP(glob))));
   // Take the difference
-  const files = includedFiles.filter(i => !excludedFiles.some(e => e === i));
+  const files = includedFiles.filter(i => excludedFiles.indexOf(i) < 0);
   for (const file of files) {
     const moduleAndArgs = [
       ...coverage ? [
--- a/scripts/run-tests.ts
+++ b/scripts/run-tests.ts
@@ -2,21 +2,25 @@ import * as path from 'path';
 import { globP, forkP } from './utils';
 
 export interface Options {
-  globs: string[],
+  includeGlobs: string[],
+  excludeGlobs?: string[],
   rootDir: string,
   coverage?: boolean,
   timeout?: number
 }
 
 export async function runTests(options: Options) {
-  const { globs, rootDir, coverage, timeout } = options;
+  const { includeGlobs, excludeGlobs, rootDir, coverage, timeout } = options;
   function nodule(nodule: string) {
     return path.relative(rootDir, `node_modules/${nodule}`);
   }
   let testNum = 0;
-  const files = ([] as string[])
-    .concat(...await Promise.all(globs.map(glob => globP(glob))))
-    .filter(_ => true);
+  const excludedFiles = ([] as string[])
+    .concat(...await Promise.all((excludeGlobs || []).map(glob => globP(glob))));
+  const includedFiles = ([] as string[])
+    .concat(...await Promise.all(includeGlobs.map(glob => globP(glob))));
+  // Take the difference
+  const files = includedFiles.filter(i => !excludedFiles.some(e => e === i));
   for (const file of files) {
     const moduleAndArgs = [
       ...coverage ? [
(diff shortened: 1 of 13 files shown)


scripts/run-tests.ts
```diff
@@ -2,21 +2,25 @@ import * as path from 'path';
...
-  const { globs, rootDir, coverage, timeout } = options;
+  const { includeGlobs, excludeGlobs, rootDir, coverage, timeout } = options;
   function nodule(nodule: string) {
     return path.relative(rootDir, `node_modules/${nodule}`);
   }
   let testNum = 0;
-  const files = ([] as string[])
-    .concat(...await Promise.all(globs.map(glob => globP(glob))))
-    .filter(_ => true);
+  const excludedFiles = ([] as string[])
```
jinwoo: I think this is better than type-casting:
```ts
const excludedFiles: string[] = [].concat(...);
```

kjin:
Not sure why code coverage decreased, it's only being reported from Circle, in which there shouldn't be any changes.

scripts/run-tests.ts
```diff
@@ -2,21 +2,25 @@ import * as path from 'path';
...
+    .concat(...await Promise.all((excludeGlobs || []).map(glob => globP(glob))));
+  const includedFiles = ([] as string[])
+    .concat(...await Promise.all(includeGlobs.map(glob => globP(glob))));
+  // Take the difference
+  const files = includedFiles.filter(i => !excludedFiles.some(e => e === i));
```
jinwoo: Use `indexOf()` instead of `some()`?
```ts
const files = includedFiles.filter(i => excludedFiles.indexOf(i) < 0);
```

system-test/trace-express.js
```diff
@@ -134,7 +134,7 @@ describe('express + datastore', () => {
       const trace = traces[0];
       if (usingAsyncHooks) {
         assert.equal(trace.spans.length, 3, 'should be 3 spans: parent, child, auth');
-        assert.equal(trace.spans[2].name, 'accounts.google.com');
+        assert.equal(trace.spans[2].name, 'www.googleapis.com');
```
jinwoo: Was this failing before?

jinwoo:
Re: code coverage change, could it be because test files are not excluded from the coverage calculations?

scripts/run-tests.ts
kjin: That syntax doesn't work for TS. `[]` must be known to be a `string[]` first.

An alternative is

```ts
const excludedFiles: string[] = [];
excludedFiles.concat(...);
```

But I think that's a little more cumbersome.

system-test/trace-express.js
kjin: No, it actually didn't fail before, surprisingly. My guess is that it's a strange artifact of npm having picked a different patch-level version of the auth library based on the larger dependency tree that it had to construct when all of the dependencies were listed in a single package.json file.

scripts/run-tests.ts
kjin: Done

kjin:
@jinwoo I thought that, but Circle CI still runs all of the test files and reports them the same way as they did before. I noticed that the lines marked as "not hit" seem to be based on version flags. Also, the codecov difference only surfaced when 3204952 was pushed, which removes gRPC as a devDependency. This is why I think it depends on the order of the Node tests -- it could be that Node 4 used to be the fastest to install and thus reported first, and now it reports last. This is something we need to fix, of course, but I think it's out of scope for this PR.

PR #694, opened 2018-03-17
refactor: externalize singleton accessors from trace writer

kjin:
This change adds a class that abstracts the "singleton" concept and uses it for the TraceWriter singleton. This makes singletons (including the future PluginLoader class) easier to mock.

I recommend looking at each commit separately

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -17,6 +17,7 @@
 import * as fs from 'fs';
 import Module = require('module');
 import * as path from 'path';
+import {Logger} from '@google-cloud/common';  // for types only.
 
 /**
  * Trace API expects stack frames to be a JSON string with the following
@@ -39,6 +40,38 @@ export interface StackFrame {
 interface PackageJson {
   name: string;
   version: string;
+}
+
+export interface ClassOf<T, Config> {
+  new(logger: Logger, config: Config): T;
+  prototype: T;
+  name: string;
+}
+
+/**
+ * An class that provides access to a singleton.
+ * Instances of this type should only be constructed in module scope.
+ */
+export class Singleton<T, Config> {
+  private singleton: T|null = null;
+
+  constructor(private implementation: ClassOf<T, Config>) {}
+
+  create(logger: Logger, config: Config&{forceNewAgent_: boolean}): T {
+    if (!this.singleton || config.forceNewAgent_) {
+      this.singleton = new this.implementation(logger, config);
+      return this.singleton;
+    } else {
+      throw new Error(`${this.implementation.name} has already been created.`);
+    }
+  }
+
+  get(): T {
+    if (!this.singleton) {
+      throw new Error(`${this.implementation.name} has not yet been created.`);
+    }
+    return this.singleton;
+  }
 }
 
 /**
--- a/src/util.ts
+++ b/src/util.ts
@@ -42,22 +42,24 @@ interface PackageJson {
   version: string;
 }
 
-export interface ClassOf<T, Config> {
+export interface Constructor<T, Config> {
   new(logger: Logger, config: Config): T;
   prototype: T;
   name: string;
 }
 
 /**
- * An class that provides access to a singleton.
+ * A class that provides access to a singleton.
+ * We assume that any such singleton is always constructed with two arguments:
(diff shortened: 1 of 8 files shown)


src/util.ts
```diff
@@ -41,6 +42,38 @@ interface PackageJson {
...
 }
 
+export interface ClassOf<T, Config> {
+  new(logger: Logger, config: Config): T;
+  prototype: T;
+  name: string;
+}
+
+/**
+ * An class that provides access to a singleton.
```
jinwoo: Nit. An -> A

src/util.ts
```diff
@@ -41,6 +42,38 @@ interface PackageJson {
...
+ * Instances of this type should only be constructed in module scope.
+ */
+export class Singleton<T, Config> {
+  private singleton: T|null = null;
+
+  constructor(private implementation: ClassOf<T, Config>) {}
```
jinwoo: `private readonly implementation`?

kjin: In a test we specifically want to override the `implementation` field with a dummy class constructor. Being able to write (for example) `traceWriterSingleton['implementation'] = TestTraceWriter` is the easiest way to do this, in my opinion.

src/util.ts
kjin: Done

src/util.ts
jinwoo: I see. How is it possible to overwrite the `implementation` field when it is a private field? Doesn't the compiler prevent that?

src/util.ts
```diff
@@ -41,6 +42,38 @@ interface PackageJson {
...
+  prototype: T;
+  name: string;
+}
+
+/**
+ * An class that provides access to a singleton.
+ * Instances of this type should only be constructed in module scope.
+ */
+export class Singleton<T, Config> {
+  private singleton: T|null = null;
```
ofrobots: suggestion: use a Symbol for better privacy?

test/trace.ts
```diff
@@ -61,9 +61,11 @@ class TestTraceWriter extends TraceWriter {
     });
   }
 }
+setTraceWriterEnabled(false);
 
-let singleton: TraceWriter|null = null;
-disableTraceWriter();
+export function setTraceWriterEnabled(enabled: boolean) {
+  traceWriter['implementation'] = enabled ? TraceWriter : TestTraceWriter;
```
ofrobots: why not `traceWriter.implementation`?

src/util.ts
```diff
@@ -42,22 +42,24 @@ interface PackageJson {
...
  * Instances of this type should only be constructed in module scope.
  */
 export class Singleton<T, Config> {
   private singleton: T|null = null;
 
-  constructor(private implementation: ClassOf<T, Config>) {}
+  constructor(private implementation: Constructor<T, Config>) {}
 
-  create(logger: Logger, config: Config&{forceNewAgent_: boolean}): T {
+  create(logger: Logger, config: Config&{forceNewAgent_?: boolean}): T {
```
ofrobots: `Singleton` seems like a general purpose class, but this signature seems to specialize it for `TraceWriter`, so I am not sure if `Singleton` is the right name here. Do you really care about declaring the arguments here? Can you use `...args` and pass them onto the constructor below.

test/trace.ts
kjin: `traceWriter.implementation` is a private name. TypeScript allows `traceWriter['implementation']` as a work-around for that.

src/util.ts
kjin: We lose a significant amount of type-checking with `...args`. I don't see a need for constructors to be flexible; I believe that a logger and a configuration object should be universally applicable for anything that exists as a singleton (plugin loader, trace writer, possibly also cls in the future) within this library. Any data should be passed through configuration; any callbacks should arguably not be passed in through the constructor.

If it helps, I also wrote the assumption out in a comment that a singleton will need a logger and a config object.

src/util.ts
kjin: Done

kjin:
Node 9 CI is failing only because the way it was rebuilt (through Circle workflows) doesn't allow it to correctly detect whether it should run system tests.

PR #696, opened 2018-03-19
test: don't use exec to test preloaded modules

kjin:
It seems like using `cp.execSync` is a significant source of test flakiness, as the executed command can take more than 4 seconds. The reason for doing this in the past was possibly because it was undesirable to check a global property when all tests ran in a single process. We now run each test in a single process (though if we run tests together again in the future, we will need to ensure that other tests won't interfere with this one).

Also, I added a `TestLogger` class to easily check logger output. This is used in #686 as well.


test/logger.ts
```diff
@@ -0,0 +1,62 @@
...
+type LoggerFunction = (message: any, ...args: any[]) => void;
+
+export class TestLogger implements Logger {
+  private logs:
+      {[k in keyof Logger]:
+           string[]} = {error: [], warn: [], info: [], debug: [], silly: []};
+  private innerLogger = logger({level: logger.LEVELS[PASS_THROUGH_LOG_LEVEL]});
+
+  private makeLoggerFn(logLevel: keyof Logger): LoggerFunction {
+    // TODO(kjin): When we drop support for Node 4, use spread args.
```
jinwoo: Doesn't TS compiler convert to using `arguments`? That's what I'm observing in other projects.

kjin: Possibly only when using `target: 'es5'`. We currently need tests to be compiled to ES6 for generators (granted the use of generators is probably constrained to a single test).

jinwoo: Yes, only when the target is es5.

How does the test that uses generators run for Node 4 then?

It'd be great if the TS compiler provided a way to specify a different target for each Node version, but I don't think it supports that now.

kjin: Node 4 supports generators, so it's alright to compile to ES6 as long as we don't use things like destructuring or spread args.

I agree, the only alternative is having separate config files which isn't very clean.

PR #699, opened 2018-03-23
test: fixup issue stemming from testing http2 on node 9.9

kjin:


test/plugins/test-trace-http2.ts
```diff
@@ -219,6 +219,8 @@ describe('test-trace-http2', () => {
     server.on(
         'stream',
         (s: http2.ServerHttp2Stream&({rstWithInternalError: () => void})) => {
+          // In Node 9.9+, the error handler is not added by default.
+          s.on('error', () => {});
```
jinwoo: What happens when the error handler is not added? Does it crash the process with an unhandled promise rejection or something? Seems like a bug in http2 to me.

kjin: This was introduced in https://github.com/nodejs/node/pull/19232. There is a particular change in the tests that shows that emitting an error event on closing with `NGHTTP2_INTERNAL_ERROR` is intentional: https://github.com/nodejs/node/pull/19232/files#diff-667df0178081ac32b06a004a9bcafb17

DominicKramer: Would it be helpful to have an assert that either verifies that the `error` event was emitted or not (depending on whether or not the test expects an error).

PR #702, opened 2018-03-26
fix: add support for pg 7 changes

kjin:
This PR converts the PostGreSQL tracing plugin to TypeScript, makes changes to support the new API in pg@7, adds a few more tests, and changes the way `value` labels are collected.

* **chore: add pg7 to test fixtures**
  * This changes a JSON file that specifies the list of modules to be installed as test fixtures, adding `pg@7`.
* **refactor: port plugin-pg to typescript**
  * This converts `plugin-pg.ts` to ES6 + type annotations, and installs `pg@7` types. There should be no behavioral change.
* **fix: add support for pg 7**
  * This adds a patch for `pg@7`, based on [source code](https://github.com/brianc/node-postgres) and the [upgrade guide](https://node-postgres.com/guides/upgrading).
  * __Differences in a nutshell__
    * In `pg@6`, an object is _always_ returned by `Client#query`, and is simultaneously a `Promise[Like]`, an `EventEmitter`, and an object `{ text: string; values?: Array, callback?: Function }`.
    * In `pg@7`, the return value of `Client#query` is _either_ a `Promise`, `EventEmitter`, _or_ `undefined`, depending on the input parameters.
* **fix: get pg 6 query values before call**
  * Currently, we get query `values` from the output object of `Client#query`; in `pg@7`, these fields are not guaranteed to exist, so we must gather them from input parameters instead. However, there is a mismatch between `values` gotten from input and output; output `values` are coerced to strings and escaped to avoid SQL injection. So to keep things consistent, we now _also_ get `values` from the inputs to `Client#query` in `pg@6`. (I don't think that these values being unescaped should pose a threat because their values are simply stored, not executed; feel free to dispute that.)
* **test: modify tests**
  * Modify tests to support both v6 and v7, and also add new test cases for previously uncovered scenarios: promises and generic `Submittable`s. _Converting this test to idiomatic TypeScript is out of the scope of this PR._

### DIFF
--- a/src/plugins/plugin-pg.ts
+++ b/src/plugins/plugin-pg.ts
@@ -18,13 +18,13 @@ import {EventEmitter} from 'events';
 import * as shimmer from 'shimmer';
 import {Readable} from 'stream';
 
-import {Patch, Plugin} from '../plugin-types';
+import {Patch, Plugin, SpanData} from '../plugin-types';
 
 import {pg_6, pg_7} from './types';
 
 // TS: Client#query also accepts a callback as a last argument, but TS cannot
 // detect this as it's a dependent type. So we don't specify it here.
-type PG7QueryArguments =
+type ClientQueryArguments =
     [{submit?: Function} & pg_7.QueryConfig]|[string]|[string, {}];
 type PG7QueryReturnValue = (pg_7.QueryConfig&({submit: Function}&EventEmitter)|
                             pg_7.Query)|Promise<pg_7.QueryResult>;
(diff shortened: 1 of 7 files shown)


jinwoo:
Didn't look at the code yet but CI is failing.

src/plugins/plugin-pg.ts
```diff
@@ -13,35 +13,65 @@
...
-            return pgQuery;
+            return query.apply(this, arguments);
           }
-          if (api.enhancedDatabaseReportingEnabled()) {
-            span.addLabel('query', pgQuery.text);
-            if (pgQuery.values) {
-              span.addLabel('values', pgQuery.values);
+          const args: pg_6.QueryReturnValue[] =
+              Array.prototype.slice.call(arguments, 0);
+          if (args.length >= 1) {
```
jinwoo: Nit. Doing the length check before `Array.prototype.slice.call()` would be better when there's no arguments.

src/plugins/plugin-pg.ts
```diff
@@ -60,14 +90,120 @@ module.exports = [
...
+          // - (query: pg.Submittable) => EventEmitter
+          //   - Note: return value is the same as the argument.
+          // - ([*], callback: (err, res: pg.Result) => void) => void
+          // - ([*]) => Promise<pg.Result>
+          // where [*] is one of:
+          // - ...[query: { text: string, values?: Array<any> }]
+          // - ...[text: string, values?: Array<any>]
+          // See: https://node-postgres.com/guides/upgrading
+
+          if (args.length >= 1) {
```
jinwoo: Same here.

src/plugins/plugin-pg.ts
```diff
@@ -60,14 +90,120 @@ module.exports = [
...
+            // Extract query text and values, if needed.
+            if (api.enhancedDatabaseReportingEnabled()) {
```
jinwoo: Maybe we can factor out this block of code and use both for 6 & 7?

src/plugins/plugin-pg.ts
```diff
@@ -60,14 +90,120 @@ module.exports = [
...
+            // optionally adding labels as well.
+            const callback = args[args.length - 1];
+            if (typeof callback === 'function') {
+              args[args.length - 1] = api.wrap((err, res) => {
+                if (api.enhancedDatabaseReportingEnabled()) {
+                  if (err) {
+                    span.addLabel('error', err);
+                  }
+                  if (res) {
+                    span.addLabel('row_count', res.rowCount);
```
jinwoo: These four lines seem reusable both from here and from line 180 below too.

PR #703, opened 2018-03-27
test: rewrite google-gax test and remove datastore test

kjin:
~~Some authentication mechanism changed in client libraries, causing our client library tests to fail. This change comments them out for now, to unblock #702.~~

~~Arguably, we should find a different approach to testing tracing client libraries.~~

This change re-writes the google-gax tracing test to avoid network operations completely, and removes the datastore test (which is not isolated enough to run reliably).


jinwoo:
Are you deleting these tests because they are not simple to fix? How do we make sure things are not broken until new tests are added?

jinwoo:
Hmm... How come does the code coverage not change when these tests are deleted?

kjin:
@jinwoo We currently also have a system test that makes calls against the datastore service. Datastore also uses `google-gax`. This doesn't run on PRs so it can only be so useful, but ultimately I think that's where it should belong.

Actually, I now think I should re-work this PR to be a unit test for `google-gax` that doesn't make network connections. But I don't think we should be keeping that datastore test.

test/plugins/test-trace-google-gax.ts
```diff
@@ -13,52 +13,81 @@
...
+
+interface ApiCallSettings {
+  merge: () => {
+    otherArgs: {}
+  };
+}
+type Callback<T> = (err: Error|null, res?: T) => void;
+type InnerApiCall<I, O> =
+    (request: I, metadata: {}, options: {}, callback: Callback<O>) => void;
+type OuterApiCall<I, O> =
```
jinwoo: \<rant\> I still don't understand the google-gax api... but that's not the problem of trace agent... \</rant\>

kjin: 🤣

kjin: At its core it's a `Promise<Function> => Function` transform

PR #705, opened 2018-03-28
test: privatize forceNewAgent_

kjin:
This PR makes the flag to re-create a new instance of a singleton (or the Trace Agent as a whole) a module-private symbol `util#FORCE_NEW`. Beyond hiding this from the public API, this symbol's name now more accurately describes its general role of forcing the creation of a new singleton.

Besides replacing `forceNewAgent_` occurrences with the symbol, there are no other logical changes.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -50,6 +50,10 @@ export interface Constructor<T, Config> {
   name: string;
 }
 
+export const FORCE_NEW = Symbol();
+
+export type Forceable<T> = T&{[FORCE_NEW]?: boolean};
+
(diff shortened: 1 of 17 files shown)


src/util.ts
```diff
@@ -50,6 +50,10 @@ export interface Constructor<T, Config> {
   name: string;
 }
 
+export const FORCE_NEW = Symbol();
```
jinwoo: Giving a string value to the symbol would help debugging.
```ts
Symbol('force-new');
```

PR #710, opened 2018-04-02
feat: hapi 17 tracing support

### LINKED ISSUE #601: Support hapi@17
This issue tracks support for hapi 17 tracing.

kjin:
This PR adds support for tracing Hapi 17. `GCLOUD_TRACE_NEW_CONTEXT` must be set.

Fixes #601

### DIFF
--- a/src/plugins/plugin-hapi.ts
+++ b/src/plugins/plugin-hapi.ts
@@ -20,11 +20,19 @@ import {parse as urlParse} from 'url';
 
 import {PluginTypes} from '..';
 
-import {hapi_16} from './types';
+import {hapi_16, hapi_17} from './types';
+
+// Used when patching Hapi 17.
+const ORIGINAL = Symbol();
 
 type Hapi16Module = typeof hapi_16;
-
-const SUPPORTED_VERSIONS = '8 - 16';
+type Hapi17RequestExecutePrivate = {
+  (this: hapi_17.Request): Promise<void>;
+      [ORIGINAL]?: Hapi17RequestExecutePrivate;
+};
+type Hapi17Request = hapi_17.Request&{
+  _execute: Hapi17RequestExecutePrivate;
+};
 
 function getFirstHeader(req: IncomingMessage, key: string): string|null {
   let headerValue = req.headers[key] || null;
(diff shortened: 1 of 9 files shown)


src/plugins/plugin-hapi.ts
```diff
@@ -20,11 +20,19 @@ import {parse as urlParse} from 'url';
...
 
-type Hapi16Module = typeof hapi_16;
+// Used when patching Hapi 17.
+const ORIGINAL = Symbol();
 
-const SUPPORTED_VERSIONS = '8 - 16';
+type Hapi16Module = typeof hapi_16;
+type Hapi17RequestExecutePrivate = {
+  (this: hapi_17.Request): Promise<void>;
+      [ORIGINAL]?: Hapi17RequestExecutePrivate;
```
jinwoo: Indentation seems off. Is this how clang-format does?

src/plugins/plugin-hapi.ts
```diff
@@ -34,87 +42,121 @@ function getFirstHeader(req: IncomingMessage, key: string): string|null {
...
-  },
-  unpatch: (hapi) => {
-    shimmer.unwrap(hapi.Server.prototype, 'connection');
-  }
-} as PluginTypes.Patch<Hapi16Module>];
+const plugin: PluginTypes.Plugin = [
+  {
+    versions: '8 - 16',
+    patch(hapi, api) {
+      function createMiddleware(): hapi_16.ServerExtRequestHandler {
```
jinwoo: This can be an arrow function, no?

src/plugins/plugin-hapi.ts
kjin: clang-format seems to not format `type` declarations correctly. But this can changed to an interface, so I'll do that.

src/plugins/plugin-hapi.ts
kjin: Yep, done

jinwoo: What I asked was about `createMiddleware()` :) But `patch` and `unpatch` also seem better as arrow functions.

kjin: Ah, gotcha -- done (x2)

BenCrux:
This is great @kjin - any idea when this would become a version we'd be able to pull down and use?

PR #711, opened 2018-04-02
build: add script to fetch plugin types

kjin:
This PR adds a script that installs all type definitions for plugins in a directory distinct from `node_modules`. We need to do this as `npm` doesn't allow direct installation of more than one version of a module. The `get-plugin-types` script auto-generates dummy modules with dependencies on one `@types` package each; it doesn't do anything if these modules already exist.

This PR is a pre-requisite for #710.


scripts/get-plugin-types.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+}
+
+export async function getPluginTypes() {
+  await mkdirSafeP(TYPES_DIRECTORY);
+
+  const indexTs = (await readFileP(`${TYPES_DIRECTORY}/index.d.ts`, 'utf8') as string)
+    .split('\n');
+  for (const line of indexTs) {
+    const matches = line.match(/^import \* as .* from '\.\/(.*)';\s*\/\/\s*(.*)@(.*)$/);
```
jinwoo: I think the regex's for `packageName`, `name`, and `version` must use `(.+)` rather than `(.*)`

scripts/get-plugin-types.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (!matches) {
+      continue;
+    }
+    const [_0, packageName, name, version] = matches;
```
jinwoo: You can omit `_0`.
```ts
const [, packageName, name, version] = matches;
```

kjin:
Node 9 failure is unrelated (CircleCI attempts to run system tests when rebuilding from the UI). Landing

scripts/get-plugin-types.ts
```diff
@@ -0,0 +1,41 @@
...
+  await mkdirSafeP(TYPES_DIRECTORY);
+
+  const indexTs = (await readFileP(`${TYPES_DIRECTORY}/index.d.ts`, 'utf8') as string)
+    .split('\n');
+  for (const line of indexTs) {
+    const matches = line.match(/^import \* as .* from '\.\/(.+)';\s*\/\/\s*(.+)@(.+)$/);
+    if (!matches) {
+      continue;
+    }
+    console.log(matches);
```
jinwoo: Did you add this intentionally?

PR #713, opened 2018-04-02
fix: treat instanceId metadata as a number

### LINKED ISSUE #712: Error after updating to 2.6.0
Hi,
after updating I get the following error:
```
ERROR:@google-cloud/trace-agent: TraceWriter#publish: Received error status code 400. Original error: ApiError: Invalid value at 'traces.traces[0].spans[0].labels[7].value' (TYPE_STRING), 9154848679131446000
Invalid value at 'traces.traces[1].spans[0].labels[7].value' (TYPE_STRING), 9154848679131446000
Invalid value at 'traces.traces[2].spans[0].labels[7].value' (TYPE_STRING), 9154848679131446000
```

I have got a very basic config, which is loaded at the very start of the app:
```
require('@google-cloud/trace-agent').start({
  enhancedDatabaseReporting: true,
  logLevel: 4,
  ignoreUrls: [/^\/healthz/, /^\/ready/], // ignore the "/healthz" and "/ready" endpoints.
});

require('./dist/system-server/main');
```

Running with the previous version does work fine.
Any idea? I saw in the changelog that empty spans are not allowed, but could not find a way how to fix the error above.

Thanks in advance
Daniel

kjin:
Fixes #712

This PR accounts for a change in #680 in how trace labels are applied. By directly assigning fields in a spans' labels instead of using `addLabel`, we were assuming that all default labels were strings. This isn't the case, because GCP instance IDs are numbers.

### DIFF
--- a/src/trace-writer.ts
+++ b/src/trace-writer.ts
@@ -140,7 +140,9 @@ export class TraceWriter extends common.Service {
 
     this.getHostname((hostname) => {
       this.getInstanceId((instanceId) => {
-        const labels: LabelObject = {};
+        this.defaultLabels = {};
+        // tslint:disable-next-line:no-any
+        const labels: {[key: string]: any} = {};
         labels[TraceLabels.AGENT_DATA] =
             'node ' + pjson.name + ' v' + pjson.version;
         labels[TraceLabels.GCE_HOSTNAME] = hostname;
@@ -163,8 +165,11 @@ export class TraceWriter extends common.Service {
             labels[TraceLabels.GAE_VERSION] = versionLabel;
           }
         }
-        Object.freeze(labels);
-        this.defaultLabels = labels;
+        // Coerce values to strings.
+        for (const key of Object.keys(labels)) {
+          this.defaultLabels[key] = `${labels[key]}`;
+        }
+        Object.freeze(this.defaultLabels);
         if (--pendingOperations === 0) {
           cb();
         }
@@ -192,7 +197,7 @@ export class TraceWriter extends common.Service {
         });
   }
 
-  getInstanceId(cb: (instanceId?: string) => void) {
+  getInstanceId(cb: (instanceId?: number) => void) {
     gcpMetadata.instance({property: 'id', headers})
         .then((res) => {
           cb(res.data);  // instance ID
--- a/src/trace-writer.ts
+++ b/src/trace-writer.ts
@@ -140,19 +140,24 @@ export class TraceWriter extends common.Service {
 
     this.getHostname((hostname) => {
       this.getInstanceId((instanceId) => {
-        const labels: LabelObject = {};
-        labels[TraceLabels.AGENT_DATA] =
-            'node ' + pjson.name + ' v' + pjson.version;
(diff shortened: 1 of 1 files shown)


src/trace-writer.ts
```diff
@@ -140,7 +140,9 @@ export class TraceWriter extends common.Service {
 
     this.getHostname((hostname) => {
       this.getInstanceId((instanceId) => {
-        const labels: LabelObject = {};
+        this.defaultLabels = {};
+        // tslint:disable-next-line:no-any
+        const labels: {[key: string]: any} = {};
```
jinwoo: I think you can deceive the type system by replacing `any` to `{}`.

src/trace-writer.ts
```diff
@@ -163,8 +165,11 @@ export class TraceWriter extends common.Service {
             labels[TraceLabels.GAE_VERSION] = versionLabel;
           }
         }
-        Object.freeze(labels);
-        this.defaultLabels = labels;
+        // Coerce values to strings.
+        for (const key of Object.keys(labels)) {
+          this.defaultLabels[key] = `${labels[key]}`;
```
jinwoo: Why not just convert each value to a string when it is filled into `labels`?

kjin: I could do that. Do you think `${x}` is an adequate way of doing so?

jinwoo: I think so because `${x}` is translated to `x.toString()`, I think.

kjin: Ok, sounds good. Please see the PR diff: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/713/files

I've updated it so that fields that aren't certain to be strings are now coerced to strings. That is, user-config service version and instanceID.

src/trace-writer.ts
```diff
@@ -145,14 +145,14 @@ export class TraceWriter extends common.Service {
             'node ' + pjson.name + ' v' + pjson.version;
         labels[TraceLabels.GCE_HOSTNAME] = hostname;
         if (instanceId) {
-          labels[TraceLabels.GCE_INSTANCE_ID] = instanceId;
+          labels[TraceLabels.GCE_INSTANCE_ID] = `${instanceId}`;
```
jinwoo: I'd rather define a local function that adds a new label to `labels`, which always converts the given value to a string. That'll make this code more extensible in the future when adding new labels.

src/trace-writer.ts
```diff
@@ -140,31 +140,36 @@ export class TraceWriter extends common.Service {
...
-        labels[TraceLabels.GCE_HOSTNAME] = hostname;
+        // tslint:disable-next-line:no-any
+        const addDefaultLabel = (key: string, value: any) => {
+          this.defaultLabels[key] = `${value}`;
+        };
+
+        this.defaultLabels = {};
+        addDefaultLabel(
+            TraceLabels.AGENT_DATA,
+            'node ' + pjson.name + ' v' + pjson.version);
```
jinwoo: Change to a template string while you're here?

PR #717, opened 2018-04-04
fix: fix https tracing breakage in node <9 and rewrite http tests

kjin:
`https` tracing was broken in #686 for Node 4/6/8 (except for 8.9.0). This change fixes that, and revises the http/https tracing tests (all tests are preserved except for the one measuring time specifically when calling http.get with a string argument -- I didn't think we needed that much granularity).


src/plugins/plugin-http.ts
```diff
@@ -183,7 +183,8 @@ function patchHttp(http: HttpModule, api: TraceAgent) {
   }
 }
 
-// https.get depends on https.request in <8.9 and >=8.9.1
+// https.get depends on Node http internals in 8.9.0 and 9+ instead of the
```
jinwoo: The actual code is `<8.9.0 || ^8.9.1` while this comment says `8.9.0 and 9+`. Is it intentional?

test/plugins/test-trace-http.ts
```diff
@@ -13,357 +13,420 @@
...
-  var server;
-  before(function() {
-    http = require('http');
+/**
+ * A class that represents a convenience object that allows us to await
+ * http requests. This is done in lieu of promisifying http.request/http.get,
+ * because they already have a meaningful return value.
+ */
+class WaitForResponse {
+  private resolve!: (value: string) => void;
```
jinwoo: Why do you need `!` here and below?

test/plugins/test-trace-http.ts
```diff
@@ -13,357 +13,420 @@
...
-  it('should patch the necessary functions', function() {
-    assert.strictEqual(http.request.__wrapped, true);
-    if (semver.satisfies(process.version, '>=8.0.0')) {
-      assert.strictEqual(http.get.__wrapped, true, 'should patch get');
-    } else {
-      assert.strictEqual(http.get.__wrapped, undefined, 'should not patch get');
+// Server abstraction class definitions. These are borrowed from web framework
+// tests -- which are useful because they already expose a Promise API.
+const servers = {
+  http: Express4, https: class Express4Secure extends Express4 {
```
jinwoo: Formatting is weird. Can you append a comma at the end of line 105 and reformat?

test/plugins/test-trace-http.ts
```diff
@@ -13,357 +13,420 @@
...
-      })
-    );
-  });
+    listen(port: number): number {
+      // The types of (http|https).Server are not compatible, but we don't
+      // access any properties that aren't present on both in the test.
+      // tslint:disable:no-any
+      this.server = this.https.createServer(
+                        {key: Express4Secure.key, cert: Express4Secure.cert},
+                        this.app) as any as httpModule.Server;
```
jinwoo: You can say `as {} as httpModule.Server` and delete the `tslint` pragma above.

test/plugins/test-trace-http.ts
kjin: TS doesn't know that the callback passed to the Promise constructor will be invoked synchronously (sound familiar? :))

test/plugins/test-trace-http.ts
kjin: Comma doesn't help, and for the record reversing the order of `http` and `https` results in even more weirdness.

I've moved `class Express4Secure` out be declared immediately before, that helps.

test/plugins/test-trace-http.ts
jinwoo: Now I understand. Wish TS understands how Promise works...

test/plugins/test-trace-http.ts
kjin: Done

src/plugins/plugin-http.ts
kjin: `patchHttps` is actually used in the patch for `<8.9.0 || ^8.9.1`. (Not really related, but the comment was inaccurate before.)

PR #718, opened 2018-04-05
chore: 2.7.1 release proposal

kjin:


kjin:
Node 4 failure: See https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/711#issuecomment-378085714

PR #720, opened 2018-04-09
fix: allow non-objects for plugins to disable automatic tracing

### LINKED ISSUE #719: Option to disable automatic tracing
Hello,

Thanks

kjin:
Fixes #719 

This change allows users to disable automatic tracing by passing a non-object value (such as `false`) for the value of `config.plugins`. It takes advantage of `extend`'s behavior of allowing non-object values to override object values, and vice versa.

Since we do not recommend disabling plugins I have not documented this feature.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -89,7 +89,7 @@ function initConfig(projectConfig: Forceable<Config>):
   // 4. Default Config (as specified in './config')
   const config = extend(
       true, {[FORCE_NEW]: projectConfig[FORCE_NEW]}, defaultConfig,
-      envSetConfig, projectConfig, envConfig, {plugins: {/**/}});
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -89,7 +89,9 @@ function initConfig(projectConfig: Forceable<Config>):
   // 4. Default Config (as specified in './config')
   const config = extend(
       true, {[FORCE_NEW]: projectConfig[FORCE_NEW]}, defaultConfig,
-      envSetConfig, projectConfig, envConfig);
+      envSetConfig, projectConfig, envConfig, {plugins: {/**/}});
```
ofrobots: Remove the comment from the object?

PR #723, opened 2018-04-16
test: fix system test

kjin:
System tests fail often (but not always) because they rely on an authentication call being traced, when this is highly dependent on timing conditions (authentication needs to happen after the first call is made).

### DIFF
--- a/scripts/index.ts
+++ b/scripts/index.ts
@@ -8,7 +8,7 @@
 
 const [, , ...steps] = process.argv;
 const {
-  CI_PULL_REQUEST,
+  CIRCLE_PR_NUMBER,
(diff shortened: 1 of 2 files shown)


scripts/index.ts
```diff
@@ -8,7 +8,7 @@
 
 const [, , ...steps] = process.argv;
 const {
-  CI_PULL_REQUEST,
+  CIRCLE_PR_NUMBER,
```
jinwoo: Is this env var set by CircleCI?

kjin: Yes. Forgot to mention that `CIRCLE_PR_NUMBER` is always set for PRs, but `CI_PULL_REQUEST` is unset when the CI is re-run through the CircleCI UI.

PR #724, opened 2018-04-16
fix: fix log messages and ignore falsey env vars

kjin:
This change causes the Trace Agent to now ignore empty string environmental variables. Formerly,

```
GCLOUD_PROJECT= node app.js
```

would initialize the `projectId` field to an empty string.

This change also fixes a few log message contents, and adds two info-level log messages for when new spans are created.


kjin:
Node 6 CI passes despite failure output (which is due to this PR not being rebased on #723)

PR #727, opened 2018-04-19
fix: restore context when a function run with a given context throws

kjin:
Addresses a TODO from #708: in two different cases, a function is accepted by the tracing API; we currently do not do any restoration of context if that function throws (in other words, the newly added tests would fail). This PR addresses that.

### DIFF
--- a/src/cls/async-hooks.ts
+++ b/src/cls/async-hooks.ts
@@ -86,10 +86,12 @@ export class AsyncHooksCLS<Context extends {}> implements CLS<Context> {
   runWithNewContext<T>(fn: Func<T>): T {
     const oldContext = this.currentContext.value;
     this.currentContext.value = this.defaultContext;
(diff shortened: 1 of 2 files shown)


src/cls/async-hooks.ts
```diff
@@ -86,10 +86,12 @@ export class AsyncHooksCLS<Context extends {}> implements CLS<Context> {
   runWithNewContext<T>(fn: Func<T>): T {
     const oldContext = this.currentContext.value;
     this.currentContext.value = this.defaultContext;
-    // TODO(kjin) Handle the case where fn throws. Context: PR #708
-    const res = fn();
-    this.currentContext.value = oldContext;
-    return res;
+    try {
+      const res = fn();
+      return res;
```
jinwoo: Nit. Just `return fn();`? Same below.

PR #728, opened 2018-04-19
feat: allow "disabling" cls, and relax requirements for creating root spans

### LINKED ISSUE #719: Option to disable automatic tracing

kjin:
Fixes #719

This PR introduces a new config option `clsMechanism` that can either be set to `'none'` or `'auto'` (and defaults to the latter, preserving the current behavior). Setting it to `'none'` effectively treats the entire application as sharing the same root context, so any code running subsequent to `runInRootSpan(spanConfig, (rootSpan) => { /*...*/ })` will see all child spans created under that root _until_ `rootSpan.endSpan()` is called.

The `'none'` CLS mechanism is at odds with the previous limitation that there may only be one root span per context, so this PR also changes behavior so that when a root span ends, context is cleared. __As a result, [this code path](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/edb8135a7960815e8b112aae3a2c3c34e9b3d812/src/trace-api.ts#L202-L215) should now never be executed, so I have removed it.__

The new semantics for creating root spans is as follows:

* If there is no root context in the current continuation, create a new root span, and a corresponding root context in the current continuation.
* If there is a root context in the current continuation, do not create a new root span.
* When a root span is ended, clear the root context in the current continuation. (This is the new change)

Note that the "current continuation" in the first and third bullet points may not be the same continuation, even if the root spans mentioned are the same. It is still OK to clear the root context in the current continuation (as stated in the third bullet point) because there was none to begin with.


src/cls.ts
```diff
@@ -80,30 +82,34 @@ export class TraceCLS implements CLS<RootContext> {
...
+          break;
+        } else {
           this.logger.error(
               'TraceCLS#constructor: [async-hooks]-based context',
               `propagation is not available in Node ${process.version}.`);
-        } else {
-          this.logger.error(
-              'TraceCLS#constructor: The specified CLS mechanism',
-              `[${config.mechanism}] was not recognized.`);
+          throw new Error(`CLS mechanism [${config.mechanism}] is invalid.`);
```
ofrobots: nit: the logged error message seem a lot more appropriate. Can you just use that? Also, do we need to log given that we are throwing?

ofrobots: follow-on: nit: reverse the if so that the throw happens early. Remove the indentation of the else block.

src/cls.ts
```diff
@@ -80,30 +82,34 @@ export class TraceCLS implements CLS<RootContext> {
...
+        this.rootSpanStackOffset = 8;
+        break;
+      case 'none':
+        this.CLSClass = UniversalCLS;
+        this.rootSpanStackOffset = 4;
+        break;
+      default:
+        this.logger.error(
+            'TraceCLS#constructor: The specified CLS mechanism',
+            `[${config.mechanism}] was not recognized.`);
```
ofrobots: likewise. Given that you're throwing, do we need to log?

src/cls/async-listener.ts
```diff
@@ -69,9 +69,15 @@ export class AsyncListenerCLS<Context extends {}> implements CLS<Context> {
...
+  clearContext(): void {
+    if (this.getContext() !== this.defaultContext) {
+      this.setContext(this.defaultContext);
+    }
+  }
+
   runWithNewContext<T>(fn: Func<T>): T {
     return this.getNamespace().runAndReturn(() => {
-      this.setContext(this.defaultContext);
+      this.clearContext();
```
ofrobots: I am trying to understand why a `clearContext` is necessary here? You didn't change this as part of the PR, but I don't understand the intuition behind this.

src/config.ts
```diff
@@ -166,6 +178,7 @@ export interface Config {
  * user-provided value will be used to extend the default value.
  */
 export const defaultConfig = {
+  clsMechanism: 'auto' as ContextPropagationMechanism,
```
ofrobots: make `ContextPropagationMechanism` an enum and avoid the cast.

src/index.ts
```diff
@@ -56,6 +56,7 @@ for (let i = 0; i < filesLoadedBeforeTrace.length; i++) {
 interface TopLevelConfig {
   enabled: boolean;
   logLevel: number;
+  clsMechanism: 'none'|'auto';
```
ofrobots: Can you use `ContextPropagationMechanism` type here instead of repeating the type?

test/plugins/test-trace-google-gax.ts
```diff
@@ -59,11 +59,17 @@ describe('Tracing with google-gax', () => {
   let googleGax: GaxModule;
 
   before(() => {
+    trace.setCLS();
     trace.setPluginLoader();
     trace.start();
     googleGax = require('./fixtures/google-gax0.16');
   });
 
+  after(() => {
```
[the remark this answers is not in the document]
kjin: It's not technically needed, but it's there for the sake of keeping `before` and `after` symmetrical, as `after` is responsible for reversing any global state changes enacted in `before`. Other unit tests already had similar `after` methods; I had just forgotten to write it for this one.

src/index.ts
```diff
@@ -178,12 +179,12 @@ export function start(projectConfig?: Config): PluginTypes.TraceAgent {
...
     // Initialize context propagation mechanism.
-    // TODO(kjin): Publicly expose this field.
-    cls.create(logger, {
-         mechanism: useAH ? 'async-hooks' : 'async-listener',
-         [FORCE_NEW]: config[FORCE_NEW]
-       })
-        .enable();
+    const m = config.clsMechanism;
+    const clsConfig: Forceable<TraceCLSConfig> = {
+      mechanism: m === 'auto' ? (useAH ? 'async-hooks' : 'async-listener') : m,
```
ofrobots: nit: I would prefer enum values being used rather than hard-coded strings.

src/trace-api.ts
```diff
@@ -199,20 +199,6 @@ export class TraceAgent implements TraceAgentInterface {
...
-        // up getting misattributed to the same root span – we get a root span
-        // with continuously growing number of child spans. The second case
-        // seems to have some value, but isn't representable. The user probably
-        // needs a custom outer span that encompasses the entirety of work.
-        this.logger!.warn(`TraceApi#createChildSpan: [${
-            this.pluginName}] Creating phantom child span [${
-            options.name}] because root span [${
-            rootSpan.span.name}] was already closed.`);
-        return UNCORRELATED_SPAN;
-      }
```
ofrobots: This is a useful error message. I don't see how we can remove it without an equivalent error message some where else.

We still need to be able to distinguish context confusion cases from context loss cases. As an example, this error message is very useful in identifying situations with a continuation is correlated to the wrong RootSpan – this shows up in the background log flusing use-case for example. If a Log was initialized in the first context, and flushes in the background chained on some promise, it would appear to be a continuation of the first context. This error message catches that scenario.

src/config.ts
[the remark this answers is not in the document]
jinwoo: I see. We lost type-safety from it though. If someone makes a typo here in the future, it may not be detected by the compiler.

Would it be helpful to use `Partial`? I.e. make `Config`'s field required and define `PartialConfig` out of it using `Partial`... Just a random idea. Feel free to ignore.

kjin: I don't think user-facing values can be enums. It would explicitly prohibit them from specifying string parameters unless they casted it.

src/trace-api.ts
kjin: Added it back in. I think having this logging message available actually isn't incompatible with allowing sequential root spans as I had originally thought.

src/cls/async-listener.ts
kjin: `clearContext` has been removed. It's unneeded.

src/cls/async-listener.ts
```diff
@@ -69,9 +69,15 @@ export class AsyncListenerCLS<Context extends {}> implements CLS<Context> {
     this.getNamespace().set(AsyncListenerCLS.ROOT_CONTEXT_KEY, value);
   }
 
+  clearContext(): void {
+    if (this.getContext() !== this.defaultContext) {
```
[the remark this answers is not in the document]
kjin: Yes, `set` throws if we are currently not running from within `runAndReturn`, and this check guarantees that we are. I removed `clearContext` completely, though.

kjin:
Summary of changes:

- I added the `TraceCLSMechanism` enum. It is not user-facing.
- Instead of clearing context, we instead simply explicitly allow root spans to be created when a root span context already exists, but that root span is ended, while still warning if child spans are created under those same conditions.

src/cls.ts
```diff
@@ -48,11 +50,34 @@ export interface PhantomRootContext {
  */
 export type RootContext = RealRootContext|PhantomRootContext;
 
-const asyncHooksAvailable = semver.satisfies(process.version, '>=8');
+/**
+ * An enumeration of the possible mechanisms for supporting context propagation
+ * through continuation-local storage.
+ */
+export enum TraceCLSMechanism {
```
jinwoo: Is there any reason this should be a string enum vs a numeric one?

src/span-data.ts
```diff
@@ -17,6 +17,7 @@
 import * as crypto from 'crypto';
 import * as util from 'util';
 
+import {cls} from './cls';
```
ofrobots: unused.

PR #730, opened 2018-04-25
chore: 2.8.0 release proposal

kjin:


kjin:
Closing this, as I would like to make a change to the `clsMechanism` option that would otherwise be semver-major.

PR #731, opened 2018-04-27
feat: add rootSpan.createChildSpan and change none CLS semantics

### LINKED ISSUE #719: Option to disable automatic tracing

kjin:
Fixes #719

This PR does two things:

* In #728, a new `config.clsMechanism` option was added where, when set to `'none'`, a barebones CLS mechanism would be used where when a root span was started, it became the only possible active root span context until it was ended. This PR changes this so that root span context will _never_ be written when `clsMechanism` is set to `'none'`. Instead, the root span context always has a default value of `UNTRACED`. This affects the behavior of the custom tracing API in the following ways:
  * `traceApi.runInRootSpan` will _always_ return a root span (unless disallowed by the trace policy), even if run within a function passed to another call to `runInRootSpan`.
  * `traceApi.createChildSpan` will _always_ return an `UNTRACED` child span. This is because root spans can no longer be automatically determined from the root context.
* To address the second difference above, there is now a `rootSpan.createChildSpan` which takes the same options as `traceApi.createChildSpan`. When `clsMechanism` is set to `'none'`, the latter will not work, but the former will; when `clsMechanism` is set to/defaults to `'auto'`, they should behave the same way.

### DIFF
--- a/src/cls.ts
+++ b/src/cls.ts
@@ -139,7 +139,7 @@ export class TraceCLS implements CLS<RootContext> {
   }
 
   enable(): void {
-    // if this.CLSClass = NoneCLS, the user specifically asked not to use
+    // if this.CLSClass = NullCLS, the user specifically asked not to use
     // any context propagation mechanism. So nothing should change.
     if (!this.enabled && this.CLSClass !== NullCLS) {
       this.logger.info('TraceCLS#enable: Enabling CLS.');
--- a/src/cls.ts
+++ b/src/cls.ts
@@ -21,7 +21,7 @@ import * as semver from 'semver';
 import {AsyncHooksCLS} from './cls/async-hooks';
 import {AsyncListenerCLS} from './cls/async-listener';
 import {CLS, Func} from './cls/base';
-import {DefaultCLS} from './cls/default';
+import {NullCLS} from './cls/null';
 import {SpanDataType} from './constants';
 import {SpanData, SpanOptions} from './plugin-types';
 import {Trace, TraceSpan} from './trace';
@@ -121,7 +121,7 @@ export class TraceCLS implements CLS<RootContext> {
         this.rootSpanStackOffset = 8;
         break;
       case TraceCLSMechanism.NONE:
-        this.CLSClass = DefaultCLS;
+        this.CLSClass = NullCLS;
         this.rootSpanStackOffset = 4;
         break;
       default:
@@ -130,7 +130,7 @@ export class TraceCLS implements CLS<RootContext> {
     }
     this.logger.info(
         `TraceCLS#constructor: Created [${config.mechanism}] CLS instance.`);
-    this.currentCLS = new DefaultCLS(TraceCLS.UNTRACED);
+    this.currentCLS = new NullCLS(TraceCLS.UNTRACED);
     this.currentCLS.enable();
   }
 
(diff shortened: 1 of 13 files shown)


tjmehta:
Wow you are fast @kjin! I am out of town and will check this out tomorrow

scripts/check-install.ts
```diff
@@ -19,7 +19,7 @@ export async function checkInstall() {
   await spawnP('npm', ['init', '-y'], {
     cwd: installDir
   });
-  await spawnP('npm', ['install', 'typescript', '@types/node', tgz[0]], {
+  await spawnP('npm', ['install', 'typescript', '@types/node@9', tgz[0]], {
```
jinwoo: Why is this needed?

src/cls.ts
```diff
@@ -137,23 +139,25 @@ export class TraceCLS implements CLS<RootContext> {
   }
 
   enable(): void {
-    if (!this.enabled) {
+    // if this.CLSClass = DefaultCLS, the user specifically asked not to use
```
jinwoo: So `DefaultCLS` is a CLS that does nothing, am I correct? The name seems confusing then. What about `NullCLS`?

src/span-data.ts
```diff
@@ -18,7 +18,8 @@ import * as crypto from 'crypto';
 import * as util from 'util';
 
 import {Constants, SpanDataType} from './constants';
-import {SpanData as SpanData} from './plugin-types';
+import * as types from './plugin-types';
+import {SpanData, SpanOptions} from './plugin-types';
```
jinwoo: You seem to use only `types.SpanData` in this file. Why separately import these BTW?

src/span-data.ts
```diff
@@ -120,6 +121,16 @@ export class RootSpanData extends BaseSpanData {
     this.span.kind = SpanKind.RPC_SERVER;
   }
 
+  createChildSpan(options: SpanOptions): SpanData {
+    options = options || {name: ''};
```
jinwoo: Is `options` optional? Why do you need this line?

kjin: They're optional, but in the absence of type checking it prevents an error from being thrown when options aren't passed in. It's sort of a relic of older versions of this module -- I added it to keep it consistent with the same precondition check in `traceApi.createChildSpan`.

jinwoo: Erm... I'm not familiar with the history of this module, but why not mark it as optional then?

kjin: The "history of this module" is just that it was once written in JavaScript :)

I've marked `options` as optional.

src/cls.ts
```diff
@@ -137,23 +139,25 @@ export class TraceCLS implements CLS<RootContext> {
   }
 
   enable(): void {
-    if (!this.enabled) {
+    // if this.CLSClass = NoneCLS, the user specifically asked not to use
```
jinwoo: s/NoneCLS/NullCLS

src/span-data.ts
```diff
@@ -18,7 +18,8 @@ import * as crypto from 'crypto';
 import * as util from 'util';
 
 import {Constants, SpanDataType} from './constants';
-import {SpanData as SpanData} from './plugin-types';
+import * as types from './plugin-types';
```
jinwoo: Why not just add `RootSpanData` to the import statement below? You don't seem to use anything other than those three (SpanData, SpanOptions, and RootSpanData), do you?

kjin: It conflicts with `RootSpanData` defined in this file. I couldn't think of any way to avoid this at the time.

Come to think of it, OpenCensus calls the write-only object "Span" (and SpanData refers to the read-only object given to exporters). I'll add a TODO to change this.

kjin:
@tjmehta Were you able to get a chance to look at the new API? (I understand that it's a sizeable PR -- I think the user-visible changes should most be encapsulated in [this file](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/731/files#diff-0083d751b2c247998a6ebdda46fffa61))

PR #734, opened 2018-05-02
fix: adjust async_hooks cls behavior

kjin:
This PR introduces the following changes to async_hooks-based tracing:
- We now always use the `triggerId` when determining the parent `AsyncResource` _unless_ the current `AsyncResource` is of type `PROMISE`.
- We clear `this.contexts` upon disabling CLS (in practice, this happens when the Trace Agent is disabled).
- Instead of sharing the value of the current context between different async scopes, we share a reference to it instead. This way, if we set context to a different value, the change is reflected in other scopes. (This makes things more consistent but I don't think it has behavioral consequences from the perspective of the user.)
  * Async scopes refer to either (1) scope in the sense of `AsyncResource#runInAsyncScope`, or (2) the function passed to either `AsyncHooksCLS#runInNewContext` or `AsyncHooksCLS#bindToCurrentContext`.
- Instead of using `this.currentContext` to get the current context, we now directly look up the current context in `this.contexts` based on the current execution ID.
  - As a result, the `before` hook can now be removed.

### DIFF
--- a/src/cls/async-hooks.ts
+++ b/src/cls/async-hooks.ts
@@ -14,15 +14,15 @@
  * limitations under the License.
  */
 
-// This file requires async_hooks in the AsyncHooksCLS constructor, rather than
-// upon module load.
-import * as asyncHooks from 'async_hooks';
+// This file calls require('async_hooks') in the AsyncHooksCLS constructor,
+// rather than upon module load.
+import * as asyncHooksModule from 'async_hooks';
 import {EventEmitter} from 'events';
 import * as shimmer from 'shimmer';
 
 import {CLS, Func} from './base';
 
-type AsyncHooksModule = typeof asyncHooks;
+type AsyncHooksModule = typeof asyncHooksModule;
 
 // A list of well-known EventEmitter methods that add event listeners.
 const EVENT_EMITTER_METHODS: Array<keyof EventEmitter> =
@@ -46,7 +46,7 @@ export class AsyncHooksCLS<Context extends {}> implements CLS<Context> {
   /** A map of AsyncResource IDs to Context objects. */
   private contexts: {[id: number]: Reference<Context>} = {};
   /** The AsyncHook that proactively populates entries in this.contexts. */
-  private hook: asyncHooks.AsyncHook;
+  private hook: asyncHooksModule.AsyncHook;
   /** Whether this instance is enabled. */
   private enabled = false;
 
@@ -145,14 +145,19 @@ export class AsyncHooksCLS<Context extends {}> implements CLS<Context> {
   }
 
   bindWithCurrentContext<T>(fn: Func<T>): Func<T> {
+    // Return if we have already wrapped the function.
+    if ((fn as ContextWrapped<Func<T>>)[WRAPPED]) {
+      return fn;
+    }
     // Capture the context of the current AsyncResource.
     const boundContext = this.contexts[this.ah.executionAsyncId()];
-    // Return if we have already wrapped the function, or there is no current
-    // context to bind.
-    if ((fn as ContextWrapped<Func<T>>)[WRAPPED] || !boundContext) {
+    // Return if there is no current context to bind.
+    if (!boundContext) {
       return fn;
     }
     const that = this;
+    // TODO(kjin): This code is somewhat duplicated with runWithNewContext.
+    //             Can we merge this?
     // Wrap fn so that any AsyncResource objects that are created in fn will
     // share context with that of the AsyncResource with the given ID.
     const contextWrapper: ContextWrapped<Func<T>> = function(this: {}) {
--- a/src/cls/async-hooks.ts
+++ b/src/cls/async-hooks.ts
@@ -118,7 +118,13 @@ export class AsyncHooksCLS<Context extends {}> implements CLS<Context> {
   }
 
   setContext(value: Context): void {
-    this.contexts[this.ah.executionAsyncId()].value = value;
+    const id = this.ah.executionAsyncId();
+    const current = this.contexts[id];
+    if (current) {
+      current.value = value;
+    } else {
+      this.contexts[id] = {value};
+    }
   }
 
   runWithNewContext<T>(fn: Func<T>): T {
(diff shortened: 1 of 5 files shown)


src/cls/async-hooks.ts
```diff
@@ -31,29 +31,63 @@ const EVENT_EMITTER_METHODS: Array<keyof EventEmitter> =
...
+  private hook: asyncHooks.AsyncHook;
+  /** Whether this instance is enabled. */
   private enabled = false;
 
   constructor(private readonly defaultContext: Context) {
-    this.currentContext = {value: this.defaultContext};
-    this.hook = (require('async_hooks') as AsyncHooksModule).createHook({
+    // Store a reference to the async_hooks module, since we will need to query
+    // the current AsyncResource ID often.
+    this.ah = require('async_hooks') as AsyncHooksModule;
```
jinwoo: You already `import`ed `async_hook` at the top of this file and I think you can reuse it here.

src/cls/async-hooks.ts
```diff
@@ -64,51 +98,75 @@ export class AsyncHooksCLS<Context extends {}> implements CLS<Context> {
...
+    // We don't store this.defaultContext directly in this.contexts.
+    // Getting undefined when looking up this.contexts means that it wasn't
+    // set, so return the default context.
+    const current = this.contexts[this.ah.executionAsyncId()];
+    return current ? current.value : this.defaultContext;
   }
 
   setContext(value: Context): void {
-    this.currentContext.value = value;
+    this.contexts[this.ah.executionAsyncId()].value = value;
```
jinwoo: `this.contexts[this.ah.executionAsyncId()]` can be undefined if it hasn't been assigned before. How do we make sure that it already exists? Is `setContext()` always called from within `runWithNewContext()`?

src/cls/async-hooks.ts
```diff
@@ -64,51 +98,75 @@ export class AsyncHooksCLS<Context extends {}> implements CLS<Context> {
...
+    // Return if we have already wrapped the function, or there is no current
+    // context to bind.
+    if ((fn as ContextWrapped<Func<T>>)[WRAPPED] || !this.contexts[outerId]) {
       return fn;
     }
-    const current = this.currentContext;
-    const boundContext = this.currentContext.value;
+    const that = this;
+    // Capture the context of the current AsyncResource.
+    const boundContext = this.contexts[outerId];
```
jinwoo: Nit. It'd be better to move this line to before line 145, so that `this.contexts[outerId]` can be called only once.

src/cls/async-hooks.ts
kjin: Good catch, and while the Trace Agent never calls setContext outside of runWithNewContext, it would be for the better that the function doesn't throw an error :P

src/cls/async-hooks.ts
```diff
@@ -14,15 +14,15 @@
  * limitations under the License.
  */
 
-// This file requires continuation-local-storage in the AsyncHooksCLS
-// constructor, rather than upon module load.
-import * as asyncHooksModule from 'async_hooks';
+// This file requires async_hooks in the AsyncHooksCLS constructor, rather than
+// upon module load.
```
jinwoo: Can you also comment that this import is only for types, and the actual `require` happens in the constructor?

src/cls/async-hooks.ts
jinwoo: If you intend `setContext()` to be always called from within `runWithNewContext()`, would it be better to assert here?

src/cls/async-hooks.ts
kjin: It sort of already says that. I think I worded it this way because referencing "this import" doesn't make sense for someone that is looking at the JS file itself, because the import won't be there.

jinwoo: Ah, now I get that. Backquoting `require` would've helped me catch that better :)

ofrobots: I would prefer if the import symbol was called `asyncHooksTypes` instead of `asyncHooks`. 

```
// This import is for types only. The actual `require` happens in the `AsyncHooksCLS` constructor.
// Do not use `asyncHooksTypes` directly.
```

src/cls/async-hooks.ts
```diff
@@ -31,29 +31,63 @@ const EVENT_EMITTER_METHODS: Array<keyof EventEmitter> =
...
-        this.contexts[id] = this.currentContext.value;
-      },
-      before: (id: number) => {
-        if (this.contexts[id]) {
-          this.currentContext.value = this.contexts[id];
+        // init is called when a new AsyncResource is created. We want code
+        // that runs within the scope of this new AsyncResource to see the same
+        // context as its "parent" AsyncResource. The criteria for the parent
+        // depends on the type of the AsyncResource.
+        if (type === 'PROMISE') {
```
ofrobots: Can we add a test to make sure that promises end up generating precisely this string as the resource type? If something in node-core changes, let us not break here.

The test can exist either here or in node-core.

src/cls/async-hooks.ts
```diff
@@ -64,51 +98,80 @@ export class AsyncHooksCLS<Context extends {}> implements CLS<Context> {
...
   getContext(): Context {
-    return this.currentContext.value;
+    // We don't store this.defaultContext directly in this.contexts.
+    // Getting undefined when looking up this.contexts means that it wasn't
+    // set, so return the default context.
+    const current = this.contexts[this.ah.executionAsyncId()];
+    return current ? current.value : this.defaultContext;
   }
 
   setContext(value: Context): void {
```
ofrobots: Comment for a follow-on PR: it would be good to merge this into `runWithNewContext` and implement a single `runInContext` function.

src/cls/async-hooks.ts
```diff
@@ -64,51 +98,80 @@ export class AsyncHooksCLS<Context extends {}> implements CLS<Context> {
...
+      const oldContext = that.contexts[id];
+      // Restore the captured context.
+      that.contexts[id] = boundContext;
       try {
         return fn.apply(this, arguments) as T;
       } finally {
-        current.value = oldContext;
+        // Revert the current context to what it was before it was set to the
+        // captured context.
+        that.contexts[id] = oldContext;
```
ofrobots: Comment for a follow-on PR. This snippet of code is repeated in `runInNewContext`. Is it possible to share?

src/cls/async-hooks.ts
```diff
@@ -64,51 +98,80 @@ export class AsyncHooksCLS<Context extends {}> implements CLS<Context> {
...
     }
   }
 
   bindWithCurrentContext<T>(fn: Func<T>): Func<T> {
-    if ((fn as ContextWrapped<Func<T>>)[WRAPPED] || !this.currentContext) {
+    // Capture the context of the current AsyncResource.
+    const boundContext = this.contexts[this.ah.executionAsyncId()];
+    // Return if we have already wrapped the function, or there is no current
+    // context to bind.
+    if ((fn as ContextWrapped<Func<T>>)[WRAPPED] || !boundContext) {
```
ofrobots: The first part of this check should be done before the `this.contexts` map is looked up; avoiding unnecessary lookups.

src/cls/async-hooks.ts
kjin: I'll revert it to `asyncHooksModule` as it was before. I don't think `asyncHooksTypes` is accurate because the type of the module is actually `typeof asyncHooksTypes` which doesn't read well.

Our CI should catch `asyncHooksTypes` being used directly because it'll crash in Node 6. I would prefer not to introduce comments that don't make sense in the transpiled code.

kjin: @jinwoo I've updated it to say `This file calls require('async_hooks') ...` hopefully that should be more clear.

src/cls/async-hooks.ts
kjin: Done. I think it should exist here, since it'll be hard to keep track of changes to the test elsewhere.

src/cls/async-hooks.ts
kjin: Added a TODO.

src/cls/async-hooks.ts
kjin: Added a TODO.

src/cls/async-hooks.ts
kjin: Done

PR #735, opened 2018-05-02
doc: mention that tracing doesn't work out-of-box with gcf

kjin:


kjin:
Not sure why codecov reports a CI failure. Should be unchanged anyway because this is a doc commit.

PR #736, opened 2018-05-02
fix: add build/src/cls in output files

kjin:


kjin:
@ofrobots It's the list of "reachable" d.ts definitions from `index.d.ts`. This is being detected by the install test which is run on each PR, since it type-checks (and only type-checks, not runs) the statement `import * as trace from '@google-cloud/trace-agent'`. What we could do to avoid the much more important problem of ensuring that the correct .js files are included is to actually run the code in the install test.

PR #738, opened 2018-05-03
test: update contributing guide and add require test

kjin:
This PR makes it so that `npm run test` does everything that CI does. It also adds an install test that ensures that evaluating `require('@google-cloud/trace-agent').start()` doesn't throw an error, which could have prevented 2.8.0 from being a bogus release.

The contributing guidelines have been updated accordingly.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,51 +1,20 @@
-# Building and Running Tests Locally
+# Contributing Guide
 
-To ensure that the Trace Agent works cross-platform, we use TypeScript build scripts located in the [`scripts`](scripts) directory. The entry point to these scripts is [`index.ts`](scripts/index.ts). The usage of this file is `ts-node -p ./scripts ./scripts/index.ts [command1] [...moreCommands]` (assuming that you are running in the repository root directory.)
+The Stackdriver Trace Agent is written in TypeScript. This means that when testing changes in your own application, you will need to run `npm run compile` to invoke the TypeScript compiler to compile your changes to JavaScript.
 
-The list of possible build commands is enumerated as `case` statements in `index.ts`, in addition to `npm-*` commands. See [`index.ts`](scripts/index.ts) for more details.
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -1,51 +1,20 @@
-# Building and Running Tests Locally
+# Contributing Guide
 
-To ensure that the Trace Agent works cross-platform, we use TypeScript build scripts located in the [`scripts`](scripts) directory. The entry point to these scripts is [`index.ts`](scripts/index.ts). The usage of this file is `ts-node -p ./scripts ./scripts/index.ts [command1] [...moreCommands]` (assuming that you are running in the repository root directory.)
+The Stackdriver Trace Agent is written in TypeScript. This means that when testing changes in your own application, you will need to run `npm run compile` to invoke the TypeScript compiler to compile your changes to JavaScript.
```
ofrobots: `when testing your own application`? I am really confused by this. is this about testing the Trace Agent or the user's application. The latter may not be in TS...

jinwoo: I'm confused. What does "your own application" mean here? An app that uses trace-agent, or trace-agent itself? A user's app can be written either in JS or in TS, right?

CONTRIBUTING.md
```diff
@@ -1,51 +1,20 @@
...
 
-The list of possible build commands is enumerated as `case` statements in `index.ts`, in addition to `npm-*` commands. See [`index.ts`](scripts/index.ts) for more details.
+The command `npm run test` tests code the same way that our CI will test it. This is a convenience command for a number of steps, which can run separately if needed:
```
jinwoo: Nit. Isn't `npm test` equivalent to `npm run test` and shorter?

CONTRIBUTING.md
kjin: It was supposed to mean "when testing changes to your local copy of the Trace Agent linked in your own application".

PR #739, opened 2018-05-04
refactor: make logger always last arg

kjin:
This PR just changes methods so that a logger, if any, is always the last argument. There should be no functional changes -- this is just a self-contained part of a larger change.


kjin:
@ofrobots Yes, that is the intention.

kjin:
- Node 10 failure is known
- AppVeyor failure is unrelated

Merging.

PR #741, opened 2018-05-07
 feat: add options to set the cls mechanism to async-hooks or async-listener

kjin:
~~This PR makes `async_hooks`-based Tracing the default option for tracing using Node 8+. To go back to using `async-listener` (via `continuation-local-storage`), new `config.clsMechanism` options are accepted: `async-listener` (and `async-hooks`).~~

This PR provides two new possible values for `config.clsMechanism`: `async-listener` and `async-hooks`.

Because context tracking is now specified as a config option rather than strictly as an environmental variable, a fair bit of refactoring is needed to support conditionally loading `continuation-local-storage` as late as possible. The constraints to satisfy are as following:

- `continuation-local-storage` needs to be loaded as early as possible, and definitely before any modules that do I/O.
- `continuation-local-storage` must be loaded after `start()` is called, so we know whether to load it (via the config option passed to `start`).

To accomplish this, most of the logic in `index.ts` has been moved to a new file, `tracing.ts`, where it is now encapsulated in a class `Tracing`. Barring the `feat:` change, there should be no behavioral changes; however there are some caveats that hopefully are captured in the added comments.

There is also a small change to prevent `PluginLoader#deactivate` from throwing when being called on an already deactivated plugin loader. This is only required to prevent large-scale changes to tests, as the case when it might be called when the plugin loader is de-activated is from test code that sets the `[FORCE_NEW]` flag in configuration objects.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015 Google Inc. All Rights Reserved.
+ * Copyright 2018 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -16,213 +16,91 @@
...
-  if (!config.enabled) {
-    return traceAgent;
+export function start(config?: Config): PluginTypes.TraceAgent {
+  // Determine the preferred context propagation mechanism, as
+  // continuation-local-storage should be loaded before any modules that do I/O.
+  const ahAvailable = semver.satisfies(process.version, '>=8');
+  const agentEnabled = !config || config.enabled !== false;
+  const alAutoPreferred =
+      !ahAvailable && (!config || config.clsMechanism === 'auto');
+  const alUserPreferred = config && (config.clsMechanism === 'async-listener');
```
ofrobots: Why not:
1. Keep `initConfig` in this file. As far as I can tell, it is synchronous and doesn't need external modules.
2. Modify `initConfig` with additional logic to convert the user preferences into a concrete (non-`auto`) value of `config.mechanism`.

In other words, I would prefer if `start` had logic like this:

```ts
function start(config?: Config) {
  const normalizedConfig = initConfig(config);
  if (!config.enabled) return;
  if (config.mechanism == 'async-listener') require('continuation-local-storage');
  // ...
}
```

The way you have written it right now, we don't properly deal with a config file provided through an environment variable.

ofrobots: You may need to change the `require` in `initConfig` to a `fs.readFileSync`.

If the above proposal is acceptable, are the rest of the changes (refactoring most of the logic in this file into `tracing.ts`) still needed? This is a pretty large change, and careful reasoning is needed to make sure it doesn't break behaviour. It is not immediately obvious to me why the refactoring is *required*.

ofrobots: The risk with the `require` that I was thinking of was if the users put arbitrary JS in there which ends up loading modules. I am not sure if it an important enough concern.

> For the rest of the refactoring -- I had been thinking about whether class-ifying Tracing was going overboard -- it's not strictly necessary

If it is not strictly necessary, I would prefer not to see it in *this* PR. This PR should look obviously correct to me, and currently it would require a fairly in depth review to assert that.

src/index.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015 Google Inc. All Rights Reserved.
+ * Copyright 2018 Google LLC
```
ofrobots: No need to change the copyright.

src/index.ts
```diff
@@ -106,123 +75,88 @@ function initConfig(projectConfig: Forceable<Config>):
...
-  }
-
-  if (typeof config.projectId !== 'string' &&
-      typeof config.projectId !== 'undefined') {
-    logger.error(
-        'TraceAgent#start: config.projectId, if provided, must be a string.',
-        'Disabling trace agent.');
-    stop();
-    return traceAgent;
+    throw e;
```
ofrobots: Just remove the catch block?

src/index.ts
kjin: To clarify, I believe moving logic to a separate file is the necessary part, while making it a class is less necessary. That being said, I think the change to make it a class is not going to negatively impact the diff view by a lot.

src/index.ts
```diff
@@ -106,123 +75,88 @@ function initConfig(projectConfig: Forceable<Config>):
...
+  if (!tracingSingleton) {
+    tracingSingleton = require('./tracing').tracing;
+  }
+  if (tracingSingleton.exists()) {
+    return tracingSingleton.get().traceAgent;
+  } else {
+    // This code path maintains the current contract that calling get() before
+    // start() yields a disabled custom span API. It assumes that the use case
+    // for doing so (instead of returning null) is when get() is called in
+    // a file where it is unknown whether start() has been called.
```
ofrobots: Hmmm.. this sounds odd. I don't understand the use case for this as you have described it. I don't think users would have any benefit of having a reference to a permanently disabled trace agent.

I think the reference we return from `get` will get 'enabled' once start is later called. I tested this theory out https://runkit.com/embed/3xg5kst085dt and it seems like my hunch is correct.

As such, this change is making an assumption that is incorrect.

kjin: Ok, I figured out a way to circumvent this (I didn't like it anyway and was hoping someone had ideas about how to change it).

src/index.ts
kjin: Done. I think I was about the change the error message but am not sure to what

src/index.ts
kjin: Done

src/tracing.ts
```diff
@@ -0,0 +1,173 @@
...
+  /** A logger. */
+  private readonly logger: common.Logger;
+  /** The configuration object for this instance. */
+  private readonly config: Forceable<NormalizedConfig>;
+
+  /**
+   * Constructs a new Tracing instance.
+   * @param config The configuration for this instance.
+   */
+  constructor(config: NormalizedConfig, traceAgent: TraceAgent) {
```
ofrobots: nit: make these `private readonly` here, remove assignments below and declarations above.

src/tracing.ts
```diff
@@ -0,0 +1,173 @@
...
+          `[${modulesLoadedBeforeTrace.sort().join(', ')}]`);
+    }
+  }
+
+  /**
+   * Enables automatic tracing support and the custom span API.
+   */
+  enable(): void {
+    if (this.traceAgent.isActive()) {
+      // For unit tests only.
```
ofrobots: How do we make sure this path is not reached during production? Ideally code like this should be guarded with a test that can only be true during testing. It is not immediately obvious to me why `this.traceAgent.isActive()` indicates that we are running inside a unit test.

src/tracing.ts
```diff
@@ -0,0 +1,173 @@
...
+
+      traceWriter.create(this.config, this.logger).initialize((err) => {
+        if (err) {
+          this.disable();
+        }
+      });
+
+      this.traceAgent.enable(this.config, this.logger);
+
+      pluginLoader.create(this.config, this.logger).activate();
```
ofrobots: These two statements occur after we have called `TraceWriter#initialize` – however they are independent of success/failure of `traceWriter#initialize` as they are not in the callback block. From a clarity point of view, can you make one of the following statements true:

1. If these are not dependent on the outcome of `traceWriter#initialize`, move the async operation (`traceWriter#initialize`) the last operation, and call `traceAgent.enable` and `pluginLoader.create` first.
1. If these are dependent, move them into the callback for `create`.

As written, it is not obvious whether it is an oversight – should they have been inside the callback block?

src/tracing.ts
```diff
@@ -0,0 +1,173 @@
...
+    } catch (e) {
```
ofrobots: That's a massive try block. What operation can actually throw. What operation can actually throw? How does it relate to the async error that comes from `TraceWriter#initialize`?

kjin: The Trace Writer and CLS components can throw. The rest can be outside the try-catch block.

src/tracing.ts
kjin: In response to the next comment I moved the initialization of the trace agent and plugin loader outside of the try-catch, so this probably doesn't apply anymore.

src/tracing.ts
kjin: The type of `config` is a little different (`this.config: Forceable<T>`; `config: T`). I moved `traceAgent`.

src/tracing.ts
kjin: Actually, this entire conditional isn't needed anymore. In `index.ts`, it can be statically determined that `tracing.enable()` is never called twice on the same instance.

src/tracing.ts
```diff
@@ -0,0 +1,166 @@
...
+
+  /**
+   * Constructs a new Tracing instance.
+   * @param config The configuration for this instance.
+   * @param traceAgent An object representing the custom tracing API.
+   */
+  constructor(
+      config: NormalizedConfig, private readonly traceAgent: TraceAgent) {
+    this.config = config;
+    this.traceAgent = traceAgent;
```
ofrobots: remove this line.

src/tracing.ts
```diff
@@ -0,0 +1,166 @@
...
+    if (!this.config.enabled) {
+      return;
+    }
+
+    try {
+      // Initialize context propagation mechanism.
+      const clsConfig: Forceable<TraceCLSConfig> = {
+        mechanism: this.config.clsMechanism as TraceCLSMechanism,
+        [FORCE_NEW]: this.config[FORCE_NEW]
+      };
```
ofrobots: nit: this statement doesn't need to be in the try block.

src/tracing.ts
```diff
@@ -0,0 +1,166 @@
...
+      cls.create(clsConfig, this.logger).enable();
+
+      traceWriter.create(this.config, this.logger).initialize((err) => {
+        if (err) {
+          this.disable();
+        }
+      });
+    } catch (e) {
```
ofrobots: I still think the error handling flow here needs to be explained. You're interested in catching and logging the synchronous errors throws by `traceWriter.create` but not the async ones given by `initialize`? Why the difference?

This suggests to me that `initialize` is not expected to throw. If so, why is it in the try block?

Aside: this code could be simplified in a follow-on if `initialize` was converted to an async function.

kjin: `initialize` is indeed not expected to throw. We can move it out of the `try` block.

What is the downside of keeping all of the logic together in the `try` block? The implementation of this function (`enable`) shouldn't depend on whether it's possible for a function to throw -- because those functions might have new conditions to throw later. I think trying to minimize the `try` block as much as possible leads to unneeded complexity in reading the code.

src/tracing.ts
kjin: done

PR #745, opened 2018-05-12
chore: fix types and stop testing on node 10

kjin:
This PR stop testing against Node 10 (which reports known failures) and also reacts to type changes introduced in `@types/node@10.0.8`.


kjin:
@jinwoo Makes sense. I won't revert the type changes though, because they were not introduced in Node 10 and might get backported to earlier type definition versions. (I believe they are still compatible.)

PR #747, opened 2018-05-14
feat: allow timestamps to be passed to endSpan

### LINKED ISSUE #456: Arbitrarily pass time spent to endSpan
We have a use case where we batch multiple calls to a database with a native batch API and in the response we get the `responseTime` for each individual call.

We are tracing how long that takes but we would like to track them singularly passing the number of milliseconds to endSpan or maybe some other method?

Thanks

cc/ @calufa

kjin:
Fixes #456

This PR adds an option to specify a time when a given span should be ended.

### DIFF
--- a/test/test-span-data.ts
+++ b/test/test-span-data.ts
@@ -109,7 +109,7 @@ describe('SpanData', () => {
       const spanData = new CommonSpanData(trace, 'name', '0', 0);
       const startTime = new Date(spanData.span.startTime).getTime();
       // This input Date is far enough in the future that it's unlikely that the
-      // timethis function was called could be close to it.
+      // time this function was called could be close to it.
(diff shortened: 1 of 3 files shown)


test/test-span-data.ts
```diff
@@ -106,6 +105,16 @@ describe('SpanData', () => {
           expectedTimes.slice().sort(ascending), expectedTimes);
     });
 
+    it('accepts a custom span end time', () => {
+      const spanData = new CommonSpanData(trace, 'name', '0', 0);
+      const startTime = new Date(spanData.span.startTime).getTime();
+      // This input Date is far enough in the future that it's unlikely that the
+      // timethis function was called could be close to it.
```
ofrobots: nit: `time this`

PR #766, opened 2018-06-01
fix: inject context http headers early if expect header exists

### LINKED ISSUE #763: NetworkingError: Can't set headers after they are sent.
Hi, we intermittently had this issue while uploading a file to S3 using aws-sdk.
```
NetworkingError: Can't set headers after they are sent.
    at ClientRequest.setHeader (_http_outgoing.js:371:11)
    at Object.requestTrace [as request] (/data/node_modules/@google-cloud/trace-agent/build/src/plugins/plugin-http.js:124:13)
    at Object.request (https.js:206:15)
    at features.constructor.handleRequest (/data/node_modules/aws-sdk/lib/http/node.js:42:23)
    at executeSend (/data/node_modules/aws-sdk/lib/event_listeners.js:304:29)
    at Request.SEND (/data/node_modules/aws-sdk/lib/event_listeners.js:318:9)
    at Request.callListeners (/data/node_modules/aws-sdk/lib/sequential_executor.js:101:18)
    at Request.emit (/data/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
    at Request.emit (/data/node_modules/aws-sdk/lib/request.js:683:14)
    at Request.transition (/data/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/data/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /data/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/data/node_modules/aws-sdk/lib/request.js:38:9)
    at Request.<anonymous> (/data/node_modules/aws-sdk/lib/request.js:685:12)
    at Request.callListeners (/data/node_modules/aws-sdk/lib/sequential_executor.js:115:18)
    at callNextListener (/data/node_modules/aws-sdk/lib/sequential_executor.js:95:12)
    at /data/node_modules/aws-sdk/lib/event_listeners.js:220:9
    at finish (/data/node_modules/aws-sdk/lib/config.js:320:7)
    at /data/node_modules/aws-sdk/lib/config.js:338:9
    at Credentials.get (/data/node_modules/aws-sdk/lib/credentials.js:126:7)
    at getAsyncCredentials (/data/node_modules/aws-sdk/lib/config.js:332:24)
    at Config.getCredentials (/data/node_modules/aws-sdk/lib/config.js:352:9)
    at Request.SIGN (/data/node_modules/aws-sdk/lib/event_listeners.js:192:22)
    at Request.callListeners (/data/node_modules/aws-sdk/lib/sequential_executor.js:101:18)
    at Request.emit (/data/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
    at Request.emit (/data/node_modules/aws-sdk/lib/request.js:683:14)
    at Request.transition (/data/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/data/node_modules/aws-sdk/lib/state_machine.js:14:12)
```
The stack points here: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/src/plugins/plugin-http.ts#L147-L148 
I have to admit I haven't collected much information at the moment, so tell me if there is something you'd like to know to understand the issue better.
Any idea on what's happening and how to solve it?
Thanks!

( @Macr0s FYI )

kjin:
Fixes #763

This PR fixes a bug where our HTTP instrumentation fails if `Expect` is added as an outgoing header. This is because we normally inject a trace context header for a request as part of our HTTP instrumentation, but Node has a special case where it prohibits additional headers if the `Expect` header is detected. See: https://github.com/nodejs/node/blob/v7.10.1/lib/_http_outgoing.js#L323

To fix this, I brought back the old "slower" way of injecting trace context headers that was removed in #643, but only if the `Expect` header is detected.

I also added a test that fails without the source change.

### DIFF
--- a/src/plugins/plugin-http.ts
+++ b/src/plugins/plugin-http.ts
@@ -42,6 +42,16 @@ function getSpanName(options: ClientRequestArgs|url.URL) {
   return options.hostname || options.host || 'localhost';
 }
 
+/**
+ * Returns whether the Expect header is on the given options object.
+ * @param options Options for http.request.
+ */
+function hasExpectHeader(options: ClientRequestArgs|
+                         url.URL): options is ClientRequestArgs {
+  // tslint:disable-next-line:no-any
+  return !!((options as any).headers && (options as any).headers.Expect);
+}
+
 function extractUrl(
     options: ClientRequestArgs|url.URL, fallbackProtocol: string) {
   let path;
@@ -111,6 +121,24 @@ function makeRequestTrace(
     span.addLabel(api.labels.HTTP_METHOD_LABEL_KEY, method);
     span.addLabel(api.labels.HTTP_URL_LABEL_KEY, uri);
 
+    // If outgoing request headers contain the "Expect" header, the returned
+    // ClientRequest will throw an error if any new headers are added. For this
+    // reason, only in this scenario, we opt to clone the options object to
+    // inject the trace context header instead of using ClientRequest#setHeader.
+    // (We don't do this generally because cloning the options object is an
+    // expensive operation.)
+    let traceHeaderPreinjected = false;
+    if (hasExpectHeader(options)) {
+      traceHeaderPreinjected = true;
+      // "Clone" the options object -- but don't deep-clone anything except for
+      // headers.
+      options = Object.assign({}, options);
+      options.headers = Object.assign({}, options.headers);
+      // Inject the trace context header.
+      options.headers[api.constants.TRACE_CONTEXT_HEADER_NAME] =
+          span.getTraceContext();
+    }
+
     const req = request(options, (res) => {
       api.wrapEmitter(res);
       let numBytes = 0;
@@ -144,13 +172,17 @@ function makeRequestTrace(
       }
     });
     api.wrapEmitter(req);
-    req.setHeader(
-        api.constants.TRACE_CONTEXT_HEADER_NAME, span.getTraceContext());
     req.on('error', error => {
       span.addLabel(api.labels.ERROR_DETAILS_NAME, error.name);
       span.addLabel(api.labels.ERROR_DETAILS_MESSAGE, error.message);
       span.endSpan();
     });
+    // Inject the trace context header, but only if it wasn't already injected
+    // earlier.
+    if (!traceHeaderPreinjected) {
+      req.setHeader(
+          api.constants.TRACE_CONTEXT_HEADER_NAME, span.getTraceContext());
+    }
     return req;
   };
 }
--- a/src/plugins/plugin-http.ts
+++ b/src/plugins/plugin-http.ts
@@ -46,8 +46,7 @@ function getSpanName(options: ClientRequestArgs|url.URL) {
  * Returns whether the Expect header is on the given options object.
  * @param options Options for http.request.
  */
-function hasExpectHeader(options: ClientRequestArgs|
-                         url.URL): options is ClientRequestArgs {
+function hasExpectHeader(options: ClientRequestArgs|url.URL): boolean {
   return !!(
       (options as ClientRequestArgs).headers &&
       (options as ClientRequestArgs).headers!.Expect);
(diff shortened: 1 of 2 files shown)


src/plugins/plugin-http.ts
```diff
@@ -42,6 +42,16 @@ function getSpanName(options: ClientRequestArgs|url.URL) {
   return options.hostname || options.host || 'localhost';
 }
 
+/**
+ * Returns whether the Expect header is on the given options object.
+ * @param options Options for http.request.
+ */
+function hasExpectHeader(options: ClientRequestArgs|
+                         url.URL): options is ClientRequestArgs {
```
jinwoo: Should the return type a `boolean`?

src/plugins/plugin-http.ts
```diff
@@ -42,6 +42,16 @@ function getSpanName(options: ClientRequestArgs|url.URL) {
...
+  // tslint:disable-next-line:no-any
+  return !!((options as any).headers && (options as any).headers.Expect);
```
jinwoo: I think it makes more sense to type-cast as `ClientRequestArgs` rather than as `any`.

src/plugins/plugin-http.ts
```diff
@@ -111,6 +121,24 @@ function makeRequestTrace(
...
+    // reason, only in this scenario, we opt to clone the options object to
+    // inject the trace context header instead of using ClientRequest#setHeader.
+    // (We don't do this generally because cloning the options object is an
+    // expensive operation.)
+    let traceHeaderPreinjected = false;
+    if (hasExpectHeader(options)) {
+      traceHeaderPreinjected = true;
+      // "Clone" the options object -- but don't deep-clone anything except for
+      // headers.
+      options = Object.assign({}, options);
```
jinwoo: Nit. `options = {...options};` would be simpler.

src/plugins/plugin-http.ts
kjin: `A is B` is already a `boolean`.

src/plugins/plugin-http.ts
kjin: Done

src/plugins/plugin-http.ts
kjin: I think `...` syntax isn't available in Node 4?

kjin:
@jinwoo I'm not sure. It could be unintentional -- I think I need to understand more about the `Expect` header first.

src/plugins/plugin-http.ts
jinwoo: Yeah, I know. What I mean here is that this function does NOT test that `options` is a `ClientRequestArgs` object but rather that it is a one AND it has an `Expect` header. Right?

src/plugins/plugin-http.ts
jinwoo: I believe TS transpiles it to what you wrote here.

src/plugins/plugin-http.ts
```diff
@@ -49,7 +49,9 @@ function getSpanName(options: ClientRequestArgs|url.URL) {
 function hasExpectHeader(options: ClientRequestArgs|
                          url.URL): options is ClientRequestArgs {
   // tslint:disable-next-line:no-any
```
jinwoo: You don't need this anymore.

kjin:
@jinwoo Seems like the behavior for a long time (since Node 0.3) has been that Node eagerly sends headers when `Expect` exists. https://github.com/nodejs/node/blob/d59512f6f406ceae4940fd9c83e8255f0c03173b/lib/http.js#L463

jinwoo:
That makes sense because `Expect` means that the client is expecting a 100 response before sending the request body.

src/plugins/plugin-http.ts
jinwoo: Forgot to mention that this makes the compiler make bad decisions about type info. E.g.
```ts
const options = ...;
...
if (hasExpectHeader(options)) {
  // Compiler thinks `options` is a ClientRequestArgs, which is correct.
} else {
  // Compiler thinks `options` is NOT a ClientRequestArgs, which may or may not be correct.
}
```

PR #772, opened 2018-06-06
fix: update to @google-cloud/common@0.19

### LINKED ISSUE #638: Let users provide Project Id via GOOGLE_APPLICATION_CREDENTIALS
###### Copied from original issue: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2787

<a href="/JustinBeckwith"></a>@JustinBeckwith<br>January 8, 2018 9:58 PM

Today in cloudcats I am providing the ProjectId via the PROJECT_ID env var.  I also have to provide a keyfile, which I was passing into the code.  In an effort to make things a little more smooth on the config side,  I tried to use GOOGLE_APPLICATION_CREDENTIALS instead of defining PROJECT_ID and the keyfile.  The projectId is inside of my keyfile - but we still seem to require the ProjectId along with the key file :(

```
beckwith-macbookpro:web beckwith$ npm start

> cloudcats-web@1.0.0 start /Users/beckwith/Code/cloudcats/web
> GOOGLE_APPLICATION_CREDENTIALS=keyfile.json node server.js

info: Server running at: http://0.0.0.0:8080
ERROR:@google-cloud/trace-agent: Unable to acquire the project number from metadata service. Please provide a valid project number as an env. variable, or through config.projectId passed to start(). Error: getaddrinfo ENOTFOUND metadata.google.internal metadata.google.internal:80
ERROR:@google-cloud/debug-agent: The project ID could not be determined: getaddrinfo ENOTFOUND metadata.google.internal metadata.google.internal:80
```

Can we just read the projectId from the keyfile?

kjin:
Fixes #764 and fixes #638

This PR upgrades the dependency on `@google-cloud/common`.

### Why are there so many test changes?

This PR changes implementation details of `TraceWriter`. Some tests assumed a certain order of precedence in which `TraceWriter` would look for project IDs; others assumed that traces got buffered synchronously after being passed to `writeSpan`. These are both implementation details, and they both changed.

As a result of these now invalid assumptions (especially the second one) virtually every test, except the ones that had been written/re-written recently, failed. I got through refactoring some of them before putting a stop-gap in `TraceWriter` -- it is far too painful to edit every test in one fell swoop.

In particular, the Trace Writer test `test-trace-writer` has been completely re-written to be more behavior-driven.


src/trace-writer.ts
```diff
@@ -82,7 +82,9 @@ export class TraceWriter extends common.Service {
         config);
 
     this.logger = logger;
-    this.config = config;
+    // Clone the config object
+    this.config = Object.assign({}, config);
```
jinwoo: Believe me. You can say `this.config = {...config}` and TS will transpile it correctly.

src/trace-writer.ts
```diff
@@ -283,7 +264,21 @@ export class TraceWriter extends common.Service {
...
         setImmediate(() => this.flushBuffer());
       }
-    });
+    };
+    // TODO(kjin): We should always be following the 'else' path.
+    // Any test that doesn't mock the Trace Writer will assume that traces get
+    // buffered synchronously. We need to refactor those tests to remove that
+    // assumption before we can make this fix.
+    if (this.config.projectId) {
+      afterProjectId(this.config.projectId);
```
jinwoo: Maybe add a warning log here so that if this code path is executed in production, people can notice?

src/trace-writer.ts
```diff
@@ -283,7 +264,21 @@ export class TraceWriter extends common.Service {
...
+    } else {
+      this.getProjectId().then(afterProjectId, (err: Error) => {
+        this.logger.info(
+            'TraceWriter#queueTrace: No project ID, dropping trace.');
+        return;  // if we even reach this point, disabling traces is already
```
jinwoo: What does this `return` do? And what does the comment mean?

test/logger.ts
```diff
@@ -14,26 +14,35 @@
...
-           string[]} = {error: [], warn: [], info: [], debug: [], silly: []};
-  private innerLogger = logger({level: logger.LEVELS[PASS_THROUGH_LOG_LEVEL]});
+export class TestLogger extends Logger {
+  private logs: {[k in keyof Logger]: string[]} =
+      {silent: [], error: [], warn: [], info: [], debug: [], silly: []};
+  private innerLogger =
+      new OriginalLogger({level: logger.LEVELS[PASS_THROUGH_LOG_LEVEL]});
 
-  private makeLoggerFn(logLevel: keyof Logger): LoggerFunction {
+  constructor(options?: Partial<LoggerConfig>) {
```
jinwoo: Why do you need this ctor when you only calls `super(options)`?

src/trace-writer.ts
kjin: You're right, I stand corrected.

src/trace-writer.ts
kjin: The unstated corollary to `We should always be following the 'else' path.` is `we are currently almost always following the 'if' path`. AFAIK there should be nothing user-visibly different from before; this TODO just comments on the unintuitive reason for branching.

test/logger.ts
kjin: TSC will compile an implicit constructor to an explicit one using spread args if compiling to ES6.

src/trace-writer.ts
jinwoo: LOL. I thought you meant that `if` path was executed only from tests. Is this comment useful? I was easily confused.

kjin: It's basically a long way of saying "I don't like how it's written right now, I want to change it later" 😅

test/logger.ts
jinwoo: Really? `Logger`'s ctor looks exactly the same as this one: https://github.com/googleapis/nodejs-common/blob/master/src/logger.ts#L70

TS still uses spread args? And why does it matter with this test?

kjin: It's all about Node 4 compatibility...

I think nodejs-common compiles to ES5. We can't do that since we use generators in some tests, so we just don't use ES6 features that aren't supported in Node 4.

PR #775, opened 2018-06-13
feat: support tracing for untranspiled async/await in Node 8+

### LINKED ISSUE #566: async/await
The next version of hapijs (v17) will be async/await only. Is this supported by this tracing module?

kjin:
Fixes #566

This PR removes references to `GCLOUD_TRACE_NEW_CONTEXT`. Code that depend on its value now behave as if it were always `true`.

This change makes `async_hooks`-based tracing the default behavior in Node 8+.


package.json
```diff
@@ -55,7 +55,7 @@
     "@types/glob": "^5.0.32",
     "@types/is": "0.0.20",
     "@types/methods": "^1.1.0",
-    "@types/mocha": "^5.0.0",
+    "@types/mocha": "5.2.1",
```
kjin: Currently need to keep this version fixed. 5.2.2 seems to have compilation issues.

kjin:
@ofrobots Yes

PR #782, opened 2018-06-20
feat: add getProjectId and getCurrentRootSpan

kjin:
This PR adds two methods on the TraceAgent class, `getProjectId` and `getCurrentRootSpan`, which should supercede `getWriterProjectId` and `getCurrentContextId` respectively. It also uses `TraceWriter#projectId` instead of `TraceWriter#config.projectId` to store the retrieved project ID, because this is already the behavior of `Service`. In practice, this fixes a regression with `getProjectId`.

### DIFF
--- a/src/trace-api.ts
+++ b/src/trace-api.ts
@@ -16,15 +16,14 @@
 
 import {Logger} from '@google-cloud/common';
 import * as is from 'is';
-import * as semver from 'semver';
 import * as uuid from 'uuid';
 
 import {cls, RootContext} from './cls';
 import {Constants, SpanType} from './constants';
 import {Func, RootSpan, RootSpanOptions, Span, SpanOptions, TraceAgent as TraceAgentInterface} from './plugin-types';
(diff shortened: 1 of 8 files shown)


src/trace-api.ts
```diff
@@ -173,21 +171,37 @@ export class TraceAgent implements TraceAgentInterface {
...
+      return UNTRACED_ROOT_SPAN;
+    }
+    return cls.get().getContext();
+  }
+
+  getCurrentContextId(): string|null {
+    // In v3, this will be deprecated for getCurrentRootSpan.
+    const traceContext = this.getCurrentRootSpan().getTraceContext();
+    const parsedTraceContext = util.parseContextFromHeader(traceContext);
+    if (parsedTraceContext) {
```
DominicKramer: nit: What do you think about using a ternary operator instead?
```
  return parsedTraceContext ? parsedTraceContext.traceId : null
```

PR #784, opened 2018-06-21
feat: add method to get configuration from traceApi

### LINKED ISSUE #444: Is there a way to get only some of the enhanced database reporting?
In the pg plugin, the query and the values are added to the span only if enhanced database reporting is enabled. (https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/src/plugins/plugin-pg.js#L36-L41). While we would like to have the queries reported, reporting the values and rows from the response could end up being a little too much information (think of changing a password or access token). 

Is there a way around this without manually patching pg and disabling the experimental plugin?

Thanks!

kjin:
Fixes #444

The new method allows plugins to get arbitrary information from the user, including options to toggle how much database information gets reported in a trace.


src/config.ts
```diff
@@ -80,6 +80,14 @@ export interface Config {
    */
   plugins?: {[pluginName: string]: string;};
 
+  /**
+   * An object that specifies arbitrary options on how to patch a module for
+   * instrumentation. It will be available to plugins through
+   * TraceAgent#getPluginOptions.
+   */
```
ofrobots: From this comment I am trying to figure out how users are supposed to use this. Are these options for a single plugin or aggregate options for all plugins? Is this a map where the plugin name will be the key?

I think this needs example.

src/plugin-types.ts
```diff
@@ -110,6 +111,13 @@ export interface TraceAgent {
    */
   enhancedDatabaseReportingEnabled(): boolean;
 
+  /**
+   * Gets the plugin options object. This object should not be modified.
```
ofrobots: should we freeze the object instead of using a comment?

PR #787, opened 2018-06-23
doc: start generating docs

kjin:
Starting with this PR when it lands, any future changes to master will trigger a CI job that re-builds documentation. A preview of what this would look like is available [here](https://kjin.github.io/cloud-trace-nodejs/).


kjin:
@googleapis/node-team I think this is finally ready. PTAL.

ofrobots:
It would be good to have docs. However, @kinwa91 has been thinking about doc generation more broadly across all the client libraries. I want to make sure this is not going to be divergent from that effort. /cc @JustinBeckwith .

JustinBeckwith:
I don't think we want to do something custom - would it be cool if we closed this out?

kinwa91:
I'm not entirely certain that the documentation for this repository (along with other cloud-*-nodejs, supporting libraries etc) should end up together with the rest of the `nodejs-*` cloud client libraries.

We currently only have documentations for Cloud client libraries on `cloud.google.com`, and my proposal will gather them together and move them (back) to `googleapis.github.io/google-cloud-node`, just like other languages, which will ultimately move out of GitHub Pages. 

If we want to, I can include the Stackdriver agents into the Cloud Client Libraries documentation site, or alternatively we can set them up under their own repo. Either way, can still have a single process for generating documentation across the GitHub org, regardless of where it is hosted.

sduskis:
@kjin @googleapis/yoshi-nodejs, this PR is a bit out of date.  Do we still need this?

ofrobots:
I am not quite clear on what distinguishes an agent from a client library (e.g. take Error Reporting as an example). Even this repo offers a very very high level API abstraction over the Trace service (albeit only the 'patchTraces' endpoint). Also, this classification can be a point-in-time thing. In the future this library may (or may not) end with more endpoints supported and become less distinguishable from a client library.

JustinBeckwith:
Nope nope nope.  Whatever solution we come up with for the rest of our npm modules should be what we use here :) This repo (and the other cloud- ones) will be included in that process, which will land in Q1.  I suggest we close this out 🤷‍♂️

Issue #798, opened 2018-07-13
Update gRPC tracing context key/value format


kjin:

kjin:
Fixed in #814

PR #799, opened 2018-07-14
refactor: convert plugin-grpc.ts to typescript

kjin:
This is a nice-to-have prereq for changing the grpc metadata context header as in #798.


src/plugins/plugin-grpc.ts
```diff
@@ -14,38 +14,108 @@
...
+import * as grpcModule from 'grpc';  // for types only.
+import {Client, MethodDefinition, ServerReadableStream, ServerUnaryCall, StatusObject} from 'grpc';
+import * as shimmer from 'shimmer';
 
-var shimmer = require('shimmer');
-var findIndex = require('lodash.findindex');
+import {Plugin, RootSpan, RootSpanOptions, Span, TraceAgent} from '../plugin-types';
 
-var SKIP_FRAMES = 1;
+// Re-definition of Metadata with private fields
```
ofrobots: Is there a way to get gRPC to ship a second module with just the types that could be used a devDependency here?

ofrobots: I read code wrong.

tsconfig.full.json
```diff
@@ -8,7 +8,8 @@
     "noUnusedParameters": false,
     "noImplicitAny": false,
     "noImplicitThis": false,
-    "noEmitOnError": false
+    "noEmitOnError": false,
+    "lib": ["dom", "es2015"]
```
ofrobots: why is `dom` needed?

kjin: `@types/protobufjs` requires it. Sorry, was about to mention that and then forgot.

ofrobots: Why is `@types/protobufjs` needed?

kjin: It's a transitive dependency for `grpc` types.

ofrobots: *sigh*.

ofrobots: I have two questions:
* Why is grpc leaking this to its users? `grpc-node` does not support the browser and requiring their users to have browser specific TypeScript config seems "wrong".
* I am fairly convinced that this module wouldn't leak the dependency on `dom` to our users, but can you confirm?

kjin: > Why is grpc leaking this to its users? `grpc-node` does not support the browser and requiring their users to have browser specific TypeScript config seems "wrong".
That we currently expose `protobufjs` options to the user is a symptom of the problem that is exactly what `@grpc/proto-loader` is trying to solve. At best we can re-define the options in the `grpc` types. (Actually, if you think that's a good idea I can go ahead and do that.)
> I am fairly convinced that this module wouldn't leak the dependency on `dom` to our users, but can you confirm?
That's a valid concern -- it doesn't because the package currently doesn't publish types for the gRPC patching plugin.

PR #800, opened 2018-07-16
fix: improve logs from the trace writer

### LINKED ISSUE #797: Please log the hostname on failed requests
I've spent the past our trying to figure out (still haven't) why am I getting these errors on Istio –which is a service mesh that requires users to explicitly whitelist all the external hosts that are called.

> ERROR:@google-cloud/trace-agent: TraceWriter#publish: Received error status code unknown. Original error: Error: Unexpected error while acquiring application default credentials: Could not load the default credentials. Browse to https://developers.google.com/accounts/docs/application-default-credentials for more information.

This error doesn't explain why it failed. `Received error status code unknown. Original error: Error: Unexpected error` might be one of the most _unhelpful_ errors to return here.

Here are some better examples:

- _Failed to upload traces: Could not retrieve credentials: could not connect to metadata service: TLS handshake failed_
- _Failed to upload traces: Could not retrieve credentials: could not connect to metadata service: unknown host metadata.google.internal_

Here's another unhelpful error:

> TraceWriter#publish: Received error status code unknown. Original error: Error: Sorry, we cannot connect to Cloud Services without a project ID. You may specify one with an environment variable named  "GOOGLE_CLOUD_PROJECT".

Why is it unhelpful? I can't tell:

- was it trying to find the project ID (from gce metadata) and that's why it failed?
- did it actually failed to connect cloudtrace.googleapis.com
  - if so, which port/protocol (grpc/https) did it use?

Here's a suggestion:

- _TraceWriter#publish: Failed to determine project ID from metadata service: unknown host metadata.google.internal_
- _TraceWriter#publish: Failed to upload traces (cloudtrace.googleapis.com:443): tls handshake failed_

kjin:
Fixes #797

This PR adds hostnames to error messages coming from the Trace Writer. It also changes one error message to not print "status code unknown" (this is misleading because unknown actually means an error was thrown when making an outgoing request).

Other errors are forwarded from the `@google-cloud/common` module, so any additional changes to improve error messages would need to originate there.

cc/ @ahmetb


src/trace-writer.ts
```diff
@@ -331,10 +333,13 @@ export class TraceWriter extends common.Service {
...
     this.request(options, (err, body?, response?) => {
-      const statusCode = (response && response.statusCode) || 'unknown';
+      const statusCode = response && response.statusCode;
       if (err) {
-        this.logger.error(`TraceWriter#publish: Received error status code ${
-            statusCode}. Original error: ${err}`);
+        this.logger.error(`TraceWriter#publish: Received error ${
+            statusCode ?
+                `with status code ${statusCode}` :
+                ''} while publishing traces to cloudtrace.googleapis.com: ${
```
ofrobots: nit: would it make sense to use ${uri} here instead of duplicating the address? Or maybe extract a hostname variable.

PR #812, opened 2018-07-18
feat: implement (de)serialization of binary trace context as utils

kjin:
These functions will be used by gRPC tracing to (de)serialize binary trace context.

Because JS doesn't support the full range of 64-bit numbers, I've added a dependency on [`hex2dec`](https://www.npmjs.com/package/hex2dec) which allows us to convert from arbitrary-length hex to decimal strings and vice-versa.


src/util.ts
```diff
@@ -241,3 +237,52 @@ export function createStackTrace(
...
+ * @param traceContext The trace context to serialize.
+ */
+export function serializeTraceContext(traceContext: TraceContext): Buffer {
+  //      0 byte version (0)
+  //      1 traceID field ID (0)
+  //  2->17 traceID value (traceContext.traceID)
+  //     18 spanID field ID (1)
+  // 19->26 spanID value (traceContext.spanID)
+  //     27 options field ID (2)
+  //     28 options value (traceContext.options)
```
ofrobots: nit: use this
```
 0           1           2
 0 1 2345678901234567 8 90123456 7 8
-------------------------------------
| | |                | |        | | |
-------------------------------------
 ^ ^      ^           ^    ^     ^ ^
 | |      |           |    |     | `--- options value (traceContext.options)
 | |      |           |    |     `----- options field id (2)
 | |      |           |    `----------- spanID value (traceConext.spanID)
 | |      |           `---------------- spanID field ID (1)
 | |      `---------------------------- traceID value (traceContext.traceID)
 | `----------------------------------- traceID field ID (0)
 `------------------------------------- version (0)
```

PR #814, opened 2018-07-18
feat: use well-known format for propagating trace context thru grpc

kjin:
BREAKING CHANGE: The change in distributed trace context propagation across gRPC is not backwards-compatible. In other words, distributed tracing will not work between two Node instances communicating using gRPC with v2 and v3 of the Trace Agent, respectively.

Trace context can now be propagated thru gRPC with the `'grpc-trace-bin'` key, and a binary-encoded value.

### DIFF
--- a/src/plugins/plugin-grpc.ts
+++ b/src/plugins/plugin-grpc.ts
@@ -112,12 +112,13 @@ function patchClient(client: ClientModule, api: TraceAgent) {
    * Set trace context on a Metadata object if it exists.
    * @param metadata The Metadata object to which a trace context should be
    * added.
-   * @param span The span that contains the trace context to add as a metadata
-   * entry.
+   * @param stringifiedTraceContext The stringified trace context. If this is
+   * a falsey value, metadata will not be modified.
    */
-  function setTraceContextFromString(metadata: Metadata, span: Span): void {
+  function setTraceContextFromString(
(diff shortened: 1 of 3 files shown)


kjin:
cc/ @rakyll

src/plugins/plugin-grpc.ts
```diff
@@ -108,6 +108,24 @@ function unpatchMetadata() {
...
 
 function patchClient(client: ClientModule, api: TraceAgent) {
+  /**
+   * Set trace context on a Metadata object if it exists.
+   * @param metadata The Metadata object to which a trace context should be
+   * added.
+   * @param span The span that contains the trace context to add as a metadata
+   * entry.
+   */
+  function setTraceContextFromString(metadata: Metadata, span: Span): void {
```
ofrobots: nit: based on the name, I would have expected the second parameter to be a string. The only thin we do with `span` in this function is to call `getTraceContext` on it. Could the caller do this call?

kjin: Makes sense. The way you said also mirrors the corresponding function on the server side better.

ofrobots:
Please make sure that the commit - at landing time - includes 'BREAKING CHANGE:' prefix in the body as per conventionalcommits.org.

PR #816, opened 2018-07-18
feat: deprecate Tracer methods

kjin:
See #782, which introduces methods that will replace these.


kjin:
@ofrobots Thanks for looking at this. Seeing how the other functions are being used in Bunyan Logging, it doesn't make sense to deprecate them. I'll close this.

PR #819, opened 2018-07-24
feat: read default exports from a plugin if it exists

kjin:
In order to allow plugins to export values other than the monkeypatch itself, this PR makes the plugin loader accept the `default` exported value if it exists.

Also, this adds a top-level export `defaultConfig`, which exposes the default configuration options.

### DIFF
--- a/src/trace-plugin-loader.ts
+++ b/src/trace-plugin-loader.ts
@@ -192,7 +192,14 @@ export class ModulePluginWrapper implements PluginWrapper {
   // Helper function to get the cached plugin value if it wasn't loaded yet.
   getPluginExportedValue(): Plugin {
     if (this.pluginExportedValue === ModulePluginWrapper.NOT_LOADED) {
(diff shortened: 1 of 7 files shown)


src/trace-plugin-loader.ts
```diff
@@ -192,7 +192,14 @@ export class ModulePluginWrapper implements PluginWrapper {
   // Helper function to get the cached plugin value if it wasn't loaded yet.
   getPluginExportedValue(): Plugin {
     if (this.pluginExportedValue === ModulePluginWrapper.NOT_LOADED) {
-      this.pluginExportedValue = require(this.path);
+      const moduleExports = require(this.path);
+      // Use the default export if it exists. Otherwise, assume that the
+      // entirety of the module exports is the plugin value itself.
+      if (moduleExports.default) {
```
ofrobots: It would be good to not use `default` exports – they style guide forbids them.

kjin:
Closed in favor of #826.

PR #821, opened 2018-07-25
feat: allow incoming http header to override span name

kjin:
This PR introduces `Tracer#constants.TRACE_SPAN_NAME_OVERRIDE` which, when used as a header key on an incoming request, changes the span name to the value of this header.


kjin:
@ofrobots I added a config option that turns this feature on. By default it is off.

kjin:
Closed in favor of #826.

PR #826, opened 2018-07-31
 feat: add rootSpanNameOverride option

kjin:
This PR introduces `rootSpanNameOverride`, which allows us to override span names differently than in #821.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -63,13 +63,13 @@ export interface Config {
   enhancedDatabaseReporting?: boolean;
 
   /**
-   * If set to true, traces that represent incoming HTTP spans will prefer
-   * the header value under key Tracer#constants.TRACE_SPAN_NAME_OVERRIDE to
-   * the request path. This option should only be enabled if this server's
-   * endpoints aren't publicly exposed, as it allows misbehaving clients to
-   * set span names to arbitrary values.
+   * A value that can be used to override names of spans that represent
+   * incoming requests. If specified as a string, the string will be used
+   * to replace all such span names; if specified as a function, the function
+   * will be invoked with the request path as an argument, and its return value
+   * will be used as the span name.
    */
-  useSpanNameOverrideHeader?: boolean;
+  incomingRequestSpanNameOverride?: string|((path: string) => string);
 
   /**
    * The maximum number of characters reported on a label value. This value
@@ -202,7 +202,7 @@ export const defaultConfig = {
   logLevel: 1,
   enabled: true,
   enhancedDatabaseReporting: false,
-  useSpanNameOverrideHeader: false,
+  incomingRequestSpanNameOverride: (path: string) => path,
   maximumLabelValueSize: 512,
   plugins: {
     // enable all by default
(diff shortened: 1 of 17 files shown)


ofrobots:
Why are we still supporting the header for an override?

kjin:
@ofrobots Whoops, removed.

src/config.ts
```diff
@@ -62,6 +62,15 @@ export interface Config {
...
   enhancedDatabaseReporting?: boolean;
 
+  /**
+   * A value that can be used to override names of spans that represent
+   * incoming requests. If specified as a string, the string will be used
+   * to replace all such span names; if specified as a function, the function
+   * will be invoked with the request path as an argument, and its return value
+   * will be used as the span name.
+   */
+  incomingRequestSpanNameOverride?: string|((path: string) => string);
```
ofrobots: http requests only? What about http2, grpc, or other incoming requests?

test/test-trace-web-frameworks.ts
```diff
@@ -334,6 +336,29 @@ describe('Web framework tracing', () => {
               `span name ${serverSpan.name} includes query parameters`);
         });
       });
+
+      it('uses the span name override header when assoc. option is on',
```
ofrobots: header?????

test/test-trace-web-frameworks.ts
```diff
@@ -334,6 +336,29 @@ describe('Web framework tracing', () => {
...
+         async () => {
+           const oldSpanNameOverride = testTraceModule.get()
+                                           .getConfig()
+                                           .incomingRequestSpanNameOverride;
+           testTraceModule.get().getConfig().incomingRequestSpanNameOverride =
+               (path: string) => `${path}-goodbye`;
+           const spanNameOverrideKey =
```
ofrobots: it seems like this test isn't fully baked.

kjin:
@ofrobots I changed it so that root spans are affected, not HTTP incoming requests. The code is now in the `Tracer` implementation.

src/index.ts
```diff
@@ -76,6 +76,13 @@ function initConfig(projectConfig: Forceable<Config>):
       Constants.TRACE_SERVICE_LABEL_VALUE_LIMIT) {
     config.maximumLabelValueSize = Constants.TRACE_SERVICE_LABEL_VALUE_LIMIT;
   }
+  // Make rootSpanNameOverride a function if not already.
+  if (typeof config.rootSpanNameOverride === 'string') {
+    const spanName = config.rootSpanNameOverride;
+    config.rootSpanNameOverride = () => spanName;
+  } else if (typeof config.rootSpanNameOverride !== 'function') {
+    config.rootSpanNameOverride = (name: string) => name;
```
ofrobots: AFAICT, there are no tests paths covering the `string` case.

kjin: Thanks for pointing that out. I added a test to check that a string config options gets turned into a function when the configuration is getting normalized.

kjin: TSC doesn't allow the field `rootSpanNameOverride` to be a string when accessed from a `Tracer` instance, so I am not covering that in this file.

PR #849, opened 2018-08-29
chore: pin @types/node at 10.7

### LINKED ISSUE #848: Fix TypeScript compilation errors
The TypeScript build is currently breaking on master  🤷‍♂️

kjin:
Fixes #848

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "@types/mocha": "5.2.1",
     "@types/ncp": "^2.0.1",
     "@types/nock": "^9.1.2",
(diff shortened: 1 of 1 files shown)


kjin:
Explanation: TypeScript doesn't allow auto type inference of argument `z` here:

```
let a: (x: string) => string;
let b: ((x: string) => string)|((x: number) => number);
let c: typeof a = y => y; // ok
let d typeof b = z => z; // z is implicitly `any`
```

`http.get` was recently changed to be a union type of two function signatures.

PR #858, opened 2018-09-08
feat: use small HTTP dependency

fhinkel:
Use teenyRequest instead of request. This saves 900 ms at
startup time.

Improve load time from 1500 ms to 650 ms. 

Before `request`:
screen shot 2018-09-07 at 9 27 55 pm

After with `teenyRequest`:
screen shot 2018-09-07 at 9 28 52 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "@types/pify": "^3.0.0",
     "@types/protobufjs": "^5.0.31",
     "@types/proxyquire": "^1.3.28",
-    "@types/request": "^2.0.8",
+    "@types/request": "^2.47.1",
     "@types/semver": "^5.4.0",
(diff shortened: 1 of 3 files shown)


kjin:
There is a problem where `teeny-request` expects its `body` input to be an object, while `request` expects it to be a string. We exhibit the latter behavior currently, so this PR fails system tests. I'm currently trying to fix it.

fhinkel:
Thanks, PRs to teeny are welcome! I had accidentally stringified a string, but fixed that in 3.8.

fhinkel:
ping @kjin

kjin:
@fhinkel Oh, I didn't know you were the author 😆 I'll open a PR if you haven't fixed that already.

PR #861, opened 2018-09-10
test: stop using TRACE_TEST_* for system tests

### LINKED ISSUE #859: System test with wrong permission keeps zombie Node process around
When running `npm run system-test` with insufficient permissions, a Node process keeps running in the background blocking `8080`: 

```sh
$ npm run system-test
[...] PERMISSION_DENIED: Missing or insufficient permissions.
$ npm run system-test
[...]Uncaught Error: listen EADDRINUSE :::8080
```

kjin:
Also, switches to `long-door-651` and fixes #859


.circleci/config.yml
```diff
@@ -54,6 +54,9 @@ unit_tests: &unit_tests
         command: npm run check-install
     - run:
         name: Run system tests (skipped for PRs)
+        environment:
+          GCLOUD_PROJECT: node-team-test
```
JustinBeckwith: Could we please switch this to `long-door-651`?

kjin: Sure. Could you give me permissions to the GCP project so I can check in a new key file as well?

JustinBeckwith: done!

kjin:
@JustinBeckwith I switched over to the new project, mind taking another look?

system-test/package.json
```diff
@@ -6,6 +6,7 @@
     "express": "^4.16.3",
     "google-auto-auth": "^0.10.0",
     "got": "^8.0.0",
+    "hard-rejection": "^1.0.0",
```
JustinBeckwith: If we have `--throw-deprecation` enabled, you shouldn't need this :)

kjin: We are running system tests with all of Node 6, 8, and 10, so that doesn't seem to handle all cases yet...

kjin: I'll land this now, let me know if I need to make a follow-up!

Issue #838, opened 2018-08-10
Potential Memory Leak


aftershockdan:
We recently had a memory leak in our application that was causing heap out of memory errors and causing processes to crash. After some investigation it seems like there might be a memory leak from caused by this library. Once we removed the trace-agent lib the memory profile returned to normal. The offending src seems to be the `endSpan` function in `./node_modules/@google-cloud/trace-agent/build/src/span-data.js`. I noticed a couple related issues, but that there was no clear resolution yet: https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/issues/632, https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/issues/603

Here is a screenshot of our heap profile:

![screen shot 2018-08-10 at 12 18 59 pm](https://user-images.githubusercontent.com/36009594/43981220-c7458da2-9ca5-11e8-8bf2-d8eb0d9ed8d1.png)

We are running a very simple express app with just a couple routes. The memory would very slowly build up overtime, over several days. If you need any other info please let me know. Hopefully we can track it down.

Versions:
@google-cloud/trace-agent: 3.0.0,
express: 4.16.3,
gke: 1.10.5-gke.3,
node: 8.11.3 alpine

ofrobots:
@aftershockdan thanks for opening the issue. We'll look into reproducing. Do you have info on how fast the leak is respective to load? Does it grow slowly ("over several days" as you mention) even under heavy load, or did you only try with light load? Can you also provide the version of Node you are using?

BTW, we also work with the folks who build Stackdriver Profiler support for Node.js. I think it is great that it helped you identify the suspected memory leak down to a module. If you have feedback on the usability of the profiler, we'd be quite eager for it.

aftershockdan:
Thanks @kjin !

@ofrobots I just pulled a chart of the memory use over 1 week for the same period as the heap profile above:

![screen shot 2018-08-10 at 2 42 29 pm](https://user-images.githubusercontent.com/36009594/43982920-ed5f11d2-9cac-11e8-8372-fd0a16c96d5e.png)

It looks like it is pretty consistently taking ~12 hrs from pod creation until heap overflows and crashes. We are using node v8.11.3, currently running 4 processes in cluster mode receiving ~400 reqs/second. Each process receives ~100 reqs/second to a single endpoint.

[...]

PR #870, opened 2018-09-25
feat: emit an error log on potential memory leak scenario

kjin:
Given that the scenario #838 is likely caused by context confusion on long-running requests, this warning should help us determine whether a memory leak is due to an intrinsic design detail to the agent, or context confusion. Emitting this warning is a clear sign of the latter (the remedy would be to identify the userspace queueing mechanism that breaks context, and patch it/request that it support async resources).

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -25,26 +25,6 @@ export type CLSMechanism =
 /** Available configuration options. */
 export interface Config {
   /**
-   * The trace context propagation mechanism to use. The following options are
-   * available:
-   * - 'async-hooks' uses an implementation of CLS on top of the Node core
-   *   `async_hooks` module in Node 8+. This option should not be used if the
-   *   Node binary version requirements are not met.
-   * - 'async-listener' uses an implementation of CLS on top of the
-   *   `continuation-local-storage` module.
-   * - 'auto' behaves like 'async-hooks' on Node 8+, and 'async-listener'
-   *   otherwise.
-   * - 'none' disables CLS completely.
-   * - 'singular' allows one root span to exist at a time. This option is meant
-   *   to be used internally by Google Cloud Functions, or in any other
-   *   environment where it is guaranteed that only one request is being served
-   *   at a time.
-   * The 'auto' mechanism is used by default if this configuration option is
-   * not explicitly set.
-   */
-  clsMechanism?: CLSMechanism;
-
-  /**
    * Log levels: 0=disabled, 1=error, 2=warn, 3=info, 4=debug
    * The value of GCLOUD_TRACE_LOGLEVEL takes precedence over this value.
    */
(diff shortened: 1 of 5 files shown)


src/trace-api.ts
```diff
@@ -241,6 +242,17 @@ export class StackdriverTracer implements Tracer {
             rootSpan.span.name}] was already closed.`);
         return UNCORRELATED_CHILD_SPAN;
       }
+      if (rootSpan.trace.spans.length === this.config!.spansPerTraceSoftLimit) {
```
soldair: is this the only code path where something is added to `rootSpan.trace.spans`?
just curious if the === will always hit. I didn't quite finish tracing the code path

kjin: Good catch -- you're right that this is not always the case, as `rootSpan` has an API to directly create child spans, which circumvents this code. However, no automatic tracing plugin uses the API (because no plugin creates both root and child spans from a single module in the same continuation). So unless the user is specifically using the custom span API, in which case they are circumventing the use of context propagation entirely, they will see this message if they rack up enough child spans.

I will add this as a comment.

src/trace-api.ts
```diff
@@ -241,6 +242,17 @@ export class StackdriverTracer implements Tracer {
...
+        // As in the previous case, a root span with a large number of child
+        // spans suggests a memory leak stemming from context confusion. This
+        // is likely due to userspace task queues or Promise implementations.
+        this.logger!.warn(`TraceApi#createChildSpan: [${
```
ofrobots: The problem I see with this is that warnings are not shown to the user by default because the default log level is 'error'.

I would propose making this an error (if you do this alongside my other suggestion of increasing the value by an order of magnitude).

kjin: I agree, if the limit is high enough we should just emit an error. However, I'm not sure if we should change behavior by setting a hard upper limit to # spans per trace.

src/config.ts
```diff
@@ -197,11 +205,12 @@ export interface Config {
  * user-provided value will be used to extend the default value.
  */
 export const defaultConfig = {
-  clsMechanism: 'auto' as CLSMechanism,
   logLevel: 1,
   enabled: true,
   enhancedDatabaseReporting: false,
   rootSpanNameOverride: (name: string) => name,
+  clsMechanism: 'auto' as CLSMechanism,
+  spansPerTraceSoftLimit: 25,
```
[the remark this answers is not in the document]
kjin: Ok, makes sense.

src/trace-api.ts
kjin: I changed the PR to add a hard limit. PTAL

src/trace-api.ts
kjin: BTW -- re: your approval comment, in the meantime, I also added tests.

src/config.ts
```diff
@@ -197,11 +212,13 @@ export interface Config {
...
  */
 export const defaultConfig = {
-  clsMechanism: 'auto' as CLSMechanism,
   logLevel: 1,
   enabled: true,
   enhancedDatabaseReporting: false,
   rootSpanNameOverride: (name: string) => name,
+  clsMechanism: 'auto' as CLSMechanism,
+  spansPerTraceSoftLimit: 200,
+  spansPerTraceHardLimit: 1000,
```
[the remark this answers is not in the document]
kjin: It doesn't make sense, but I don't think it's important enough to emit an error. I'll add a note in the docs to say that it should be higher.

src/trace-api.ts
```diff
@@ -242,6 +244,34 @@ export class StackdriverTracer implements Tracer {
...
+        // Note that since child spans can be created by users directly on a
+        // RootSpanData instance, this block might be skipped because it only
+        // checks equality -- this is OK because no automatic tracing plugin
+        // uses the RootSpanData API directly.
+        this.logger!.warn(`TraceApi#createChildSpan: [${
+            this.pluginName}] Adding child span [${
+            options.name}] will cause the trace with root span [${
+            rootSpan.span.name}] to contain more than ${
+            this.config!
+                .spansPerTraceSoftLimit} spans. This is likely a memory leak.`);
```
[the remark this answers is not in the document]
kjin: I changed it to be an error log. I'm not sure what's the best place to link to for context confusion -- maybe we should have a wiki page in this repo?

mikelueck:
Guys the promotion of these logs from "warn" to "error" is causing problems.  Specifically line 241 in trace-api.ts.
Is an error required here?

ofrobots:
@mikelueck this was a mistake. The issue is fixed in #882 and released as the latest version of the module.

PR #871, opened 2018-09-27
fix: update @google-cloud/common dependency to 0.25.3

kjin:
This PR upgrades @g-c/c to avoid an issue with authentication in `gcp-metadata` that was fixed in a newer version of the module.

In order to make this change, I've changed and moved tests that deal with keyFilename/credentials to be more modular (it's a test of the trace writer, not the module as a whole).

Also closes #855


JustinBeckwith:
@kjin *nudges*

PR #872, opened 2018-09-27
feat: support context propagation in bluebird

kjin:
This PR adds support for context propagation across bluebird Promises. (Edit: it also removed the knex plugin, which seems to be rendered superfluous by the bluebird plugin)


src/plugins/plugin-bluebird.ts
```diff
@@ -0,0 +1,30 @@
+import * as shimmer from 'shimmer';
```
ofrobots: add copyright

JustinBeckwith:
👋 @googleapis/node-team @kjin this is gettin a little moldy.  Are we moving forward with this PR?

kjin:
@googleapis/node-team This is ready for another look. Note that the knex plugin has now been removed, it seems to be superfluous.

test/plugins/test-cls-bluebird.ts
```diff
@@ -0,0 +1,178 @@
...
+  });
+
+  // Some (one) test cases deal with arrays rather than no deferred value.
+  // tslint:disable-next-line:no-any
+  const testCases: Array<TestCase<any>> = [
+    {
+      description: 'immediate resolve + child from then callback',
+      makePromise: () => new BPromise(res => res()),
+      thenFn: (p, cb) => p.then(cb)
+    } as TestCase,
```
ofrobots: why is this cast needed here and elsewhere?

kjin: It's a TypeScript limitation, there's no other way to make each individual `TestCase` object self-consistent (TS doesn't know that the return type of `makePromise` should be the same as the input for `thenFn`)

Issue #794, opened 2018-07-09
Prevent truncation of child spans that represent tail latencies


kjin:
It might be possible to stop the truncation child spans when a root span is ended. We would need to modify the trace queueing mechanism to send child spans by themselves (i.e. not accompanied by their local parents).

This would prevent misrepresentations of work stemming from early-truncated spans if certain RPCs have tail latencies.

cc/ @draffensperger @ofrobots

PR #876, opened 2018-09-29
refactor: improve internal TraceWriter API

kjin:
This PR contains three changes (in separate commits) that improve the TraceWriter API, narrowing down its trace-writing surface to _just_ `writeTraces` and promisifying the initialization step. This should make #794 easier to fix.

### DIFF
--- a/src/trace-writer.ts
+++ b/src/trace-writer.ts
@@ -235,7 +235,7 @@ export class TraceWriter extends common.Service {
    *
    * @param trace The trace to be queued.
    */
-  writeSpan(trace: Trace) {
+  writeTrace(trace: Trace) {
     for (const span of trace.spans) {
       if (span.endTime === '') {
         span.endTime = (new Date()).toISOString();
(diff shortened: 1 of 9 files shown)


src/trace-writer.ts
```diff
@@ -121,74 +121,65 @@ export class TraceWriter extends common.Service {
...
-    this.getProjectId().then(
-        () => {
-          this.scheduleFlush();
-          if (--pendingOperations === 0) {
-            cb();
-          }
-        },
-        (err: Error) => {
+    const gettingProjectId =
+        this.getProjectId().then(() => this.scheduleFlush(), (err: Error) => {
```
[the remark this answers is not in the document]
kjin: I need to make `gettingProjectId` an IIFE to preserve the behavior here. Let me know what you think.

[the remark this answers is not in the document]
kjin: Never mind, though I still have to wrap it in a separate function.

src/trace-writer.ts
```diff
@@ -121,102 +121,87 @@ export class TraceWriter extends common.Service {
...
+        this.logger.error(
+            'TraceWriter#initialize: Unable to acquire the project number',
+            'automatically from the GCP metadata service. Please provide a',
+            'valid project ID as environmental variable GCLOUD_PROJECT, or',
+            `as config.projectId passed to start. Original error: ${err}`);
+        throw err;
+      }
+      this.scheduleFlush();
+    };
+    const [hostname, instanceId] = await Promise.all([
```
ofrobots: `const [hostname, instanceId, ignore] = ...`

Correctness should be obvious. The missing third return value seems like an oversight at first glance.

src/trace-writer.ts
```diff
@@ -121,102 +121,87 @@ export class TraceWriter extends common.Service {
...
+      this.getHostname(), this.getInstanceId(), getProjectIdAndScheduleFlush()
+    ]);
+    // tslint:disable-next-line:no-any
+    const addDefaultLabel = (key: string, value: any) => {
```
ofrobots: Not part of your change, but given that you are here: Is `any` really needed here? I think `{}` would work?

kjin: It's a good point, it can either only be a string or a number (in the case of instance ID)

src/trace-writer.ts
kjin: `getProjectIdAndScheduleFlush` returns `void`, so IMO it would seem just as much an oversight to try to get the return value. Anyway, I added `_` as the third return value.

PR #885, opened 2018-10-19
fix: avoid memory leaks due to undisposed promise resources

kjin:
Most Promise async resources require a separate `promiseResolve` event listener. See the inline comments for more details.

### DIFF
--- a/src/cls/async-hooks.ts
+++ b/src/cls/async-hooks.ts
@@ -55,19 +55,21 @@ export class AsyncHooksCLS<Context extends {}> implements CLS<Context> {
     // Create the hook.
     this.hook = this.ah.createHook({
       init: (id: number, type: string, triggerId: number, resource: {}) => {
-        const currentId = this.ah.executionAsyncId();
         // init is called when a new AsyncResource is created. We want code
         // that runs within the scope of this new AsyncResource to see the same
         // context as its "parent" AsyncResource. The criteria for the parent
         // depends on the type of the AsyncResource. (If the parent doesn't have
         // an associated context, don't do anything.)
-        if (type === 'PROMISE' && this.contexts[currentId] !== undefined) {
+        if (type === 'PROMISE') {
           // Opt not to use the trigger ID for Promises, as this causes context
           // confusion in applications using async/await.
           // Instead, use the ID of the AsyncResource in whose scope we are
           // currently running.
-          this.contexts[id] = this.contexts[currentId];
-        } else if (this.contexts[triggerId] !== undefined) {
+          const currentId = this.ah.executionAsyncId();
+          if (this.contexts[currentId] !== undefined) {
+            this.contexts[id] = this.contexts[currentId];
+          }
+        } else {
           // Use the trigger ID for any other type. In Node core, this is
           // usually equal the ID of the AsyncResource in whose scope we are
           // currently running (the "current" AsyncResource), or that of one
@@ -77,7 +79,9 @@ export class AsyncHooksCLS<Context extends {}> implements CLS<Context> {
           // AsyncResource API, because users of that API can specify their own
           // trigger ID. In this case, we choose to respect the user's
           // selection.
-          this.contexts[id] = this.contexts[triggerId];
+          if (this.contexts[triggerId] !== undefined) {
+            this.contexts[id] = this.contexts[triggerId];
+          }
         }
         // Note that this function only assigns values in this.contexts to
         // values under other keys; it never generates new context values.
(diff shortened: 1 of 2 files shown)


src/cls/async-hooks.ts
```diff
@@ -65,7 +65,7 @@ export class AsyncHooksCLS<Context extends {}> implements CLS<Context> {
           // Instead, use the ID of the AsyncResource in whose scope we are
           // currently running.
           this.contexts[id] = this.contexts[this.ah.executionAsyncId()];
-        } else {
+        } else if (this.contexts[triggerId] !== undefined) {
```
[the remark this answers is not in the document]
kjin: I assume you mean comments after the `else if` block, since those are the comments that are now stale. I changed them.

kjin:
@ofrobots I have added tests. They fail when the source changes are not applied.

src/cls/async-hooks.ts
```diff
@@ -55,17 +55,18 @@ export class AsyncHooksCLS<Context extends {}> implements CLS<Context> {
...
       init: (id: number, type: string, triggerId: number, resource: {}) => {
+        const currentId = this.ah.executionAsyncId();
         // init is called when a new AsyncResource is created. We want code
         // that runs within the scope of this new AsyncResource to see the same
         // context as its "parent" AsyncResource. The criteria for the parent
         // depends on the type of the AsyncResource.
-        if (type === 'PROMISE') {
+        if (type === 'PROMISE' && this.contexts[currentId] !== undefined) {
           // Opt not to use the trigger ID for Promises, as this causes context
           // confusion in applications using async/await.
```
ofrobots: Here's my problem with this comment. 'Opt not to use the trigger ID for promise...' comment doesn't explain why the extra `&&` is there in the test.  Nor does the preceding comment explain why the test is the way it is.

I want a comment explaining the extra condition you added.

src/cls/async-hooks.ts
```diff
@@ -77,15 +78,29 @@ export class AsyncHooksCLS<Context extends {}> implements CLS<Context> {
...
+        // they query the current context when it is not stored in
+        // this.contexts.
       },
       destroy: (id: number) => {
         // destroy is called when the AsyncResource is no longer used, so also
         // delete its entry in the map.
         delete this.contexts[id];
+      },
+      promiseResolve: (id: number) => {
+        // Some Promise async resources do not get destroyed with a destroy
```
ofrobots: Is this accurate? The destroy hook will be called *eventually*, but that may be too far in the future.

The only reason why it may not get called (ever) is if the application is leaking promises.

kjin: As far as I can tell, the common case (esp. in newer versions of Node) is that the destroy hook is not called for Promises. I'm not sure exactly which cases a destroy hook will be called, but in my own tests it seemed to get seldom, if ever, called.

ofrobots: @bmeurer would this match your intuition? My understanding is that the Promise destroy hook should eventually get called unless the promise itself is leaking.

bmeurer: Hm, it should always be called. It might just take a long time for the major GC to kick in.

ofrobots: @kjin I would suggest experimenting with `--expose-gc` and manually calling `gloabl.gc()` (2-3 times) in strategic places to see if that changes things.

I don't think the comment is accurate as written. I still agree with the meat of this change though.

kjin: `global.gc()` definitely changes things. I agree that the comment is inaccurate now, as there must be some condition where the destroy hook will be called. That being said it seems like as far as I've observed, it still doesn't really ever happen without explicit invocation of the gc. But that is something that should change with new versions of Node.

src/cls/async-hooks.ts
```diff
@@ -55,17 +55,19 @@ export class AsyncHooksCLS<Context extends {}> implements CLS<Context> {
     // Create the hook.
     this.hook = this.ah.createHook({
       init: (id: number, type: string, triggerId: number, resource: {}) => {
+        const currentId = this.ah.executionAsyncId();
```
ofrobots:
```suggestion
        // init is called when a new AsyncResource is created. We want code
        // that runs within the scope of this new AsyncResource to see the same
        // context as its "parent" AsyncResource.
        const currentId = this.ah.executionAsyncId();
        // If the parent doesn't have an associated context, skip.
        if (this.context[currentId] !== undefined) {
          return;
        }
        // The criteria for parent depends on the type of AsyncResource.
        if (type === 'PROMISE') {
          // zig.
        } else {
          // zag.
        }
```

kjin: I realize that having an undefined current execution context is problematic, but the suggestion as-is doesn't allow trigger ID to be respected when current execution context is undefined, which I think might prevent legitimate use cases of the trigger ID to be ignored. If it's for the sake of cleaner code, PTAL at the change I made instead.

PR #895, opened 2018-10-25
refactor: address trace policy TODOs

### LINKED ISSUE #894: Sampling rate too high
The sampling rate is limited to minimum 1 per second, but this is still too high.

I would like to be able to specify for example 1 per 10 seconds, or even just a percentage like 0.1%.

kjin:
This change removes two TODOs from the codebase related the Trace Policy, namely:
- Turn `TracePolicy` into a class
- Privatize `StackdriverTracer.policy`

### DIFF
--- a/src/tracing-policy.ts
+++ b/src/tracing-policy.ts
@@ -14,15 +14,12 @@
  * limitations under the License.
  */
 
-/**
- * An object that determines whether a request should be traced.
- */
-export interface TracePolicy {
-  shouldTrace(dateMillis: number, url: string): boolean;
+interface TracePolicyPredicate<T> {
+  shouldTrace: (value: T) => boolean;
 }
 
-export class RateLimiterPolicy implements TracePolicy {
-  private traceWindow: number;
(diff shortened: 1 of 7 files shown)


src/tracing-policy.ts
```diff
@@ -42,51 +39,72 @@ export class RateLimiterPolicy implements TracePolicy {
...
+  ignoreUrls: Array<string|RegExp>;
 }
 
-export class TraceNonePolicy implements TracePolicy {
-  shouldTrace() {
-    return false;
+/**
+ * A class that makes decisions about whether a trace should be created.
+ */
+export class TracePolicy {
```
ofrobots: Should this implement `TracePolicyPredicate<RateAndUrl>`?

kjin: I opted not to do so just because there's no need to export `TracePolicyPredicate`. The only reason the interface exists is because I would run into issues typing `sampler` as a `Sampler`.

Thanks for looking at this!

PR #900, opened 2018-10-31
feat: add contextHeaderBehavior option

### LINKED ISSUE #879: [Feature] configuration option: only trace when x-cloud-trace-context is set
When working in cross-service tracing contexts, I would like to define policies and filters at the ingress service and not have to also set the same url filters on each of the microservices. It would be nice if I could set all but the public trace server to require the header to be set, and then I could rate limit and filter just on the public microservice.

Since the `x-cloud-trace-context` is used for tracing spans, would a filter based on the header being set or not be easy to add?

https://github.com/googleapis/cloud-trace-nodejs/blob/f8b74e6ec0af23adb531ff48aa18ad1d89703834/doc/trace-api.md#cross-service-trace-contexts

kjin:
Fixes #879, but also rolls the "experimental" `ignoreContextHeader` option into a new `contextHeaderOptions`.


src/config.ts
```diff
@@ -137,6 +142,22 @@ export interface Config {
    */
   samplingRate?: number;
 
+  /**
+   * Specifies how to use incoming trace context headers. The following options
+   * are available:
+   *   'default' -- Trace context headers will be respected for incoming
```
ofrobots: What does 'respected' mean here?

src/config.ts
```diff
@@ -137,6 +142,22 @@ export interface Config {
...
+   *   requests that support it. A new trace will be created for requests
+   *   without trace context headers.
+   *   'require' -- Same as default, but traces won't be created for requests
+   *   without trace context headers. This should not be set for end user-facing
```
ofrobots: Please also mention that the option should only be set if the headers are in the trustable. If an attacker can control the heade, they may cause excessive tracing.

src/config.ts
```diff
@@ -137,6 +142,22 @@ export interface Config {
...
+   *   services, as this header is usually set by other traced services rather
+   *   than by users.
+   *   'ignore' -- Always ignore trace context headers, so a new trace with a
+   *   unique ID will be created for every request.
```
ofrobots: The original comment for `ignoreContextHeader` did a much better job explaining when this is useful. As part of your transition you have lost that comment.

src/config.ts
kjin: `require` is more strict than `default`, so I don't think that applies.

src/config.ts
kjin: I've updated it to try to include the accurate portion of the docs (`Settings this to true... same incoming request.` is misleading).

src/config.ts
kjin: I've reworded this.

PR #901, opened 2018-11-01
chore: update the CircleCI config

JustinBeckwith:


ofrobots:
All the tests were broken. Why did this land?

alexander-fenster:
+1 to @ofrobots - since we have much more stable CI as before, I (again) suggest that we disable admin override on CI tasks for all repositories.

PR #905, opened 2018-11-05
doc: fix samplingRate comment in README

### LINKED ISSUE #904: Unclear description of the samplingRate configuration
There is a snippet in the README.md describing the usage of the `samplingRate`:

```js
require('@google-cloud/trace-agent').start({
  samplingRate: 500, // sample one trace every half-second.
  ignoreUrls: [ /^\/ignore-me#/ ] // ignore the "/ignore-me" endpoint.
});
// ...
```

However, it seems that the description `sample one trace every half-second.` is not correct.
And I think it should be `sample 500 traces every second.`

I found the `samplingRate` is used to create a `Sampler`:
https://github.com/googleapis/cloud-trace-nodejs/blob/master/src/tracing-policy.ts#L84

And the `Sampler` treat it as `samplesPerSecond`:
https://github.com/googleapis/cloud-trace-nodejs/blob/master/src/tracing-policy.ts#L25

kjin:
Fixes #904

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Optionally, you can pass a [configuration object](src/config.ts) to the `start()
 
 ```js
 require('@google-cloud/trace-agent').start({
-  samplingRate: 500, // sample one trace every half-second.
+  samplingRate: 500, // sample 500 traces per second, or at most 1 every 2 milliseconds.
   ignoreUrls: [ /^\/ignore-me#/ ] // ignore the "/ignore-me" endpoint.
 });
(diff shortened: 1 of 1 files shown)


kjin:
@soldair it's up to users, 500 is just for illustrative purposes. Thanks!

rueian:
Thank you! The previous comment cost me more than  USD 200 per day on Stackdriver Tracing Billing.

I think this fix will save a lot of people.

PR #913, opened 2018-11-10
feat: support child spans with tail latencies

### LINKED ISSUE #794: Prevent truncation of child spans that represent tail latencies

kjin:
Fixes #794

### DIFF
--- a/src/span-data.ts
+++ b/src/span-data.ts
@@ -104,6 +104,9 @@ export abstract class BaseSpanData implements Span {
   }
 
   endSpan(timestamp?: Date) {
+    if (!!this.span.endTime) {
+      return;
+    }
     timestamp = timestamp || new Date();
     this.span.endTime = timestamp.toISOString();
   }
@@ -114,6 +117,23 @@ export abstract class BaseSpanData implements Span {
  */
 export class RootSpanData extends BaseSpanData implements types.RootSpan {
   readonly type = SpanType.ROOT;
+  // Locally-tracked list of children.
+  private children: ChildSpanData[] = [];
+  // A function that replaces the ChildSpanData#endSpan prototype method on
+  // a child span instance.
+  private static overrideChildEndSpanHandler = function(
+      this: ChildSpanData, timestamp?: Date) {
+    if (!!this.span.endTime) {
+      return;
+    }
+    ChildSpanData.prototype.endSpan.call(this, timestamp);
+    // Also, publish just this span.
+    traceWriter.get().writeTrace({
+      projectId: this.trace.projectId,
+      traceId: this.trace.traceId,
+      spans: [this.span]
+    });
+  };
 
   constructor(
       trace: Trace, spanName: string, parentSpanId: string,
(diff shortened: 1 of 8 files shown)


JustinBeckwith:
👋 @googleapis/node-team @kjin can someone take a look at this one?

src/span-data.ts
```diff
@@ -114,6 +117,23 @@ export abstract class BaseSpanData implements Span {
  */
 export class RootSpanData extends BaseSpanData implements types.RootSpan {
   readonly type = SpanType.ROOT;
+  // Locally-tracked list of children.
+  private children: ChildSpanData[] = [];
+  // A function that replaces the ChildSpanData#endSpan prototype method on
+  // a child span instance.
+  private static overrideChildEndSpanHandler = function(
+      this: ChildSpanData, timestamp?: Date) {
```
ofrobots: Can you use an identifier other than `this`?

src/span-data.ts
```diff
@@ -125,16 +145,29 @@ export class RootSpanData extends BaseSpanData implements types.RootSpan {
...
+    if (!!this.span.endTime) {
+      return;
+    }
     super.endSpan(timestamp);
     traceWriter.get().writeTrace(this.trace);
+    this.children.forEach(child => {
+      if (!child.span.endTime) {
+        // Child hasn't ended yet.
+        // Override the endSpan function to make it re-publish only itself.
+        child.endSpan = RootSpanData.overrideChildEndSpanHandler;
```
ofrobots: Is there another way to do this instead of monkey patching? Could you have a test in the `endSpan` function? It is best to avoid monkey patching functions unless absolutely necessary. A reader will have to know that the `endSpan` method is part of the dynamic state rather than just the data.

src/span-data.ts
kjin: `this` is a type annotation only, it'll be erased on compilation.

src/span-data.ts
kjin: I've added a commit that changes the way this is being done, LMK what you think.

test/test-trace-writer.ts
```diff
@@ -63,18 +63,20 @@ function mockNoMetadata() {
   };
 }
 
-function createDummyTrace(): Trace {
+let traceIdHighWater = 0;
```
ofrobots: nit: complete the name to `traceIdHighWatermark`

src/span-data.ts
```diff
@@ -125,16 +129,29 @@ export class RootSpanData extends BaseSpanData implements types.RootSpan {
...
     super.endSpan(timestamp);
     traceWriter.get().writeTrace(this.trace);
+    this.children.forEach(child => {
+      if (!child.span.endTime) {
+        // Child hasn't ended yet.
+        // Inform it that it needs to self-publish.
+        child.shouldSelfPublish = true;
+      }
+    });
+    this.children = [];
```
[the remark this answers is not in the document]
kjin: It's really more just that the `this.children` served its one-time purpose, so its value no longer holds any real meaning. I don't think there is significant memory impact either way, though there is the miniscule added benefit that in unforeseen circumstances where the root span _is_ being retained in memory for whatever reason, that is has a slightly smaller footprint.

PR #920, opened 2018-11-15
add ignoreMethods option

b-flightly:
pull request for: https://github.com/googleapis/cloud-trace-nodejs/issues/919

### DIFF
--- a/src/plugins/plugin-hapi.ts
+++ b/src/plugins/plugin-hapi.ts
@@ -51,6 +51,7 @@ function instrument<T>(
   const options: PluginTypes.RootSpanOptions = {
     name: req.url ? (urlParse(req.url).pathname || '') : '',
     url: req.url,
+    method: req.method,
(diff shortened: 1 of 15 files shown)


b-flightly:
I signed it!

src/plugins/plugin-hapi.ts
```diff
@@ -51,6 +51,7 @@ function instrument<T>(
   const options: PluginTypes.RootSpanOptions = {
     name: req.url ? (urlParse(req.url).pathname || '') : '',
     url: req.url,
+    method: req.method,
```
kjin: Can you add this line to `plugin-connect` and `plugin-express` as well?

b-flightly: I see a few "traceApi.runInRootSpan" under test/ that dont have the "method:" key.  

For instance: test/test-default-ignore-ah-health.ts

I am not very familiar with typescript though and dont know if thats OK.  Let me know if I should add something like: "method: ''" as well.

PR #922, opened 2018-11-15
feat: add ignoreMethods option

### LINKED ISSUE #919: ignoreMethods configuration option
With CORS enabled our stackdriver trace UI is getting overwhelmed with OPTIONS requests.  And I think it might not be reporting important requests since I have a sampling rate < # requests / second.

Having an ignoreMethods configuration option similar to ignoreUrls method would solve this.

kjin:
Fixes #919

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -135,6 +135,12 @@ export interface Config {
   ignoreUrls?: Array<string|RegExp>;
 
   /**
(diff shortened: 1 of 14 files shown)


kjin:
Hmm, didn't realize this was already opened as a PR earlier. Will look at the other one for now

PR #927, opened 2018-11-29
Update region tags

fhinkel:
Update region tags to match our samples tracker.


kjin:
CI failure is known -- merging since this PR doesn't involve source changed.

PR #930, opened 2018-12-03
fix: end child spans correctly in pg

kjin:
Test failures in PR #913 (support child spans with tail latencies) exposes an issue with the instrumentation code for Postgres, where child spans were not being closed if a generic `Submittable` is used in lieu of a `Query`. This PR fixes that.

### DIFF
--- a/src/plugins/plugin-pg.ts
+++ b/src/plugins/plugin-pg.ts
@@ -138,8 +138,6 @@ class PostgresPatchUtility {
     return this.tracer.wrap((err: Error|null, res?: pg_7.QueryResult) => {
       this.maybePopulateLabelsFromOutputs(span, err, res);
       span.endSpan();
-      // TS: Type cast is safe as we know that callback is a
-      // Function.
       callback(err, res);
     });
   }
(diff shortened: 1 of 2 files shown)


src/plugins/plugin-pg.ts
```diff
@@ -66,46 +61,148 @@ function populateLabelsFromOutputs(
...
+    return pgQuery;
+  }
+
+  patchCallback(callback: Callback<pg_7.QueryResult>, span: Span):
+      Callback<pg_7.QueryResult> {
+    return this.tracer.wrap((err: Error|null, res?: pg_7.QueryResult) => {
+      this.maybePopulateLabelsFromOutputs(span, err, res);
+      span.endSpan();
+      // TS: Type cast is safe as we know that callback is a
+      // Function.
```
ofrobots: which type cast?

test/plugins/test-trace-pg.ts
```diff
@@ -23,6 +23,11 @@ var assert = require('assert');
 
 var pgVersions = ['6', '7'];
 
+// process.on('unhandledRejection', (err) => {
+//   console.error(err);
+//   process.exit(1);
+// });
```
ofrobots: Remove this?

PR #933, opened 2018-12-05
chore: update codecov and nyc config

JustinBeckwith:


kjin:
This looks like an amendment of a previous commit from #929

PR #938, opened 2018-12-12
test: cleanup grpc and http2 tests to minimize flakes

kjin:
`grpc` and `http2` tests fail occasionally on nightly builds because the acceptable range for span time ranges isn't ideal for the CI environment. I've moved them off the old `common.assertSpanDuration` to the newer `utils.assertSpanDuration` which allows us to use more reasonable time ranges. There is also a bit of minor cleanup involved as some `common.*` values are no longer used anywhere.

Note: The `test-trace-http2.ts` diff seems large because of an indentation change, but the tests themselves are the same.


test/utils.ts
```diff
@@ -47,13 +53,17 @@ export function wait(ms: number) {
...
 // Assert that the given span's duration is within the given range.
-export function assertSpanDuration(span: TraceSpan, bounds: [number, number]) {
+export function assertSpanDuration(span: TraceSpan, bounds: [number, number?]) {
   const spanDuration = Date.parse(span.endTime) - Date.parse(span.startTime);
+  const lowerBound = bounds[0];
+  const upperBound = bounds[1] !== undefined ? bounds[1] : bounds[0];
   assert.ok(
-      spanDuration >= bounds[0] && spanDuration <= bounds[1],
+      spanDuration >= lowerBound - ASSERT_SPAN_TIME_TOLERANCE_MS &&
+          spanDuration <= upperBound + ASSERT_SPAN_TIME_TOLERANCE_MS,
```
ofrobots: Do we need leniency on both sides of the test here? I would expect that the upper bound would be rigid.

kjin: I added it for simplicity's sake. I don't see any opportunity for false positives as long as we keep the interval small.

Issue #942, opened 2018-12-15
req.connection.remoteAddress doesn't return real client ip behind proxies


sh:
We're using trace-agent for connect + express in pods behind a loadbalancer, the value of `/http/source/ip` is always the ip of the loadbalancer instead of the real client.

This makes debugging individual issues based on the `/http/source/ip` currently impossible.

Using `req.ip` instead of `req.connection.remoteAddress` for express and koa would respect any `x-forwarded-for` headers and report the real client ip.

API
- https://expressjs.com/en/4x/api.html#req.ip
- https://koajs.com/#request-ip

Docs
- https://expressjs.com/en/4x/api.html#trust.proxy.options.table
- https://koajs.com/#settings

I'd happily contribute a pull request for this :)

sh:
Done, thanks for taking the time to review this :)

kjin:
@sh no problem, thanks for your contribution! Published in 3.5.0 -- let us know if there are any problems!

PR #944, opened 2018-12-17
fix: #942 - use req.ip in express and koa plugin

sh:
fix for #942 :)

### DIFF
--- a/src/plugins/plugin-koa.ts
+++ b/src/plugins/plugin-koa.ts
@@ -86,7 +86,7 @@ function startSpanForRequest<T>(
     // middlewear runs.
     root.addLabel(api.labels.HTTP_METHOD_LABEL_KEY, req.method);
     root.addLabel(api.labels.HTTP_URL_LABEL_KEY, url);
-    root.addLabel(api.labels.HTTP_SOURCE_IP, req.connection.remoteAddress);
+    root.addLabel(api.labels.HTTP_SOURCE_IP, req.ip);
 
     // wrap end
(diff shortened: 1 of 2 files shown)


sh:
I signed it!

src/plugins/plugin-koa.ts
```diff
@@ -86,7 +86,7 @@ function startSpanForRequest<T>(
     // middlewear runs.
     root.addLabel(api.labels.HTTP_METHOD_LABEL_KEY, req.method);
     root.addLabel(api.labels.HTTP_URL_LABEL_KEY, url);
-    root.addLabel(api.labels.HTTP_SOURCE_IP, req.connection.remoteAddress);
+    root.addLabel(api.labels.HTTP_SOURCE_IP, req.ip);
```
kjin: It seems like `req` (= `ctx.req`) here is a raw `IncomingMessage`, so TS doesn't recognize `req.ip` as a field. Maybe this is what we want instead?

```suggestion
    root.addLabel(api.labels.HTTP_SOURCE_IP, ctx.request.ip);
```

sh: Yup, sorry I've made the changes without vscode 🤐

kjin: No problem! Thanks for the contribution!

sh: Cheers!

SifouElYasLakhHa:
I signed it!

PR #950, opened 2019-01-10
feat: allow users to specify a propagation mechanism

### LINKED ISSUE #949: Feature request: Extract and propagate B3 headers
Istio supports distributed tracing on an ingress into a cluster, and between services.

To do this, [it propagates a standard set of B3 headers](https://istio.io/docs/tasks/telemetry/distributed-tracing) (originally defined for Zipkin, I think?) 

```
x-request-id
x-b3-traceid
x-b3-spanid
x-b3-parentspanid
x-b3-sampled
x-b3-flags
x-ot-span-context
```

It would be great if the server side portions of this tracer recognised and propagated the information from that source. [opencensus-node](https://github.com/census-instrumentation/opencensus-node) apparently supports this, though it is itself in very early alpha, and has no GRPC support.

I can share simple istio config that defines this behaviour, if it's helpful.

kjin:
Fixes #949

This feature allows users to supply a `Propagation` (or an absolute path to a file that exports that interface) to specify how trace context headers are injected and extracted from HTTP requests. It also replaces the built-in propagation mechanism with that of `@opencensus/propagation-stackdriver`, which has the same behavior.

Some notes:
* Some OpenCensus propagation libraries either don't export a consistent interface or have implementation bugs, so they won't work out-of-the-box as of yet.
* The propagation object exposed through the `StackdriverTracer` object doesn't have the same interface as the input object, the latter of which is modeled on OpenCensus while the former is just something I deemed to be more suitable for this module. Particularly of note is that OpenCensus treats all span IDs as length-16 hex strings, while we use decimal integer strings.
* It may be better _not_ to accept an OpenCensus `Propagation` object because of the above reasons. I'm open to discussion on this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
   },
   "dependencies": {
     "@google-cloud/common": "^0.32.1",
+    "@opencensus/propagation-stackdriver": "0.0.11",
     "builtin-modules": "^3.0.0",
     "console-log-level": "^1.4.0",
     "continuation-local-storage": "^3.2.1",
--- a/src/config.ts
+++ b/src/config.ts
@@ -102,12 +102,21 @@ export interface Config {
    * A list of trace plugins to load. Each field's key in this object is the
    * name of the module to trace, and its value is the require-friendly path
    * to the plugin. (See the default configuration below for examples.)
-   * Any user-provided value will be used to extend its default value.
-   * To disable a plugin in this list, you may override its path with a falsy
-   * value. Disabling any of the default plugins may cause unwanted behavior,
-   * so use caution.
+   * Any user-provided value for this field will be used to extend its default
+   * value. To disable a plugin in this list, you may override its path with a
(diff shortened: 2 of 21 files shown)


kjin:
Blocked by: https://github.com/census-instrumentation/opencensus-node/issues/340

ofrobots:
@kjin census-instrumentation/opencensus-node#340 is done. Is this still blocked?

kjin:
It'll be unblocked when 0.0.10 is released. In other words, now blocked by [Milestone #2](https://github.com/census-instrumentation/opencensus-node/milestone/2)

emillg:
@kjin 0.0.10 has been released. Is this still blocked?

kjin:
@emillg Thanks for the reminder. I'll move this forward shortly

kjin:
I'm going to take a different approach on this... will leave the issue open so it won't get dropped.

PR #954, opened 2019-01-16
fix: wrap gRPC server async handlers

### LINKED ISSUE #952: Issues when enabling tracing on gRPC service
We were using the latest version of this module but after enabling it we couldn't see any traces whatsoever. After digging we found many of logs with
```
 severity:  "ERROR"  
 textPayload:  "original object and wrapper must be functions"
```
We believe it's because of the type checking one of the dependencies is doing ([see PR](https://github.com/othiym23/shimmer/pull/14)). We found also that switching to versions older to v3.5.0 seems to solve the problem.

Any idea on how to solve the issue for the latest version?

kjin:
Fixes #952

`plugin-grpc.ts` is best viewed with `?w=1`.

### DIFF
--- a/src/plugins/plugin-grpc.ts
+++ b/src/plugins/plugin-grpc.ts
@@ -345,47 +345,50 @@ function patchServer(server: ServerModule, api: Tracer) {
     // handlerSet.func is the gRPC method implementation itself.
     // We wrap it so that a span is started immediately beforehand, and ended
     // when the callback provided to it as an argument is invoked.
-    shimmer.wrap(handlerSet, 'func', (serverMethod) => {
-      return function serverMethodTrace(
-          this: Server, call: ServerUnaryCall<S>,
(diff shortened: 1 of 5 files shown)


src/plugins/plugin-grpc.ts
```diff
@@ -345,47 +345,49 @@ function patchServer(server: ServerModule, api: Tracer) {
...
-                  }
-                }
-                rootSpan.endSpan();
-                return callback(err, result, trailer, flags);
-              };
-          return serverMethod.call(this, call, wrappedCb);
-        });
+    // TODO(kjin): shimmer cannot wrap AsyncFunction objects.
+    // Once shimmer introduces this functionality, change this code to use it
+    // here, and in other server wrap* methods.
```
ofrobots: Maybe include link to https://github.com/othiym23/shimmer/pull/14

PR #968, opened 2019-02-12
test: test hapi tails

### LINKED ISSUE #382: Support `tails` in Hapi
As described by @AdriVanHoudt on PR [PR 379](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/379):

"In hapi there is a concept of `tails` where you can still do stuff in your handler after you replied.
In this case you end the span when you call `reply`. I have some routes where I do some queries after replying for performance reasons and it would be great if those queries would also end up in the trace of that request.
Would it be possible to integrate this? You have access to the full request object on that event and it is fired only once per request.
See https://github.com/hapijs/hapi/blob/master/API.md#server-events and https://github.com/hapijs/hapi/blob/master/API.md#requesttailname for more info.
And https://pbs.twimg.com/media/CL-bp3_W8AAimRi.png:large for a nice overview of how the hapi lifecycle works"

kjin:
Fixes #382

Since child spans are now allowed to track tail latencies, this PR adds a test to ensure that users of the Hapi tails API (up to v16, but also including the more general recommended mechanism in the [v17 release notes](https://github.com/hapijs/hapi/issues/3658) under "Misc") can observe these child spans.

Note that child spans still cannot begin _after_ a root span has ended.


ofrobots:
> Note that child spans still cannot begin _after_ a root span has ended.

Does this still fix #382 ?

kjin:
@ofrobots The Hapi API for request tails doesn't give us enough control to define a new root span under which tail work will be associated. It's also a removed API in v17+ so I would consider it not high priority to support additional instrumentation specific to Hapi tails beyond just this test.

PR #970, opened 2019-02-13
test: don't make network request in gRPC context test

kjin:
We should stop making a request to `http://www.google.com/` in our unit test.


kjin:
@DominicKramer It hasn't been converted to idiomatic TypeScript. This has been an incremental (and rather stalled) process over the last two years.

PR #981, opened 2019-02-27
doc: add warning for knex database.raw

### LINKED ISSUE #946: mysql2 requests aren't being attached to root scope
The log is outputting `INFO TraceApi#createChildSpan: [mysql2@1.6.4:lib/connection.js] Creating phantom child span [mysql-query] because there is no root span.`.

My setup is using knex with the client `mysql2`. The query is running in the knex `raw` method.

This is an example app: https://github.com/davidharrisonlb/mysql-tracer-test

kjin:
Fixes #946


README.md
```diff
@@ -154,6 +154,8 @@ These mechanisms work great in most cases. However, they do have some limitation
...
 * If your application uses untranspiled `async` functions, you must use Node 8+. (Untranspiled `async` functions are supported from Node 7.6 onward, but we do not support tracing these functions in Node 7.)
+* Requests originating from the body of a `then` implementation in a userspace thenable may not get traced if the thenable is `await`-ed. This is largely an unconventional case but is present in the `knex` module, which monkeypatches the Bluebird Promise's prototype to make database calls. __If you are using `knex` (esp. the `raw` function), see [https://github.com/googleapis/cloud-trace-nodejs/issues/946](#946) for more details on whether you are affected, as well as a suggested workaround.__
```
ofrobots: `userspace thenable` may not be well defined enough to most casual readers. How about something like this:

* Node.js `async_hooks` API has issues tracking context around `await`-ed "thenables" (rather than real promises). Requests originating from the body of a `then` implementation in such a user-space "thenable" may not get traced. This is largely an unconventional case but is present in the `knex` module, which monkeypatches the Bluebird Promise's prototype to make database calls. __If you are using `knex` (esp. the `raw` function), see [https://github.com/googleapis/cloud-trace-nodejs/issues/946](#946) for more details on whether you are affected, as well as a suggested workaround.__

kjin: Thanks, I changed it

PR #989, opened 2019-03-26
fix: don't let trace context injection throw

kjin:
It's possible for `setHeader` to throw if an outgoing request was created with a header `Expect` but not capitalized exactly that way. This change makes checking for the `expect` header a little more robust, and also swallows an error that would be thrown for further unexpected capitalizations.

It's possible that `hasExpectHeader` can either (1) stay the same or (2) be even more robust and actually iterate through all header keys to do a case-insensitive check. I'm definitely OK with the first option, the second option might have too much overhead to accomadate for a rare issue.

### DIFF
--- a/src/plugins/plugin-http.ts
+++ b/src/plugins/plugin-http.ts
@@ -44,12 +44,16 @@ function getSpanName(options: ClientRequestArgs|url.URL) {
 
 /**
  * Returns whether the Expect header is on the given options object.
+ * Assumes only that the header key is either capitalized, lowercase, or
+ * all-caps for simplicity purposes.
  * @param options Options for http.request.
  */
 function hasExpectHeader(options: ClientRequestArgs|url.URL): boolean {
   return !!(
       (options as ClientRequestArgs).headers &&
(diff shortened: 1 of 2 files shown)


src/plugins/plugin-http.ts
```diff
@@ -179,8 +183,14 @@ function makeRequestTrace(
     // Inject the trace context header, but only if it wasn't already injected
     // earlier.
     if (!traceHeaderPreinjected) {
-      req.setHeader(
-          api.constants.TRACE_CONTEXT_HEADER_NAME, span.getTraceContext());
+      try {
+        req.setHeader(
+            api.constants.TRACE_CONTEXT_HEADER_NAME, span.getTraceContext());
+      } catch (e) {
```
ofrobots: Can you make this error a bit more specific rather than swallowing all errors? For example, swallow the error only if `e.code` matches your expectation.

kjin:
@ofrobots Thanks for the suggestions, this PR is ready for another look.

PR #1002, opened 2019-04-25
chore: add node 12 ci job

kjin:
This change adds Node 12 as a Circle CI job, and de-facto adds support for Node 12. The following changes need to be made in the process:

* Because `hiredis` doesn't seem to build on Node 12 I've also opted to not test it on Node 12. I recommend viewing [files changed with `?w=1`](https://github.com/googleapis/cloud-trace-nodejs/pull/1002/files?w=1) to see a more illustrative diff. (Also see [`describeInterop` implementation notes](https://github.com/googleapis/cloud-trace-nodejs/blob/v3.6.1/test/utils.ts#L186-L200))
* References to Promises no longer exist on `PROMISE`-type async resources, so I've amended a test to count an expected number of async hook invocations rather than compare exact values retrieved from those async hook invocations.
* Node 12 appears to have a bug where async hook objects can't be re-enabled after being disabled.  See #1017. In the meantime, before this bug gets fixed, I've changed our implementation to create a dummy async hook that is always enabled. (This allows some tests to pass, but there will be a performance penalty for customers that run the disabled trace agent, e.g. if they start the agent but with invalid credentials).
* Node 12 error messages have changed (at least in HTTP/2). I've amended the tests to account for this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,6 +90,7 @@ workflows:
       - node8
       - node10
       - node11
+      - node12
   tests:
     jobs:
       - node8:
@@ -98,11 +99,14 @@ workflows:
           filters: *release_tags
       - node11:
           filters: *release_tags
+      - node12:
+          filters: *release_tags
       - publish_npm:
           requires:
             - node8
             - node10
             - node11
+            - node12
           filters:
             branches:
               ignore: /.*/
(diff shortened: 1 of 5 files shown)


JustinBeckwith:
@kjin it looks like the container exists, but there are a lot of deprecation warnings.  Mind taking a gander?

kjin:
@JustinBeckwith Yes, it requires a slight refactor of one of our tests. I'll get around to it momentarily.

kjin:
@jkwlui Thanks for taking a look. @ofrobots and I decided offline that we would wait until the bug (bullet 3) is fixed in Node before landing this PR.

kjin:
The above mentioned issue was fixed and released in Node 12.3.

PR #1007, opened 2019-04-29
fix: support tracing awaited mongoose queries

### LINKED ISSUE #1000: Awaited Mongoose queries are not traced
Due to this [issue](https://github.com/nodejs/node/issues/26064), mongoose queries do not get traced when they are awaited because the execution of query occurs in the body of a custom thenable.

kjin:
Fixes #1000

The problem occurs because `Query#exec` (which makes the MongoDB API call) occurs within the body of a userspace `then` call, which has no context when `await`ed. To remedy this, we patch new `Query` objects so that `exec` has the same context as where the `Query` was created. This assumes that we want context at the API call to be the same as the place where `Query` was constructed. As much (if not all) of the Mongoose read APIs create and execute the `Query` objects in a single function call, I believe this to be a safe assumption.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "@types/knex": "^0.15.1",
     "@types/methods": "^1.1.0",
     "@types/mocha": "^5.2.5",
+    "@types/mongoose": "^5.3.26",
     "@types/ncp": "^2.0.1",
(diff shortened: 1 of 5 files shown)


kjin:
TBR=@ofrobots

PR #1008, opened 2019-04-29
update to .nycrc with --all enabled

bcoe:
to work properly, nyc should also be updated to 14


kjin:
@JustinBeckwith @bcoe Would be nice to add conventional-commits prefix next time.

bcoe:
whoops, my bad; I should have thought of that.

PR #1009, opened 2019-04-30
remove the hash from the ignoreUrls in docs

sdowding:
Hello, the docs for ignoreUrls have a mysterious hash that me and my team think may be a mistake in the docs. I have updated it to prevent others from having the same discussion.

... or maybe is has some mysterious power we can't figure out?


sdowding:
I signed it!

PR #1011, opened 2019-05-02
build!: upgrade engines field to >=8.10.0

bcoe:
This drops support for all Node.js versions < 8.10.0, please upgrade

BREAKING CHANGE: library now requires Node >= 8.10.0


kjin:
This should wait until I have made another release on the 3.x branch (pending)

PR #1013, opened 2019-05-02
chore: fix code coverage, drop ncp

kjin:
* `ncp` causes issues with running scripts locally. It hasn't been updated in 4 years, so I'm dropping it favor of ~~just `cp -r`~~ `cpy`.
* We were writing coverage info to `coverage` but `codecov` appears to expect these files to exist in `.coverage`.
* I disabled `--all` as it slows down CI a ton, enough to cause flakes.
* I added CI Codecov flags. Not 100% sure if it will work until we can view diffs, but this would allow different coverage reports for different Node versions.


ofrobots:
`cp -r` wouldn't work on windows.

kjin:
@ofrobots ah, thanks for the reminder.

ofrobots:
For posterity, what are the problems that `ncp` is causing locally?

kjin:
`ncp` has an issue where sometimes a callback isn't called. Since I promisify it that means a Promise won't resolve, and my build script never continues -- it just exits with error code 0. https://github.com/AvianFlu/ncp/issues/131 is an example, their issue list seems to suggest several such problems.

The repo and module seem unmaintained, so I don't have hope that this will be fixed.

PR #1015, opened 2019-05-03
feat: use souce-map-support wrapCallSite to apply source maps to call stacks

AaronFriel:
This produces nice call stacks that display line and column numbers from transpiled TypeScript, Coffee Script, etc source files like this:

```
Callstack
runInSpan (/app/src/trace-util.ts:129:30)
runDispatcherInTrace [as runDispatcherInTrace] (/app/src/trace-util.ts:167:18)
source.asCallback (/app/src/dispatcher/simpleDispatcher.ts:57:29)
dispatchMessage (/app/src/source/natsSource.ts:141:9)
trace_util_1.runInSpan (/app/src/source/natsSource.ts:74:19)
that.runWithContext (/app/node_modules/@google-cloud/trace-agent/build/src/cls/async-hooks.js:143:49)
runWithContext [as runWithContext] (/app/node_modules/@google-cloud/trace-agent/build/src/cls/async-hooks.js:121:20)
contextWrapper (/app/node_modules/@google-cloud/trace-agent/build/src/cls/async-hooks.js:143:25)
tracing_1.tracer.runInRootSpan (/app/src/trace-util.ts:144:17)
cls_1.cls.get.runWithContext (/app/node_modules/@google-cloud/trace-agent/build/src/trace-api.js:193:20)
```

PR #1016, opened 2019-05-03
test: test source map support for stack traces

kjin:
Tests for #1015


kjin:
@soldair thanks for the heads up, our tests run with it by default. I have verified that it still works without the `-r` though.

Issue #994, opened 2019-04-06
Tracer periodically forgets configuration and traces every request (badly)


AaronFriel:
Periodically tracing stops working in a weird way that results in *every message* being traced, disregarding the sampling rate *and* the top level root span.

# What I expect to happen

Tracing is configured as so:

```ts
const TRACE_HEADER_BEHAVIOR = process.env.TRACE_HEADER_BEHAVIOR;
let contextHeaderBehavior: 'default' | 'require' = 'default';
let samplingRate: number = 1/60;
if (TRACE_HEADER_BEHAVIOR === 'require') {
  contextHeaderBehavior = TRACE_HEADER_BEHAVIOR;
  samplingRate = 0;
}

export const tracer = traceAgent.start({
  contextHeaderBehavior,
  enabled: process.env.TRACING === 'enabled',
  keyFilename: process.env.GOOGLE_APPLICATION_CREDENTIALS,
  logLevel: 0,
  projectId: process.env.TRACING_PROJECT,
  samplingRate,
});
```

We should expect that we will see one trace per minute, and we do when it is functioning normally. A service is configured to send a heartbeat once per second to a queue service, it runs in a root span called `heartbeat`.

```ts
export async function runHeartbeat(
  parentLogger: Logger,
  dispatcherUri: string,
  rateHz: number,
) {
  const chance = new Chance();

  const durationMs = 1000 / rateHz;

  return pForever(
    async i =>
      tracer.runInRootSpan({ name: 'heartbeat' }, async rootSpan => {
        const logger = traceLogger(parentLogger);

        const message: Payload = {
          payload: JSON.stringify({
            hops: 0,
            i,
            origin: `heartbeat-${os.hostname}-${i}`,
            [chance.string()]: chance.string(),
          }),
        };

        const uri = `${dispatcherUri}`;

        try {
          await postHeartbeat(logger, uri, message);
        } catch (err) {
          logger.error(
            { message, err, dispatcherUri },
            'Error posting message to Dispatcher',
          );
        } finally {
          rootSpan.endSpan();
        }

        await sleep(durationMs);

        return i + 1;
      }),
    1,
  );
}

async function postHeartbeat(
  parentLogger: Logger,
  uri: string,
  message: Payload,
) {
  const logger = traceLogger(parentLogger);
  logger.trace({ message }, 'Posting heartbeat');

  const response = await got.post(uri, {
    body: JSON.stringify(message),
    headers: {
[...]

PR #1021, opened 2019-05-08
refactor: explicitly assign every internal config field

kjin:
Prerequisite to addressing a feature request in #994... this change makes every internal config option assignment explicit, and re-organizes internal config options so that instead of `interface ParentConfig extends ChildConfig` it becomes `interface ParentConfig { childConfig: ChildConfig }`. This means that each configurable component (internally) only gets config options that it asks for, rather than a large object containing _all_ config options.

Tested behavior should remain the same; some tests have been changed because of the re-organization of internal options.


src/index.ts
```diff
@@ -35,73 +36,132 @@ let traceAgent: StackdriverTracer;
...
+  return {
+    [FORCE_NEW]: forceNew,
+    enabled: projectConfig.enabled,
+    logLevel: lastOf(
+        projectConfig.logLevel, Number(process.env.GCLOUD_TRACE_LOGLEVEL)),
+    clsConfig: {
+      [FORCE_NEW]: forceNew,
+      mechanism: transform(
+          projectConfig.clsMechanism,
+          (clsMechanism):
```
ofrobots: Too much details inline. Can you extract this out of line in a named function.

src/index.ts
```diff
@@ -35,73 +36,132 @@ let traceAgent: StackdriverTracer;
...
-
-  // If the CLS mechanism is set to auto-determined, decide now what it should
-  // be.
-  const ahAvailable = semver.satisfies(process.version, '>=8');
-  if (config.clsMechanism === 'auto') {
-    config.clsMechanism = ahAvailable ? 'async-hooks' : 'async-listener';
-  }
-
-  return config;
+  const projectConfig: (typeof defaultConfig)&Forceable<Config> =
```
ofrobots: `mergedConfig` would be a better name.

src/index.ts
```diff
@@ -35,73 +36,132 @@ let traceAgent: StackdriverTracer;
...
+            projectConfig.serviceContext.version,
+            process.env.GAE_MODULE_VERSION, process.env.GAE_VERSION),
+        minorVersion: lastOf<string|undefined>(
+            projectConfig.serviceContext.minorVersion,
+            process.env.GAE_MINOR_VERSION)
+      }
+    },
+    pluginLoaderConfig: {
+      [FORCE_NEW]: forceNew,
+      plugins: Object.assign({}, projectConfig.plugins),
```
ofrobots: style nit: `plugins: {... projectConfig.plugins)`

src/index.ts
kjin: Done

src/index.ts
kjin: Done

src/index.ts
kjin: IIRC this is not available in Node 6... which we haven't officially ripped out of CI yet.

ofrobots: it is available in TypeScript tho

kjin: We have to target es2015 to preserve generators, so it's not available even in TypeScript, unfortunately.

kjin: Ok, that being said I'll probably just drop node6 CI in a different PR.

kjin: Done

src/index.ts
```diff
@@ -35,73 +36,128 @@ let traceAgent: StackdriverTracer;
...
-      minorVersion: process.env.GAE_MINOR_VERSION || undefined
-    }
-  };
-
-  let envSetConfig: Config = {};
+function initConfig(userConfig: Forceable<Config>): Forceable<TopLevelConfig> {
+  // Helper function which gets the last non-null/undefined value among its
+  // arguments. The first argument must therefore be guaranteed to be
+  // non-null/undefined.
+  const lastOf =
```
ofrobots: this should be unit-tested.

src/index.ts
```diff
@@ -35,73 +36,128 @@ let traceAgent: StackdriverTracer;
...
+      <T>(defaultValue: T, ...otherValues: Array<T|null|undefined>): T => {
+        for (let i = otherValues.length - 1; i >= 0; i--) {
+          // tslint:disable:no-any
+          if (otherValues[i] != null &&
```
ofrobots: Is the `!=` intentional or did you want a `!==`?

src/index.ts
```diff
@@ -35,73 +36,128 @@ let traceAgent: StackdriverTracer;
...
+    const ahAvailable = semver.satisfies(process.version, '>=8');
+    if (clsMechanism === 'auto') {
+      return ahAvailable ? TraceCLSMechanism.ASYNC_HOOKS :
+                           TraceCLSMechanism.ASYNC_LISTENER;
+    }
+    return clsMechanism as TraceCLSMechanism;
+  };
+  const getInternalMaximumLabelValueSize = (maximumLabelValueSize: number) => {
+    // Enforce the upper limit for the label value size.
+    if (maximumLabelValueSize > Constants.TRACE_SERVICE_LABEL_VALUE_LIMIT) {
```
ofrobots: nit: Would this work:

```ts
return Math.min(Constants.TRACE_SERVICE_LABEL_VALUE_LIMIT, maximumLabelValueSize);
```

src/index.ts
```diff
@@ -35,73 +36,128 @@ let traceAgent: StackdriverTracer;
...
+      return Constants.TRACE_SERVICE_LABEL_VALUE_LIMIT;
+    }
+    return maximumLabelValueSize;
+  };
+  const getInternalRootSpanNameOverride =
+      (rootSpanNameOverride: string|((name: string) => string)) => {
+        // Make rootSpanNameOverride a function if not already.
```
ofrobots: Style nit: is this more readable:
```ts
switch (typeof rootSpanNameOverride) {
  case 'string': return () => rootSpanNameOverride;
  case 'function': return (name: string) => name;
  default: return rootSpanNameOverride;
}
```

src/index.ts
kjin: It was intentional. That being said, I've separated it out into `!== undefined` and `!== null` for readability.

src/index.ts
kjin: Yeah that's better, thanks!

src/index.ts
kjin: Done, I've moved it into `src/util.ts`.

src/index.ts
```diff
@@ -35,73 +36,109 @@ let traceAgent: StackdriverTracer;
...
+    // If the CLS mechanism is set to auto-determined, decide now
+    // what it should be.
+    const ahAvailable = semver.satisfies(process.version, '>=8');
+    if (clsMechanism === 'auto') {
+      return ahAvailable ? TraceCLSMechanism.ASYNC_HOOKS :
+                           TraceCLSMechanism.ASYNC_LISTENER;
+    }
+    return clsMechanism as TraceCLSMechanism;
+  };
+  const getInternalMaximumLabelValueSize = (maximumLabelValueSize: number) =>
```
ofrobots: nit: now that this is a single line function, it would be okay to move it inline. I'll leave it up to you though.

PR #1027, opened 2019-05-13
 feat!: allow users to specify a trace policy impl

### LINKED ISSUE #994: Tracer periodically forgets configuration and traces every request (badly)

kjin:
Fixes #994

This PR allows users to override the built-in `TracePolicy` object with their own implementation. Such an implementation has the following interface:

```ts
interface TracePolicy {
  shouldTrace: (
    requestDetails: {
      timestamp: number;
      url: string;
      method: string;
      traceContext: {traceId: string; spanId: string; options: number;}|null;
      options: {};
    }
  ) => boolean;
}
```

It also makes a breaking change to the config option:
* `contextHeaderBehavior` now only affects sampling.
* `propagateTraceContextFromHeader` has been added to allow users to ignore trace context when creating new trace and span IDs.
* `ignoreContextHeader` has been removed. It was already deprecated. (`propagateTraceContextFromHeader`) is meant to represent the semantic opposite of it.

Internally, a `TracePolicy` object is now instantiated outside of `StackdriverTracer` and passed in (it used to be instantiated by the `StackdriverTracer` instance).


src/config.ts
```diff
@@ -144,25 +172,31 @@ export interface Config {
...
+   * behavior of this field, as it controls whether trace context will be
+   * inherited from the incoming request.
    */
   contextHeaderBehavior?: ContextHeaderBehavior;
 
+  /**
+   * For advanced usage only.
+   * If specified, overrides the built-in trace policy object. Note that
+   * ignoreUrls, ignoreMethods, samplingRate, and contextHeaderBehavior will be
+   * ignored if this field is specified.
```
ofrobots: Add a throw if any of ignored fields are provided along with `tracePolicy`.

src/config.ts
```diff
@@ -200,12 +234,10 @@ export interface Config {
   onUncaughtException?: string;
 
   /**
-   * Setting this to true or false is the same as setting contextHeaderBehavior
-   * to 'ignore' or 'default' respectively. If both are explicitly set,
-   * contextHeaderBehavior will be prioritized over this value.
-   * Deprecated: This option will be removed in a future release.
+   * Whether or not to use incoming trace context headers for distributed
+   * trace context propagation.
```
ofrobots: Can you illustrate what happens in either case? How is `contextHeaderBehavior` used when this is `false`?

src/trace-api.ts
```diff
@@ -175,49 +154,58 @@ export class StackdriverTracer implements Tracer {
...
+      if (traceContext && this.config!.propagateTraceContextFromHeader) {
+        rootContext = new RootSpanData(
+            {
+              projectId: '',
+              traceId: traceContext.traceId,
+              spans: []
+            },                   /* Trace object */
+            name,                /* Span name */
+            traceContext.spanId, /* Parent's span ID */
+            options.skipFrames || 0);
```
ofrobots: It is hard to see what is different between the `then`/`else` blocks. Instead use local variables like so:
```ts
const traceId = traceContext && this.config!.propagateTraceContextFromHeader ?
  traceContext.traceId || traceId: uuid.v4().split('-').join('');
const spanId = traceContext && this.config!.propagateTraceContextFromHeader ?
  traceContext.spanId : 0;
```

src/tracing-policy.ts
```diff
@@ -76,15 +126,21 @@ export interface TracePolicyConfig {
...
+   * A field that controls filtering based on incoming trace context.
+   */
+  contextHeaderBehavior: TraceContextHeaderBehavior;
 }
 
 /**
  * A class that makes decisions about whether a trace should be created.
  */
-export class TracePolicy {
+export class TracePolicy implements config.TracePolicy {
```
ofrobots: nit: rename to `BuiltinTracePolicy`

src/config.ts
kjin: I've removed this field.

src/trace-api.ts
kjin: I've removed the then/else blocks.

src/config.ts
kjin: I've made it a warning... AFAICT we try not to throw, right?

ofrobots: This clearly indicates user error. This is a good example for when one should throw.

kjin: Ok, makes sense. I've changed it to throw instead.

src/config.ts
```diff
@@ -24,6 +24,34 @@ export type CLSMechanism =
...
+  traceContext: {traceId: string; spanId: string; options: number;}|null;
+  /**
+   * The original options object used to create the root span that corresponds
+   * to this request.
+   */
+  options: {};
+}
+
+export interface TracePolicy {
+  shouldTrace: (requestDetails: RequestDetails) => boolean;
```
ofrobots: It seems like boilerplate to pass in an object wrapping a funciton. Could you just accept a function?

src/config.ts
```diff
@@ -278,6 +302,7 @@ export const defaultConfig = {
   ignoreMethods: [],
   samplingRate: 10,
   contextHeaderBehavior: 'default',
+  propagateTraceContextFromHeader: true,
```
ofrobots: Is this needed?

src/config.ts
kjin: I'd rather we accept the object. With propagation we would have another overriding class `Propagation` where it makes more sense to have an object, because there are three fields. This would also allow users to easily implement a class and pass in an instance of a class.

Issue #1025, opened 2019-05-13
getCurrentContextId() returns null when not sampling


naartjie:
I am trying to use `LogEntry.trace` as a correlation identifier to tie different log messages together. I wasn't sure if this would be a good idea, but I got some confirmation from [here](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#FIELDS.trace_sampled):

> A non-sampled `trace` value is still useful as a request correlation identifier.

Being able to click on "Show matching entries" in the log view seems like a really useful thing to have:

<img width="448" alt="Screenshot 2019-05-13 at 20 33 18" src="https://user-images.githubusercontent.com/514563/57646097-16e83c80-75c0-11e9-8ace-a6cca64fb2f0.png">

The Bunyan logging wrapper uses the [`agent.getCurrentContextId()`](https://github.com/googleapis/nodejs-logging-bunyan/blob/d451b937139e5906ff66819788af6245256c08aa/src/index.ts#L44-L66). The problem is that `getCurrentContextId()` only seems to return a value if the current request is being sampled, and `null` otherwise.

Is this intended? Would it be possible to still return a traceId from `getCurrentContextId()` or perhaps expose another function to do this, for the sake of logging the trace id, to be able to correlate log entries.

kjin:
I believe not having a context ID for unsampled requests is for performance reasons. We don't currently have an option give context IDs even for unsampled requests, but it's something I can look into implementing, especially given that our own documentation suggests it.

naartjie:
Thanks for the super fast response @kjin 🚀 

If you think of any (even hacky) way to get the context ID in the meantime, please could you post it here also.

kjin:
IIUC, the Bunyan logging wrapper's use of getCurrentContextId is internal, so if there would also have to be a way for you to override its value. I can't remember off the top of my head whether this is possible -- will take a look at the code in a bit.

naartjie:
@kjin 

My project is using GKE/Cloud Run, and I'm just logging to stdout/stderr, so I'm not actually using the Bunyan library myself, hence I would be happy to use a workaround to extract the Context ID in my code.

Once a solution existed, do you think it would be a problem to take your suggested changes, and modify the internals of the (3?) Node.js GCP logging libraries to include `LogEntry.trace` in all log statements?

kjin:
[...]

PR #1032, opened 2019-05-21
refactor: remove traceWriter singleton from span-data.ts

kjin:
In preparation for a fix for #1025, this PR removes the references to the `traceWriter` singleton used in `SpanData` classes. It was used for two things:
* Publishing spans
* Getting span label limits

Now:
* Spans can be _enabled for publish_ by having a `TraceWriter` instance passed to them. This is so that in the future, we can have some spans that are "sampled" but not published, in order to address the issue mentioned above.
* Span label limits have been moved to `Tracer` config options instead, and passed in through the corresponding `SpanData` constructor.

### DIFF
--- a/src/span-data.ts
+++ b/src/span-data.ts
@@ -21,7 +21,7 @@ import { Constants, SpanType } from './constants';
 import { RootSpan, Span, SpanOptions } from './plugin-types';
 import { SpanKind, Trace, TraceSpan } from './trace';
 import { TraceLabels } from './trace-labels';
-import { traceWriter } from './trace-writer';
+import { TraceWriter } from './trace-writer';
 import * as traceUtil from './util';
 
 // Use 6 bytes of randomness only as JS numbers are doubles not 64-bit ints.
@@ -41,12 +41,20 @@ function randomSpanId() {
   return parseInt(spanRandomBuffer().toString('hex'), 16).toString();
 }
 
+export interface SpanLabelLimits {
+  stackTraceLimit: number;
+  maximumLabelValueSize: number;
+}
+
 /**
  * Represents a real trace span.
  */
 export abstract class BaseSpanData implements Span {
   readonly span: TraceSpan;
   abstract readonly type: SpanType;
+  // A TraceWriter instance. If not assigned via assignTraceWriter, the trace
+  // won't be enqueued for later publish.
+  protected traceWriter: TraceWriter | null = null;
 
   /**
    * Creates a trace context object.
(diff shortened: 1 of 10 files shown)


src/trace-api.ts
```diff
@@ -241,7 +242,7 @@ export class StackdriverTracer implements Tracer {
       rootContext = UNTRACED_ROOT_SPAN;
     } else {
       // Create a new root span, and invoke fn with it.
-      rootContext = new RootSpanData(
+      const rootSpan = new RootSpanData(
```
ofrobots: Why did you change this and add a new variable? Assigning to `rootContext` here should still work right?

src/span-data.ts
```diff
@@ -159,14 +179,16 @@ export class RootSpanData extends BaseSpanData implements RootSpan {
...
     super.endSpan(timestamp);
-    traceWriter.get().writeTrace(this.trace);
-    this.children.forEach(child => {
-      if (!child.span.endTime) {
-        // Child hasn't ended yet.
-        // Inform the child that it needs to self-publish.
-        child.shouldSelfPublish = true;
-      }
-    });
+    if (this.traceWriter) {
```
ofrobots: Consider the scenario where this code is reached, and `this.traceWriter` is not defined. Is that an error, or is that something very expected – and hence, silently ignorable, as being done here?

kjin: Right now there's no code path that causes `this.traceWriter` not to be defined, but this code path will exist in a forthcoming PR, where we would opt to not publish a trace, but still propagate context.

src/trace-api.ts
kjin: No, since `rootContext` is not a `RootSpanData`, and won't have a known `assignTraceWriter` field.

src/span-data.ts
ofrobots: IMO it would be better for the code to be self-documenting. `if (this.shouldPublish) { /* publish code */ }` would be more obviously readable. Using the validness of `this.traceWriter` to accomplish that, but it doesn't clarify whether it is possible that some bug may cause failure to call `assignTraceWriter`, causing traces to be silently dropped.

IOW, my question is: did we really need to get rid of the singleton? Wouldn't it make sense to keep it a singleton, but add boolean state to each span on whether it should be published?

kjin: > did we really need to get rid of the singleton?

I've decided that we don't need to get rid of it in order to address the issue at hand; I'll open a different PR with an alternative approach. That being said, I still think it is better not to depend on the singleton, and have the `TraceWriter` passed in -- it makes the code more testable and cleaner IMO. It should be considered a testable behavior of `Tracer` that it always calls `assignTraceWriter` so I don't know that a bug could crop up as long as we keep our tests intact. (I'll admit though that `assignTraceWriter` might not be the best name.)

Given that the change is ultimately inconsequential to users, though, I'd be OK with closing this PR.

kjin:
Superceded by work in #1033.

PR #1033, opened 2019-05-22
fix!: always assign a trace ID to each request

### LINKED ISSUE #1025: getCurrentContextId() returns null when not sampling

kjin:
Fixes #1025

We already store untraced requests (via a single object representing ALL untraced requests) in CLS, this changes it so that these objects also have a trace ID (and span ID, though not sure if it is useful).

### DIFF
--- a/src/span-data.ts
+++ b/src/span-data.ts
@@ -18,7 +18,7 @@ import * as crypto from 'crypto';
 import * as util from 'util';
 
 import { Constants, SpanType } from './constants';
-import { RootSpan, Span, SpanOptions } from './plugin-types';
+import { RootSpan, Span, SpanOptions, TraceContext } from './plugin-types';
 import { SpanKind, Trace, TraceSpan } from './trace';
 import { TraceLabels } from './trace-labels';
 import { traceWriter } from './trace-writer';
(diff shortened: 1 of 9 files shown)


kjin:
Actually, marking this as WIP (don't review) while I figure out what to do when we call `Tracer#createChildSpan`.

src/span-data.ts
```diff
@@ -236,10 +277,9 @@ export const UNCORRELATED_CHILD_SPAN = createPhantomSpanData(
 
 /**
  * A virtual trace span that indicates that a real child span couldn't be
- * created because the corresponding root span was disallowed by user
- * configuration.
+ * created because the Trace Agent was disabled.
  */
-export const UNTRACED_CHILD_SPAN = createPhantomSpanData(SpanType.UNTRACED);
+export const DISABLED_CHILD_SPAN = createPhantomSpanData(SpanType.UNTRACED);
```
ofrobots: This is going to sound nit-picky...

Is there a reason why the concept was renamed on the left hand side but left the same on the right hand side?

Based on your decision to rename the export, it seems that you prefer external users to conceptually think about this concept as `disabled`. If that is the right conceptualization, why is not the right conceptualization for the internals in this file?

kjin: The variable formerly named `UNTRACED_CHILD_SPAN` is now only used when the Trace Agent is disabled which is why the LHS changed (where as before it is used for a disabled agent or for unsampled traces). As for the RHS, there continues to be multiple reasons for a span object to represent "untraced" work -- either unsampled, or disabled.

I could get behind the idea that we could just introduce `UNSAMPLED` and `DISABLED` in place of `UNTRACED`, though.

ofrobots: +1 on the rename. I won't block this PR, but conceptual clarity is important.

PR #1034, opened 2019-05-23
refactor: drop usage of pify

JustinBeckwith:
With the deprecation of support for Node.js 6.x, we can rely on `util.promisify`.


scripts/utils.ts
```diff
@@ -1,19 +1,19 @@
...
-export const readFileP: (path: string, encoding?: string) => Promise<Buffer|string> = pify(readFile);
-export const writeFileP: (path: string, data: Buffer|string, encoding?: string) => Promise<void> = pify(writeFile);
-export const statP: (path: string) => Promise<Stats> = pify(stat);
-export const tmpDirP: () => Promise<string> = pify(tmp.dir);
+export const globP = promisify(glob);
+export const mkdirP = promisify(mkdir);
+export const readFileP = promisify(readFile);
+export const writeFileP = promisify(writeFile);
+export const statP = promisify(stat);
+export const tmpDirP = promisify(tmp.dir) as () => Promise<string>;
```
ofrobots: Why is the cast necessary?

JustinBeckwith: For some reason, the magic casting work `promisify` usually does didn't work here.  In the `tmpDirP` method, it still assumed the method would take a single parameter that had a callback.  In *most* cases the types work out, but I've seen this in a few other places as well.

PR #1052, opened 2019-06-24
fix: copy credentials in internal config

### LINKED ISSUE #1048: KeyFilename credentials not copied in internal config
My problem is when i tried to switch from version 3.6.1 to 4.0.0 I cannot trace anymore because of a 403 when I'm publishing my traces. I did some digging in your code and I found out that in your index.js you guys did a massive change on how you copy the user defined configuration to your internal configuration to setup the tracing. In my project i'm using a keyFilename as authentication/authorization, but it is not copied over!

Before you guys where doing 
``` 
	 // Configuration order of precedence:
    // 1. Environment Variables
    // 2. Project Config
    // 3. Environment Variable Set Configuration File (from GCLOUD_TRACE_CONFIG)
    // 4. Default Config (as specified in './config')
    const config = extend(true, { [util_1.FORCE_NEW]: projectConfig[util_1.FORCE_NEW] }, config_1.defaultConfig, envSetConfig, contextHeaderBehaviorUnderride, projectConfig, envConfig, { plugins: {} });
    ...

    //and you where returning the object config directly
    return config;
```

At first, you where returning the object config directly which was properly transferring everything....now you guys do 

```
// Configuration order of precedence:
    // 1. Environment Variables
    // 2. Project Config
    // 3. Environment Variable Set Configuration File (from GCLOUD_TRACE_CONFIG)
    // 4. Default Config (as specified in './config')
    const mergedConfig = extend(true, {}, config_1.defaultConfig, envSetConfig, userConfig);
    ...

    //here!!! you return another object the takes parts of the mergeConfig....which cause the missing properties
     return {
        [util_1.FORCE_NEW]: forceNew,
        enabled: mergedConfig.enabled,
        logLevel: util_1.lastOf(mergedConfig.logLevel, Number(process.env.GCLOUD_TRACE_LOGLEVEL)),
        clsConfig: {
            [util_1.FORCE_NEW]: forceNew,
            mechanism: getInternalClsMechanism(mergedConfig.clsMechanism),
        },
        writerConfig: {
            [util_1.FORCE_NEW]: forceNew,
            projectId: util_1.lastOf(mergedConfig.projectId, process.env.GCLOUD_PROJECT),
            onUncaughtException: mergedConfig.onUncaughtException,
            bufferSize: mergedConfig.bufferSize,
            flushDelaySeconds: mergedConfig.flushDelaySeconds,
            stackTraceLimit: mergedConfig.stackTraceLimit,
            maximumLabelValueSize: Math.min(mergedConfig.maximumLabelValueSize, constants_1.Constants.TRACE_SERVICE_LABEL_VALUE_LIMIT),
            serviceContext: {
                service: util_1.lastOf(mergedConfig.serviceContext.service, process.env.GAE_MODULE_NAME, process.env.GAE_SERVICE),
                version: util_1.lastOf(mergedConfig.serviceContext.version, process.env.GAE_MODULE_VERSION, process.env.GAE_VERSION),
                minorVersion: util_1.lastOf(mergedConfig.serviceContext.minorVersion, process.env.GAE_MINOR_VERSION),
            },
        },
        pluginLoaderConfig: {
            [util_1.FORCE_NEW]: forceNew,
            plugins: Object.assign({}, mergedConfig.plugins),
            tracerConfig: {
                enhancedDatabaseReporting: mergedConfig.enhancedDatabaseReporting,
                rootSpanNameOverride: getInternalRootSpanNameOverride(mergedConfig.rootSpanNameOverride),
                spansPerTraceHardLimit: mergedConfig.spansPerTraceHardLimit,
                spansPerTraceSoftLimit: mergedConfig.spansPerTraceSoftLimit,
            },
        },
        tracePolicyConfig: {
            samplingRate: mergedConfig.samplingRate,
            ignoreMethods: mergedConfig.ignoreMethods,
            ignoreUrls: mergedConfig.ignoreUrls,
            contextHeaderBehavior: mergedConfig.contextHeaderBehavior,
        },
        overrides: {
            tracePolicy: mergedConfig.tracePolicy,
            propagation: mergedConfig.propagation,
        },
    };

```

you are returning another object which takes some parts of the mergeConfig, which is missing user defined properties...like my keyFilename...for my Authorization

kjin:
Fixes #1048

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -125,10 +125,6 @@ function initConfig(userConfig: Forceable<Config>): TopLevelConfig {
     },
     writerConfig: {
       [FORCE_NEW]: forceNew,
-      projectId: lastOf<string | undefined>(
-        mergedConfig.projectId,
-        process.env.GCLOUD_PROJECT
-      ),
       onUncaughtException: mergedConfig.onUncaughtException,
       bufferSize: mergedConfig.bufferSize,
       flushDelaySeconds: mergedConfig.flushDelaySeconds,
@@ -153,6 +149,18 @@ function initConfig(userConfig: Forceable<Config>): TopLevelConfig {
           process.env.GAE_MINOR_VERSION
         ),
       },
+      /**
+       * Our TypeScript interface suggests that only credentials, keyFilename,
+       * and projectId are accepted, but by passing the entire object to the
+       * Trace Writer, we can allow users to supply other fields that are
+       * publicly supported by the Google Auth Library.
+       */
+      authOptions: Object.assign({}, mergedConfig, {
+        projectId: lastOf<string | undefined>(
+          mergedConfig.projectId,
+          process.env.GCLOUD_PROJECT
+        ),
+      }),
     },
     pluginLoaderConfig: {
       [FORCE_NEW]: forceNew,
--- a/src/trace-writer.ts
+++ b/src/trace-writer.ts
@@ -40,8 +40,8 @@ const SCOPES: string[] = ['https://www.googleapis.com/auth/trace.append'];
 /* The API endpoint of the Stackdriver Trace service */
 const TRACE_API_ENDPOINT = 'cloudtrace.googleapis.com';
 
-export interface TraceWriterConfig extends common.GoogleAuthOptions {
-  projectId?: string;
+export interface TraceWriterConfig {
+  authOptions: common.GoogleAuthOptions;
   onUncaughtException: string;
   bufferSize: number;
   flushDelaySeconds: number;
@@ -124,7 +124,7 @@ export class TraceWriter extends common.Service {
         baseUrl: `https://${TRACE_API_ENDPOINT}/v1`,
         scopes: SCOPES,
       },
-      config
+      config.authOptions
     );
 
     this.logger = logger;
--- a/src/tracing.ts
+++ b/src/tracing.ts
@@ -154,8 +154,8 @@ export class Tracing implements Component {
       .activate();
 
     if (
-      typeof this.config.writerConfig.projectId !== 'string' &&
-      typeof this.config.writerConfig.projectId !== 'undefined'
+      typeof this.config.writerConfig.authOptions.projectId !== 'string' &&
+      typeof this.config.writerConfig.authOptions.projectId !== 'undefined'
     ) {
       this.logger.error(
         'StackdriverTracer#start: config.projectId, if provided, must be a string.',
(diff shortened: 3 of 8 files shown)


src/index.ts
```diff
@@ -153,6 +153,13 @@ function initConfig(userConfig: Forceable<Config>): TopLevelConfig {
           process.env.GAE_MINOR_VERSION
         ),
       },
+      /**
+       * Our TypeScript interface suggests that only keyFilename and credentials
+       * are accepted, but by passing the entire object to the Trace Writer,
+       * we can allow users to supply other fields that are publicly supported
+       * by the Google Auth Library.
+       */
+      authOptions: mergedConfig,
```
ofrobots: Auth options can only come from `userConfig` right? While `mergedConfig` should work, the tradeoff I see is that it may pass too many options and potentially conflict with some other named property in Auth down the road. Passing `userConfig` means that we are implicitly assuming that none of the auth options would be overridden by the environment or default configs.

kjin: I believe it's plausible that variables in `GCLOUD_TRACE_CONFIG` can be used for authentication. At least, that was the behavior from before.

src/index.ts
```diff
@@ -153,6 +149,18 @@ function initConfig(userConfig: Forceable<Config>): TopLevelConfig {
...
+      /**
+       * Our TypeScript interface suggests that only credentials, keyFilename,
+       * and projectId are accepted, but by passing the entire object to the
+       * Trace Writer, we can allow users to supply other fields that are
+       * publicly supported by the Google Auth Library.
+       */
+      authOptions: Object.assign({}, mergedConfig, {
+        projectId: lastOf<string | undefined>(
+          mergedConfig.projectId,
+          process.env.GCLOUD_PROJECT
```
ofrobots: Doesn't the auth library already consult this environment variable?

JustinBeckwith: Yeah, we do.

kjin: It's unfortunate but I realized that Trace Agent has historically regarded `GCLOUD_PROJECT` as taking precedence over the `projectId` config option, while the auth library does the [opposite](https://github.com/googleapis/google-auth-library-nodejs/blob/197dd87f0019ac4314043e961eb8b3ba78670c4c/src/auth/googleauth.ts#L171). Maybe it's possible that we change this now, though.

ofrobots: That would be semver major :/. Not worth it.

PR #1058, opened 2019-06-27
build: use config file for linkinator

JustinBeckwith:


JustinBeckwith:
This repo was already using the config file 🙃

PR #1087, opened 2019-07-31
test: accomodate private method name change

kjin:
google-auth-library@5.1.1 made a private method name change which affects our tests, this updates our test to accomadate for this change.


kjin:
TBR=@googleapis/node-team

PR #1122, opened 2019-09-26
chore: add protos/ to .eslintignore

alexander-fenster:


kjin:
Force-merging as this should not have any effect on anything.

PR #1163, opened 2019-12-05
fix(deps): pin TypeScript below 3.7.0

bcoe:


bcoe:
@kjin :wave: unfortunately, `TypeScript@3.7.0` is not compatible with `3.5.3` (which is the version Angular is pinned to, and a lot of our users use).

Any thoughts on the failing circle-ci tests here?

kjin:
@bcoe I was OOO yesterday, will look at them today. Thanks for the reminder!

bcoe:
@kjin I would say medium urgent, it means our TypeScript folks on Angular can't run trace; I don't have permission to force push unfortunately.

PR #1166, opened 2019-12-06
chore: downgrade to got@9 in system tests

kjin:
got@10 doesn't support Node 8, and also makes a breaking API change which requires an amendment to our system tests (this isn't a bad or breaking change for us, but for now, we will postpone making a change in our tests).


kjin:
TBR=@googleapis/node-team

PR #1174, opened 2019-12-24
build: use c8 for coverage

JustinBeckwith:


alexander-fenster:
Tests for this package are broken, who we gonna call?

PR #1191, opened 2020-02-01
test: modernize mocha config

JustinBeckwith:


DominicKramer:
Looks like `source-map-support` is used in the `src/utils.ts` file.

PR #1213, opened 2020-03-06
chore: set up docker containers for github actions

kjin:


bcoe:
@kjin just let me know when you're getting closer to a test configuration you'd like, and we can turn on and off required checks as needed.

kjin:
Hey @bcoe, sorry for the delay. I think things seem to be green now, PTAL. Thanks!

PR #1226, opened 2020-03-30
DO NOT MERGE: experiment with GitHub Actions

JustinBeckwith:


kjin:
Hey Justin, I finally started making progress on #1213 again after seeing this PR. Granted it uses services rather than actions as you did here but it should be working now. PTAL and let me know if it covers what you want here, thanks!

bcoe:
addressed in https://github.com/googleapis/cloud-trace-nodejs/pull/1213

PR #1228, opened 2020-03-31
build: don't overwrite ci.yml, retire .circleci

bcoe:
now that GitHub actions are working:

1. don't stop them by working by moving to our standard config (with no mongo).
2. stop running circle-ci.


bcoe:
@kjin fair point, re: replace, I think in this case I'm comfortable carving out an exception though, vs., needing to get too fancy with a replacement.

PR #1247, opened 2020-04-18
build: attempt to make the build more consistent

JustinBeckwith:


JustinBeckwith:
This is busted right now because I was experimenting and codecov isn't working

PR #1257, opened 2020-05-04
fix: sample app TypeError

### LINKED ISSUE #1246: app sample broken: TypeError: The `GET` method cannot be used with a body
#### Environment details

  - OS: macOS
  - Node.js version: 10.20.1
  - npm version:   6.14.4
  - `@google-cloud/trace-agent` version: 4.2.5

#### Steps to reproduce

```
cd samples
npm install
npm start
curl localhost:8080
```

```
-> "TypeError: The `GET` method cannot be used with a body"
```

#### Notes

It looks like this example is using the `got` API wrongly: The `json` argument is supposed to be a serializable POJO, not a boolean: https://github.com/sindresorhus/got#json

ZiweiZhao:
App sample has error: TypeError: The `GET` method cannot be used with a body
This pull request is to update the body to an json response option

Fixes #1246 🦕

### DIFF
--- a/samples/app.js
+++ b/samples/app.js
@@ -31,7 +31,7 @@ const DISCOVERY_URL = 'https://www.googleapis.com/discovery/v1/apis';
 app.get('/', async (req, res) => {
   // This outgoing HTTP request should be captured by Trace
   try {
-    const {body} = await got(DISCOVERY_URL, {json: true});
+    const {body} = await got(DISCOVERY_URL, {responseType: 'json'});
     const names = body.items.map(item => item.name);
     res.status(200).send(names.join('\n')).end();
   } catch (err) {


bcoe:
failing Windows tests are being discussed here:

https://github.com/nodejs/node-gyp/issues/2114

kintel:
@bcoe To move forward while the Node issues in question gets resolved (new Node version or new needle version), would it make sense to pin Node on Windows to 12.16.2 here? https://github.com/googleapis/cloud-trace-nodejs/blob/master/.github/workflows/ci.yaml#L61

bcoe:
@kintel yes, I would just switch to Node 10 on this branch for the purposes of landing. I believe the upstream issue will be fixed soon in Node.

JustinBeckwith:
Nice fix!  One question though - are we missing an integration test for the sample?  I'm a little surprised this didn't get captured in the end to end test.

kintel:
@JustinBeckwith technically there is a test, but the test just checks that the process starts, not that it does the right thing. We should probably open a separate issue for improving the test.

kintel:
Opened follow-up issue: #1259

PR #1309, opened 2020-09-18
refactor: avoid semver where possible

TrySound:
Checking node version is not necessary anymore with "engines" field in
package.json which prevents from installing this package with legacy node.

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/cloud-trace-nodejs/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes #<issue_number_goes_here> 🦕

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [10, 12, 13]
+        node: [10, 12, 14, 15]
     # Docker containers to run for database RPC tracing integration tests.
     services:
       mongo:
--- a/.kokoro/release/docs-devsite.sh
+++ b/.kokoro/release/docs-devsite.sh
@@ -37,9 +37,13 @@ NAME=$(cat .repo-metadata.json | json name)
 mkdir ./_devsite
 cp ./yaml/$NAME/* ./_devsite
 
+# Clean up TOC
 # Delete SharePoint item, see https://github.com/microsoft/rushstack/issues/1229
 sed -i -e '1,3d' ./yaml/toc.yml
 sed -i -e 's/^    //' ./yaml/toc.yml
+# Delete interfaces from TOC (name and uid)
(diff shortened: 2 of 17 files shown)


JustinBeckwith (approved):
Thanks for this!  I'm curious - did you see other places we were using the module?  Wondering if we can remove it from package.json all together.

TrySound:
Yes, I found `isSupported` method which is used for plugins.
https://github.com/googleapis/cloud-trace-nodejs/blame/master/doc/plugin-guide.md#L10

bcoe:
👋 most folks do not have `--engines-strict` enabled for their `npm i`, so engines is actually not a great way to protect users from these types of issues. Throwing a clear error message is a much better user experience.

I'd be open to dropping the SemVer dependency, but I would want to see it instead replaced with logic that checks the major version, [like this](https://github.com/yargs/yargs-parser/blob/master/lib/index.ts#L10).

bcoe:
@TrySound this has stalled for a little while, did you have any interested in updating the PR to still throw on unsupported engine versions?

TrySound:
Oops, sorry. Forgot about it. Will update today.

bcoe:
@TrySound looks reasonable to me, could I just bother you to run `npm run fix`, for the benefit of the linter.

TrySound:
```
$ npm run fix

> @google-cloud/trace-agent@5.1.1 fix /.../cloud-trace-nodejs
> gts fix

version: 14

/.../cloud-trace-nodejs/samples/app.js
  20:11  error  "@google-cloud/trace-agent" is not found  node/no-missing-require

/.../cloud-trace-nodejs/samples/snippets.js
  18:9  error  "@google-cloud/trace-agent" is not found  node/no-missing-require

✖ 2 problems (2 errors, 0 warnings)
```

TrySound:
Hm... fix does not have any effect

PR #1395, opened 2021-10-20
chore(cloud-rad): delete api-extractor config

fhinkel:


bcoe (approved):

gkf (commented):
Is there a bug or other reference for why this config should be deleted? I am not familiar with api-extractor so I would just like to understand the context.

fhinkel:
It's part of this template change https://github.com/googleapis/cloud-trace-nodejs/pull/1355

All the api-extractor config and procedure are inside the nodejs-cloud-rad package now.

PR #1442, opened 2022-05-20
build!: update library to use Node 12

sofisl:
Migrate library to Node 12


dnoe:
@aabmass Can you take a look? Let me know if you haven't had the approver bit turned on for your user yet and you need me to stamp this.

.github/workflows/ci.yaml
```diff
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [10, 12, 14, 16]
+        node: [12, 14, 16]
```
aabmass, line 12: Should we add node 18 as well since it's current?

aabmass (commented):
Do recommend we make a major version release of this repo after this?

In addition to removing official support for node10, I know some TS major version bumps can produce incompatibilities in the compiled output

aabmass (approved):
LGTM

.github/workflows/ci.yaml
sofisl: It's not yet stable, but we will do this once we migrate to 14!

PR #1489, opened 2023-03-07
feat: support restify v9-v11

### LINKED ISSUE #1488: Update Restify Plugin to support Restify versions 9-11
The existing Restify plugin is [defined](https://github.com/googleapis/cloud-trace-nodejs/blob/96b803392dde3c9281982d20c3cd5746c5191715/src/plugins/plugin-restify.ts#L28) to only operate up to version 8. It was last updated in Apr 2020 from #1250 . The [restify changelog](https://github.com/restify/node-restify/blob/master/CHANGELOG.md) has newer, recent releases:
9.0.0 2022-11-05 (drops node <8)
10.0.0 2022-11-29 (adds support for node 18)
11.0.0 2023-01-17

Is it possible to upgrade the restify plugin version range to enable support for the new versions - especially so that node 18 can be supported with restify and this plugin?

6utt3rfly:
Fixes #1488

(see also #1250 for reference)

### DIFF
--- a/.readme-partials.yaml
+++ b/.readme-partials.yaml
@@ -59,7 +59,7 @@ body: |-
     * [gRPC](https://www.npmjs.com/package/grpc) server (version ^1.1)
     * [hapi](https://www.npmjs.com/package/hapi) (versions 8 - 19)
     * [koa](https://www.npmjs.com/package/koa) (version 1 - 2)
-    * [restify](https://www.npmjs.com/package/restify) (versions 3 - 8)
+    * [restify](https://www.npmjs.com/package/restify) (versions 3 - 11)
 
     The agent will also automatically trace RPCs from the following modules:
     * Outbound HTTP requests through `http`, `https`, and `http2` core modules
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ The trace agent can do automatic tracing of the following web frameworks:
 * [gRPC](https://www.npmjs.com/package/grpc) server (version ^1.1)
 * [hapi](https://www.npmjs.com/package/hapi) (versions 8 - 19)
 * [koa](https://www.npmjs.com/package/koa) (version 1 - 2)
-* [restify](https://www.npmjs.com/package/restify) (versions 3 - 8)
+* [restify](https://www.npmjs.com/package/restify) (versions 3 - 11)
 
 The agent will also automatically trace RPCs from the following modules:
 * Outbound HTTP requests through `http`, `https`, and `http2` core modules
--- a/src/plugins/plugin-restify.ts
+++ b/src/plugins/plugin-restify.ts
@@ -25,7 +25,7 @@ type Response = restify_5.Response;
 type Next = restify_5.Next;
 type CreateServerFn = (options?: restify_5.ServerOptions) => restify_5.Server;
 
-const SUPPORTED_VERSIONS = '<=8.x';
+const SUPPORTED_VERSIONS = '<=11.x';
 
 function unpatchRestify(restify: Restify5) {
   shimmer.unwrap(restify, 'createServer');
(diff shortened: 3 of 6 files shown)


punya:
Please take a look at the test failures. For example,

```
  1) Web framework tracing
       Tracing restify@9
         "before all" hook for "accurately measures get time (1 handler)":

      AssertionError [ERR_ASSERTION]: Handler [handler-0 on GET /one-handler] accepts a third argument (the 'next" callback) but is also an async function. Middleware handlers can be either async/await or callback-based. Async handler functions should accept maximum of 2 arguments: (req, res). Non-async handlers should accept three arguments: (req, res, next).
      + expected - actual


      at Chain.add (test/plugins/fixtures/restify9/node_modules/restify/lib/chain.js:91:16)
      at forEach (test/plugins/fixtures/restify9/node_modules/restify/lib/router.js:211:15)
      at Array.forEach (<anonymous>)
      at Router.mount (test/plugins/fixtures/restify9/node_modules/restify/lib/router.js:203:14)
      at Server.serverMethod [as get] (test/plugins/fixtures/restify9/node_modules/restify/lib/server.js:1750:33)
      at _a.addHandler (/home/runner/work/cloud-trace-nodejs/cloud-trace-nodejs/test/web-frameworks/restify.ts:38:19)
      at Context.<anonymous> (/home/runner/work/cloud-trace-nodejs/cloud-trace-nodejs/test/test-trace-web-frameworks.ts:113:22)
      at processImmediate (node:internal/timers:466:21)
      at process.topLevelDomainCallback (node:domain:161:15)
      at process.callbackTrampoline (node:internal/async_hooks:128:24)
```

Could you confirm whether you've tested this change locally, and with what versions?

6utt3rfly:
@punya - I had trouble running local unit tests until I saw this PR's build log and realized I needed docker images running (or TRACE_TEST_EXCLUDE_INTEGRATION set). It might be worth adding something to the CONTRIBUTING.md file?

The existing restify plugin works if I manually patch the SUPPORTED_VERSIONS. I've pushed a commit to fix the failing unit test by avoiding `async (req, res, next)`, which breaks in restify 9+

punya (changes requested):
Thanks for looking into it @6utt3rfly! I have a couple of small suggestions, but otherwise this looks good to me.

6utt3rfly:
> Thanks for looking into it @6utt3rfly! I have a couple of small suggestions, but otherwise this looks good to me.

I actually hadn't noticed `.then` accepts both `resolve` and `reject`, and thought it was only in `new Promise(resolve, reject)`, so I like your suggestion!

punya:
I see some lint failures that should be easy to fix:

```
/home/runner/work/cloud-trace-nodejs/cloud-trace-nodejs/test/web-frameworks/restify.ts
Warning:   20:3   warning  'WebFrameworkResponse' is defined but never used  @typescript-eslint/no-unused-vars
Error:   41:17  error    Replace `(response)` with `response`              prettier/prettier
```

But there are also several failing tests (starting at https://github.com/googleapis/cloud-trace-nodejs/actions/runs/4694210640/jobs/8322181339?pr=1489#step:7:2986) that I don't understand.

Reading the Restify docs in some more detail, I'm not sure that we're supposed to use 3-argument handlers with Promises (regardless of whether we use `async/await` syntax to generate the Promise).

I wonder if it might be necessary to have totally different code paths for Restify <= v8 and Restify >= v9.

6utt3rfly:
@punya - `Cannot find node module 'node:process'` seems to indicate a node compatibility issue with restify (even though their release notes indicate node 10 and 12 support). Are you okay with restricting the unit tests to node 14+ (see [commit](https://github.com/googleapis/cloud-trace-nodejs/pull/1489/commits/bf9ec4b90cd4901ccca5e3fe4166ffaec1ad003e))?

punya (approved):

punya (approved):

6utt3rfly:
@punya - I'm not sure why the release action hanged (5h 58m!), because it looks like everything passed, but then it didn't get the coverage
image

vs the PR test:
image

... maybe worth just retrying that job?

